PR #24, opened 2022-03-04
Tiles updater from Broadcast Receiver

yschimke:
For discussion

Scope: Sending a Tile Update from the Broadcast Receiver which does not have lifecycle to bind to a service.

See client code with all 3 approaches in https://github.com/google/horologist/pull/24/files#diff-fdf2926051fc0d341bde50cf559bd4760fd4cc0fe02642c60ab61ffc3380f024

General TODO - put this behind a TileUpdateRequesterCompat class with requestUpdate and requestUpdateAsync methods.

----

Option 1 - Use ApplicationContext **(BAD)**

Not reliable, but fixes the problem (context.applicationContext vs context).  Maybe appropriate advice when updates are via broadcast receivers for apps that are assumed to be running.  **I will remove this from the PR.**
 
----

Option 2 - Use a Service.  All code and manifest entries for the service would be in the horologist-tiles.  So this would be a 1 liner.

See https://github.com/google/horologist/pull/24/files#diff-b739305d93549d86f7d39ee18f950a3b7c7934ea81ad1fab28f5e05af247acd5

TODO - Make this a foreground service.

---

Option 3 - User WorkManager - this is apparently the preferred way to do 

See https://github.com/google/horologist/pull/24/files#diff-4923c1f91ad21993c63125203c5749d93d2932528458ef0e7e18bf701ba19d4f


yschimke:
Will fix and reopen later with just the service impl.

yschimke:
Next plan is to support this in CoroutinesTileService.

PR #27, opened 2022-03-18
Add slots for WearNavScaffold

yschimke:
Add slots for WearNavScaffold.

Needed to allow setting leading and trailing text.

Also adds a basic test to confirm the right calls are made.


yschimke:
@JohnNichol I do wonder whether this it too flexible, allowing a different TimeText per screen.  If that is desired you probably need the ViewModel like PositionIndicator gets or route?

JohnNichol:
I don't see a need for a different TimeText per screen personally. I think there are typically 2 key questions, 1) what is the status text(s) leading/trailng, and 2) do we want time text animated away for $reasons. Typically I expect it to be a really simple case of simple leading _status_ text before the time.

yschimke:
Should the TimeText be common? A single composable that is hidden or scaled/faded based on the screen?

When you say "leading status text" do you mean a String? Or a slot still?

In my use case I put network icons in this (for dev builds only) so I hope it can feature non-text.

yschimke:
@JohnNichol agree on not needing a TimeText per screen, and I've removed PositionIndicator as a param.  but this does allow you to customise TimeText, such as including leading or trailing content.

compose-layout/src/main/java/com/google/android/horologist/compose/navscaffold/WearNavScaffold.kt
```diff
@@ -161,6 +151,36 @@ public fun WearNavScaffold(
...
+
+@Composable
+private fun NavPositionIndicator(viewModel: NavScaffoldViewModel) {
+    when (viewModel.scrollType) {
+        NavScaffoldViewModel.ScrollType.ScrollState ->
+            PositionIndicator(
+                scrollState = viewModel.scrollableState as ScrollState
+            )
+        NavScaffoldViewModel.ScrollType.ScalingLazyColumn -> {
+            val scalingLazyListState =
```
luizgrp, line 169: nit: unnecessary variable

compose-layout/src/main/java/com/google/android/horologist/compose/navscaffold/WearNavScaffold.kt
```diff
@@ -161,6 +151,36 @@ public fun WearNavScaffold(
...
+
+            PositionIndicator(
+                scalingLazyListState = scalingLazyListState
+            )
+        }
+        NavScaffoldViewModel.ScrollType.LazyList ->
+            PositionIndicator(
+                lazyListState = viewModel.scrollableState as LazyListState
+            )
+        else -> {}
```
luizgrp, line 180: nit: should add a comment to make it explicit that the intention is to have a no-op else branch

compose-layout/src/androidTest/java/com/google/android/horologist/compose/navscaffold/NavScaffoldTest.kt
```diff
@@ -0,0 +1,266 @@
...
+            }
+        }
+
+        composeTestRule.waitForIdle()
+
+        val timeText = composeTestRule.onNodeWithTag("timeText")
+        val body = composeTestRule.onNodeWithTag("body")
+
+        // expect time text in the top 50 pixels
+        assertThat(bounds?.bottom).isLessThan(75)
```
luizgrp, line 252: nit: should it be `bounds!!` instead of `bounds?` to fail test in the scenario that bounds is null?

luizgrp (approved):

compose-layout/src/androidTest/java/com/google/android/horologist/compose/navscaffold/NavScaffoldTest.kt
yschimke: It will because of the isLessThan, but should be a cleaner error.

compose-layout/src/main/java/com/google/android/horologist/compose/navscaffold/WearNavScaffold.kt
yschimke: This was a bug I found in a test and fixed, I'll merge main so the fix is in.

sample/build.gradle
```diff
@@ -100,6 +100,6 @@ dependencies {
     androidTestImplementation libs.compose.ui.test.junit4
     androidTestImplementation libs.espresso.core
     androidTestImplementation libs.junit
-    androidTestImplementation "androidx.test.ext:junit:1.1.3"
-    androidTestImplementation "androidx.test.ext:junit-ktx:1.1.3"
+    androidTestImplementation libs.androidx.test.ext
+    androidTestImplementation libs.androidx.test.ext.ktx
```
luizgrp, line 104: 👍

luizgrp (approved):

PR #29, opened 2022-03-22
Catch request focus exception

yschimke:
Apps using NavScaffold may be ignoring the focusRequestor.  If not attached then using it fails badly.

This is a temporary fix, we should make it near impossible to request a focusRequestor without using it, but that means API changes.

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/navscaffold/NavScaffoldViewModel.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/navscaffold/NavScaffoldViewModel.kt
@@ -16,6 +16,7 @@
 
 package com.google.android.horologist.compose.navscaffold
 
+import android.util.Log
 import androidx.compose.foundation.ScrollState
 import androidx.compose.foundation.gestures.ScrollableState
 import androidx.compose.foundation.lazy.LazyListState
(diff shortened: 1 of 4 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/navscaffold/NavScaffoldViewModel.kt
```diff
@@ -152,6 +153,16 @@ public open class NavScaffoldViewModel(
         return _scrollableState as LazyListState
     }
 
+    public fun requestFocus() {
+        if (focusRequested) {
```
JohnNichol, line 157: If requestFocus is called would that not result in focusRequested going to true? I don't follow why we would ignore the call to requestFocus here.

yschimke: The ViewModel has this check, precisely so that it doesn't create the focusRequestor itself.  If the client doesn't call focusRequestor then it knows to skip making this call. 

The awkward thing about this NavScaffold is the PositionIndicator and the Scrollable Component are separated by Time and Space.  Well really App/Library and Activity/Nav Destination and can only communicate through the NavScaffoldViewModel which is owned by the NavBackStackEntry (a ViewModelStoreOwner).  So there needs to be some contract for them to communicate whether these things should be wired up.

The bug happens when the app gets a focusRequestor but doesn't use it.

yschimke: It bit me when I had a screen with a LazyColumn, but I didn't include it until the data had loaded from Room DB.  In that case you get this error.

yschimke: But specifically for your question, no, if `focusRequested == false` then this method is a no-op.

Issue #34, opened 2022-03-22
Add screenshot tests 


yschimke:
The bugs, both caused by framework changes, and 1 I introduced myself show we need more test coverage here.

yschimke:
First one https://github.com/google/horologist/pull/36/files

yschimke:
To close when paparazzi release is available.

PR #30, opened 2022-03-22
Fix volume scrolling.

yschimke:
Broke after a wear compose update.
Tracked in https://b.corp.google.com/issues/226089343

PR #35, opened 2022-03-22
Fix scrolling behaviour when autoHide = true.

yschimke:
Bug from https://github.com/google/horologist/pull/30

Raised https://github.com/google/horologist/issues/34 to add some tests which it's obvious we are missing.  Samples aren't enough as they are not automated.


yschimke:
Landing to unblock an internal PR.

PR #36, opened 2022-03-22
Add PositionIndicator test

yschimke:


yschimke:
Landing as test only - I'll follow up with further reviews. Let me know if you'd rather I wait for test PRs.

PR #37, opened 2022-03-23
Update compose dependencies

yschimke:
Removes the need to reverse the VolumePositionIndicator.


yschimke:
https://user-images.githubusercontent.com/231923/159770591-39c9d91c-0138-4a1c-83cd-4f17479c207e.mp4

audio-ui/src/main/java/com/google/android/horologist/audioui/VolumePositionIndicator.kt
```diff
@@ -61,26 +59,19 @@ public fun VolumePositionIndicator(
         }
     }
 
-    // Position Indicator is hardcoded to right side,
-    // so workaround it.
-    Box(
-        modifier = Modifier.rotate(180f)
+    AnimatedVisibility(
```
yschimke, line 62: No longer hardcoded in Wear Compose

audio-ui/src/main/java/com/google/android/horologist/audioui/VolumePositionIndicator.kt
```diff
@@ -41,8 +39,8 @@ import kotlinx.coroutines.delay
  */
 @Composable
 public fun VolumePositionIndicator(
-    modifier: Modifier = Modifier,
     volumeState: () -> VolumeState,
+    modifier: Modifier = Modifier,
```
luizgrp, line 43: nice one! from the Modifier [docs](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier):

> It should appear as the first optional parameter in the parameter list; after all required parameters (except for trailing lambda parameters) but before any other parameters with default values.

luizgrp (approved):
LGTM

audio-ui/src/main/java/com/google/android/horologist/audioui/VolumePositionIndicator.kt
yschimke: I agree and started that way, but deliberately changed it to match

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/compose/compose-material/src/commonMain/kotlin/androidx/wear/compose/material/PositionIndicator.kt;l=289?q=PositionIndicator

yschimke: Yep - sorry I misread this previously.

PR #41, opened 2022-03-24
Test Activity Lifecycle

yschimke:
Since I was caught out by lifecycle logic, penance is a test doing mainly that.

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -94,4 +94,10 @@ dependencies {
     implementation libs.androidx.wear.tiles
 
     implementation libs.kotlin.stdlib
+
+    androidTestImplementation libs.compose.ui.test.junit4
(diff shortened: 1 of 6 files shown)


sample/build.gradle
```diff
@@ -94,4 +94,10 @@ dependencies {
     implementation libs.androidx.wear.tiles
 
     implementation libs.kotlin.stdlib
+
+    androidTestImplementation libs.compose.ui.test.junit4
+    androidTestImplementation libs.espresso.core
+    androidTestImplementation libs.junit
+    androidTestImplementation "androidx.test.ext:junit:1.1.3"
```
yschimke, line 101: Will move to version catalog.

yschimke:
Landing, as test only and is passing.  The test doesn't test much apart from that we don't fail on activity pause and resume.

PR #44, opened 2022-03-25
Change order of params of WearNavScaffold

luizgrp:
#### WHAT
Change order of `startDestination` param in the constructor of WearNavScaffold

#### WHY
According to the Modifier [docs](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier): 

> Composables that accept a Modifier as a parameter to be applied to the whole component represented by the composable function should name the parameter modifier and assign the parameter a default value of Modifier. It should appear as the first optional parameter in the parameter list; after all required parameters (except for trailing lambda parameters) but before any other parameters with default values.

As `startDestination` is a required parameter, it was moved to appear before the Modifier parameter.

#### HOW
Change order of `startDestination` param in the constructor of WearNavScaffold. All the usages of this constructor in the current code are using named parameters, so there were no further changes needed.


yschimke:
Under the documentation gradle tasks there it a metalava task to update the public api signatures.

yschimke (approved):
after signature fix.

PR #46, opened 2022-03-25
Basic pager component

yschimke:
A Wear Material Compliant Pager screen.
 
 Combines the Accompanist Pager, with the Wear Compose HorizontalPageIndicator.
 Also uses lifecycle, which allows attaching logic such requesting focus
 to page events.
 
IT has one major quirk, while using the Lifecycle means that the composable can react to page events like PAUSE/RESUME similar to a top level activity, that does mean only the active page is resumed. So likely a glitch when you scroll to the next page.

We may decide this is not worth it and instead prefer explicit page events.


luizgrp (approved):

yschimke:
@JohnNichol @alexvanyo feedback welcome, but I'm using this form of it for a media app, and would like to connect it nicely with the NavScaffold.  so time for revisions.

PR #49, opened 2022-03-29
Fadeaway API test

yschimke:


yschimke:
https://user-images.githubusercontent.com/231923/160613192-ef5365b7-6900-4abc-87e8-188f69a3f753.mp4

shumelchyk:
Looks good, but I have one comment about showing timetext when there is an offset (0:07 on the video when tap 'Fade Away SLC') - should we in this such case hide timetext from the start?

yschimke:
@shumelchyk Yep - it's bad layout in the sample.  I'll fix as a follow up.

The idea is that you have to tell the Scaling Lazy List that the initial position is, it isn't exposed in the State.

```
public fun Modifier.fadeAwayScalingLazyList(
    initialIndex: Int = 1,
    initialOffset: Int = 0,
    scrollStateFn: () -> ScalingLazyListState,
): Modifier =
```

PR #56, opened 2022-04-01
API 26 for tiles

yschimke:
Fixes a build issue with tiles, which depends on 26.

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -23,7 +23,8 @@ android {
 
     defaultConfig {
         applicationId "com.google.android.horologist.sample"
-        minSdk 25
+        // Min because of Tiles
+        minSdk 26
         targetSdk 30
 
         versionCode 1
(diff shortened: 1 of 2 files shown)


shumelchyk:
So at the end Tiles require `minSDK=26`?

shumelchyk (approved):

yschimke:
> So at the end Tiles require `minSDK=26`?

I believe so https://github.com/android/wear-os-samples/blob/main/WearTilesKotlin/app/build.gradle#L27

fstanis:
This used to be the case, but now tiles is min 25: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/tiles/tiles/build.gradle;l=45

yschimke:
tiles-testing is the issue.  https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/tiles/tiles-testing/build.gradle

PR #60, opened 2022-04-05
Update audio output repository to use media router instead

wrkben:
The current implementation of `AudioOutputRepository` uses `AudioManager` to show what outputs are available and what output is connected. The implementation is assuming that if there is an A2DP device available that this is the connected output. Even if this is highly likely this is in my understanding not always correct. A media application can choose to select another media route using `androidx.mediarouter.media.MediaRouter`, even if Bluetooth is connected.

This change modifies the `AudioOutputRepository` to instead use `androidx.mediarouter.media.MediaRouter` to track what output is used. As a consequence the `AudioOutput` model can no longer have dependencies to `AudioManager` classes/data. This change also removes the dependency from the model so we're free to use whatever implementation we want.

An additional advantage of using `MediaRouter` is that any registered non-system routes would now work with the `AudioOutputRepository`. After this change they will show up as `AudioOutput#Unknown` devices, but in the future this interface can be expanded to cover more specific non-system route use-cases.

Lastly the change also contains a commits that prepares for this change. For this reason change is most likely best to review commit by commit.

### DIFF
--- a/audio-ui/src/main/java/com/google/android/horologist/audioui/VolumeViewModel.kt
+++ b/audio-ui/src/main/java/com/google/android/horologist/audioui/VolumeViewModel.kt
@@ -44,7 +44,8 @@ import kotlinx.coroutines.flow.StateFlow
 @OptIn(ExperimentalAudioApi::class)
 public open class VolumeViewModel(
     internal val volumeRepository: VolumeRepository,
-    internal val audioOutputRepository: AudioOutputRepository
+    internal val audioOutputRepository: AudioOutputRepository,
+    private val onCleared: () -> Unit = {},
 ) : ViewModel() {
     public val volumeState: StateFlow<VolumeState> = volumeRepository.volumeState
 
@@ -66,8 +67,7 @@ public open class VolumeViewModel(
     }
 
     override fun onCleared() {
-        volumeRepository.close()
-        audioOutputRepository.close()
+        onCleared.invoke()
     }
 
     @ExperimentalAudioUiApi
@@ -81,8 +81,10 @@ public open class VolumeViewModel(
             val volumeRepository = SystemVolumeRepository.fromContext(application)
             val audioOutputRepository = SystemAudioOutputRepository.fromContext(application)
 
-            // onCleared will release repositories
-            return VolumeViewModel(volumeRepository, audioOutputRepository) as T
+            return VolumeViewModel(volumeRepository, audioOutputRepository, onCleared = {
+                volumeRepository.close()
+                audioOutputRepository.close()
+            }) as T
         }
     }
 }
--- a/audio/src/main/java/com/google/android/horologist/audio/AudioOutput.kt
+++ b/audio/src/main/java/com/google/android/horologist/audio/AudioOutput.kt
@@ -26,7 +26,7 @@ public interface AudioOutput {
     /**
      * A unique audio output id.
      */
-    public val id: Int?
+    public val id: String
 
     /**
      * The user meaningful display name for the device.
@@ -38,14 +38,14 @@ public interface AudioOutput {
      */
     public object None : AudioOutput {
         override val name: String = "None"
-        override val id: Int? = null
+        override val id: String = "None"
     }
 
     /**
      * A bluetooth headset paired with the watch.
      */
     public data class BluetoothHeadset(
-        override val id: Int,
+        override val id: String,
         override val name: String
     ) : AudioOutput
 
@@ -53,7 +53,7 @@ public interface AudioOutput {
      * The one device watch speaker.
      */
     public data class WatchSpeaker(
-        override val id: Int,
+        override val id: String,
         override val name: String
     ) : AudioOutput
 
@@ -61,7 +61,7 @@ public interface AudioOutput {
      * An unknown audio output device
      */
     public data class Unknown(
-        override val id: Int,
+        override val id: String,
         override val name: String
     ) : AudioOutput
 }
--- a/audio/src/main/java/com/google/android/horologist/audio/AudioOutput.kt
+++ b/audio/src/main/java/com/google/android/horologist/audio/AudioOutput.kt
@@ -16,9 +16,6 @@
 
 package com.google.android.horologist.audio
 
-import android.media.AudioDeviceInfo
-import android.media.AudioManager
-
 /**
  * A device capable of playing audio.
  *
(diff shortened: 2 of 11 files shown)


audio-ui/src/main/java/com/google/android/horologist/audioui/VolumeViewModel.kt
```diff
@@ -81,8 +81,10 @@ public open class VolumeViewModel(
             val volumeRepository = SystemVolumeRepository.fromContext(application)
             val audioOutputRepository = SystemAudioOutputRepository.fromContext(application)
 
-            // onCleared will release repositories
-            return VolumeViewModel(volumeRepository, audioOutputRepository) as T
+            return VolumeViewModel(volumeRepository, audioOutputRepository) {
```
yschimke, line 84: tioli: onClosed seems like it deserves to be a named parameter inside the braces.

audio/src/main/java/com/google/android/horologist/audio/SystemAudioOutputRepository.kt
```diff
@@ -31,72 +30,52 @@ import kotlinx.coroutines.flow.StateFlow
...
-            if (currentAvailable.find { it.id == output.id } == null) {
-                currentAvailable.add(output)
-                added = output
-            }
-        }
+    init {
+        mediaRouter.addCallback(
+            MediaRouteSelector.Builder()
+                .addControlCategory(MediaControlIntent.CATEGORY_LIVE_AUDIO)
+                .addSelector(selector)
```
yschimke, line 64: What does https://developer.android.com/reference/androidx/mediarouter/media/MediaRouteSelector.Builder#addSelector(androidx.mediarouter.media.MediaRouteSelector) mean?

"Adds the contents of an existing media route selector to the builder."

Are you effectively filtering down the constructor parameter, which defaults to All?

audio/src/main/java/com/google/android/horologist/audio/SystemAudioOutputRepository.kt
```diff
@@ -31,72 +30,52 @@ import kotlinx.coroutines.flow.StateFlow
...
+        )
+        update()
+    }
 
-        if (added != null) {
-            _available.value = currentAvailable
-            _output.value = currentAvailable.findBluetooth()
-        }
+    private fun update() {
+        mediaRouter.fixInconsistency()
```
yschimke, line 72: Can you link to the tracking bug?

audio-ui/src/main/java/com/google/android/horologist/audioui/VolumeViewModel.kt
```diff
@@ -66,8 +67,7 @@ public open class VolumeViewModel(
     }
 
     override fun onCleared() {
-        volumeRepository.close()
-        audioOutputRepository.close()
+        onCleared.invoke()
```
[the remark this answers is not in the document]
wrkben: That would call itself, leading to infinite recursion. Wanted to keep the same name though since it was so clear. What would you prefer, keeping the name or removing `invoke()`? Any name suggestion if not `onCleared`?

audio/src/main/java/com/google/android/horologist/audio/SystemAudioOutputRepository.kt
wrkben: The other way around. There's no routes unless a control category is added. The least we would need to get speaker/bt here would be `CATEGORY_LIVE_AUDIO` that  "include local speakers and Bluetooth headsets". Other possible categories would be remote speakers, but that isn't supported. If an app would add custom media route provider, possibly a custom remote route, then those categories would have to be added too, thus the injected selector.

audio/src/main/java/com/google/android/horologist/audio/AudioOutput.kt
```diff
@@ -27,9 +24,9 @@ import android.media.AudioManager
 @ExperimentalAudioApi
 public interface AudioOutput {
     /**
-     * The id from [AudioManager] if applicable.
+     * A unique audio output id.
```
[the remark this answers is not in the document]
wrkben: For this change I would say it's unknown, so wouldn't want to be specific about it quite yet. Before the full PR it is documented as "the internal device ID". It doesn't specify the lifecycle of the ID. For the commit introducing changing this to a string this is still the case.
The ID as it ends up at the end of the PR is per route, and is documented as "an application can use this id as a token to remember the selected route across restarts or to communicate its identity to a service", but again, for this PR I wouldn't want to expand these expectations. Any suggestions on how to make it more clear, yet not too committing?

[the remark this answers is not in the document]
yschimke: Make it internal for now?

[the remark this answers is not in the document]
wrkben: Can interfaces have internals?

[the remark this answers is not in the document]
yschimke: I hate it when you point out I don't know basic kotlin things in public :)

OK, joking.  I guess it could be a class instead.  But resolving.

audio-ui/src/main/java/com/google/android/horologist/audioui/VolumeViewModel.kt
```diff
@@ -44,7 +44,8 @@ import kotlinx.coroutines.flow.StateFlow
 @OptIn(ExperimentalAudioApi::class)
 public open class VolumeViewModel(
     internal val volumeRepository: VolumeRepository,
-    internal val audioOutputRepository: AudioOutputRepository
+    internal val audioOutputRepository: AudioOutputRepository,
+    private val onCleared: () -> Unit = {},
```
luizgrp, line 48: question: any particular reason for allowing the onCleared to be provided?

luizgrp (approved):
thanks for the detailed description of the PR 👍

audio-ui/src/main/java/com/google/android/horologist/audioui/VolumeViewModel.kt
wrkben: If using something like hilt the repositories would be created by the hilt modules and would need to be removed when they go out of scope. In this case they should not be closed directly by the view model. If the view model is created by the factory in the end of the file the factory is responsible to create the repositories too and there's no scope to listen to for to close, so closing have to be tied into the view model instead. The instance doesn't know the difference, so the function is injected optionally instead.

yschimke (approved):

PR #62, opened 2022-04-06
Use debug as default again.

yschimke:
Was to make sure we ran fast on device, but it breaks previews and tests in Android Studio.


yschimke:
Landing as a simple default change.  Already made to wear-os-samples.

PR #70, opened 2022-04-07
Change volume repository to use media router

wrkben:
I believe that we need to move the volume repository to the media router too, in order to support non-system routes in the future. The api is a bit different though:

- no minimum volume, I'm assuming 0 is min
- no mute state
- the volume adjustment is based on a delta int, not sure if 1/-1 is the right way here, or a fixed size increment based on max is more appropriate

There's also a few interesting behaviours observed:
- The media router doesn't update its volume state immediately, but the new volume is returned with a callback. This is different from how the implementation using the AudioManager
- Running an app that's playing audio on bluetooth and then running another app that changes the route to the speaker doesn't affect the playing app. This makes sense to me, it's the route for the other app that's changed. However, when updating volume on the speaker route on the non-playing app the bluetooth volume is the changed one. I guess this is a matter of what app has audio focus, but I would at least expect the volume API not to work, not for it to change the incorrect device. This would have to be investigated further.


yschimke:
```
 com.google.android.horologist.audio.SystemAudioRepositoryTest > testVolumeRepository[test(AVD) - 9] FAILED 
	expected to be greater than: 11
	but was                    : 11
Tests on test(AVD) - 9 failed: There was 1 failure(s).
```

audio-ui/src/main/java/com/google/android/horologist/audioui/VolumeViewModel.kt
```diff
@@ -78,11 +77,9 @@ public open class VolumeViewModel(
 
             val application = extras[APPLICATION_KEY]!!
 
-            val volumeRepository = SystemVolumeRepository.fromContext(application)
```
yschimke, line 81: Rebase as there are changes here.

audio/README.md
```diff
@@ -7,8 +7,8 @@ For more information, visit the documentation: https://google.github.io/horologi
 Domain model for Volume and Audio Output.
 
 ```kotlin
-val volumeRepository = SystemVolumeRepository.fromContext(application)
-val audioOutputRepository = SystemAudioOutputRepository.fromContext(application)
+val volumeRepository = SystemAudioRepository.fromContext(application)
```
yschimke, line 10: remove one?

audio-ui/src/main/java/com/google/android/horologist/audioui/VolumeViewModel.kt
wrkben: done

yschimke (approved):

PR #71, opened 2022-04-11
Add different launcher icon

kul3r4:
For devices which have only icons in the menu and not chips, this makes it a bit easier to know which activity is tapped


kul3r4:
Screenshot_20220411_154803

yschimke (approved):

PR #74, opened 2022-04-12
Add PauseButton

luizgrp:
#### WHAT
Add `PauseButton`.

Screen Shot 2022-04-12 at 17 38 43

#### WHY
In order to provide common media controls.

#### HOW
- Add `PauseButton` implementation;

#### Checklist :clipboard:
- [x] Added explicit visibility modifier and explicit return types for public declarations
- [x] Updated metalava's signature text files


yschimke (approved):

yschimke:
The PreviewScreenshots are really useful.  They do make me wonder what the right style should be, I think what we have works because we have a certain background. To worry about later.

PR #81, opened 2022-04-14
Add SeekForwardButton

luizgrp:
#### WHAT
Add `SeekForwardButton`.

Screen Shot 2022-04-14 at 09 59 26

#### WHY
In order to provide common media controls.

#### HOW
- Rename `SeekBackButtonIncrement` to `SeekButtonIncrement`  and move to its own file so it can be shared by both `SeekBackButton` and `SeekForwardButton`;
- Add `SeekForwardButton` implementation;
- Add tests;

#### Checklist :clipboard:
- [x] Added explicit visibility modifier and explicit return types for public declarations
- [x] Updated metalava's signature text files


yschimke (approved):
The icon for unknown, may not be ideal, but it can be avoided so not a problem you have to solve today if there isn't a clear replacement.

PR #84, opened 2022-04-19
Add default value for enabled param

luizgrp:
#### WHAT
Add `true` as default value for `enabled` param for media ui components.

#### WHY
In order to provide convenience and to be consistent with compose libraries, as per in [Button](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/compose/compose-material/src/commonMain/kotlin/androidx/wear/compose/material/Button.kt;l=82).

#### HOW
- Add `true` as default value for `enabled` param;
- Change order of `enabled` to after `modifier` in the param list as per [guidelines](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier);
- Remove redundant values being passed to `enabled` param when matching the default value;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/api/current.api
+++ b/media-ui/api/current.api
@@ -9,19 +9,19 @@ package com.google.android.horologist.mediaui {
 package com.google.android.horologist.mediaui.components.controls {
 
   public final class MediaButtonKt {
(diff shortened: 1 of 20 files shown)


yschimke (approved):
Seems safe enough.  I'll trust your judgement on this.

PR #90, opened 2022-04-22
Pickers starting point.

yschimke:
#### WHAT

image

image

#### WHY

Provide high quality UI components

#### HOW

Extracted from https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/compose/integration-tests/demos/src/main/java/androidx/wear/compose/integration/demos/PickerDemo.kt

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
@JohnNichol these are not complete.

They have untranslated content in them, like AM/PM.  Hour, Minute, Second.

The APIs probably need to change.  We need some real tests for them, eventually screenshot tests.

But I'd like to do that incrementally, and these are marked as experimental.

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -0,0 +1,204 @@
...
+        BoxWithConstraints(modifier = Modifier.fillMaxSize()) {
+            val boxConstraints = this
+            Column(
+                verticalArrangement = Arrangement.Center,
+                horizontalAlignment = Alignment.CenterHorizontally
+            ) {
+                Spacer(Modifier.height(16.dp))
+                Text(
+                    text = when (selectedColumn) {
+                        0 -> "Day"
```
JohnNichol, line 115: nit: I am sure you have already have this on todo list - but i18n all of the strings in all of the places.

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -0,0 +1,204 @@
...
+        }
+        val monthNames = remember {
+            val months = 0..11
+            months.map {
+                // Translate month index into 3-character month string e.g. Jan.
+                // Using deprecated Date constructor rather than LocalDate in order to avoid
+                // requirement for API 26+.
+                val calendar = Calendar.getInstance()
+                calendar.set(2022, it, 1)
+                SimpleDateFormat("MMM").format(calendar.time)
```
JohnNichol, line 102: I don't know whether we need to care about i18n for formatting here - probably need some expert guidance.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -0,0 +1,331 @@
...
+public fun TimePickerWithHoursMinutesSeconds(
+    check: @Composable () -> Unit,
+    onClick: () -> Unit
+) {
+    // Omit scaling according to Settings > Display > Font size for this screen
+    val typography = MaterialTheme.typography.copy(
+        display3 = MaterialTheme.typography.display3.copy(
+            fontSize = with(LocalDensity.current) { 30.dp.toSp() }
+        )
+    )
```
JohnNichol, line 74: Does this need to be fixed size? Should we make this a param?

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -0,0 +1,331 @@
...
+                listOf(focusRequester1, focusRequester2, focusRequester3)[selectedColumn]
+                    .requestFocus()
+            }
+        }
+    }
+}
+
+@ExperimentalComposablesApi
+@Composable
+public fun TimePickerWith12HourClock(
```
JohnNichol, line 166: I wonder if we need this as a separate composable - it does have an extra button which points towards probably yes - but it feels like we should have one composable for TimePicker and then either adjust the layout if we need AM/PM button or have a second composable that adds the button and uses the first composable as-is?

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -0,0 +1,331 @@
...
+    check: @Composable () -> Unit,
+    onClick: () -> Unit
+) {
+    // Omit scaling according to Settings > Display > Font size for this screen,
+    val typography = MaterialTheme.typography.copy(
```
JohnNichol, line 171: again does this need to be fixed size font? Fixed size is necessary when the picker is large and at risk of not fitting larger screens with larger user selected font size. I guess fixed size if the safer option as we know it will fit - but it does mean we need to ensure its really usable by those who select a larger font size in settings for personal visibility reasons.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
JohnNichol, line 166: not: Kdoc

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -0,0 +1,331 @@
...
+    Spacer(Modifier.width(width))
+    Text(
+        text = ":",
+        style = textStyle,
+        color = MaterialTheme.colors.onBackground
+    )
+    Spacer(Modifier.width(width))
+}
+
+@Composable fun PickerWithRSB(
```
JohnNichol: uber-nit: We usually put @Composable on previous line

JohnNichol: do we want this to be public?

composables/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,11 @@
+<resources>
```
JohnNichol, line 1: This doesn't seem to be related to this patch

JohnNichol (changes requested):
Lots of suggestions and ideas here - but overall this is looking good. We should think about how we want many API 'things' we want here - I am not against having descriptive and opinionated implementations, but I think we should look to have a common core internal implementation for each things and then relatively thin wrapper helpers around them.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
yschimke: Already fixed

```
@Composable internal fun PickerWithRSB(
```

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
yschimke: Will follow up.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
yschimke: Will follow up.

yschimke:
As discussed offline, will land and either myself or someone on the wear-compose team addresses the review comments in individual PRs.

Issue #91, opened 2022-04-22
Picker a11y tests


yschimke:
Follow up on https://github.com/google/horologist/pull/90

- [x] provide way of getting the value
- [x] minimal tests
- [x] API for onChange without submitting
- [x] Remove any non localised content.  AM/PM.  hours, minutes, seconds.
- [x] Review APIs
- [ ] ~Screenshot tests~
- [x] Theming, customising (date range, error states)
- [ ] a11y review + tests

yschimke:
cc @JohnNichol

PR #93, opened 2022-04-25
Add test and output method for Pickers

yschimke:
#### WHAT

Towards https://github.com/google/horologist/issues/91

#### WHY

Needs test and example of getting a value out.

#### HOW

Add Compose test and verify that date is returned.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
+++ b/composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@file:OptIn(ExperimentalComposablesApi::class)
+
+package com.google.android.horologist
+
+import androidx.compose.foundation.layout.size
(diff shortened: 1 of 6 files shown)


yschimke:
cc @BowerSteve

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
```diff
@@ -0,0 +1,70 @@
...
+                onClick = {
+                    date = it
+                },
+                initial = LocalDate.of(2022, 4, 25)
+            )
+        }
+
+        composeTestRule.onNodeWithContentDescription("check").performClick()
+
+        assertThat(date).isEqualTo(LocalDate.of(2022, 4, 25))
```
luizgrp, line 68: I think this test might be flaky as this line can run on the test runner thread before the `date` value is changed by the code in the [onClick](https://github.com/google/horologist/pull/93/files#diff-0272f15b9a90efe839b42c06db185db595f7da78cdeedd9bbfd848b0f7a53c43R60) running on the main thread.  

<details>
 <summary>Similarly to what I've seen in #92</summary>
Caveat: after merging that PR to main, the test that was supposed to be fixed, surprisingly <a href="https://github.com/google/horologist/runs/6153847717?check_suite_focus=true">failed</a>. There is no information about the failure in the archived logcat and test report. So please assess my thoughts here, as maybe that the test was never flaky, or there are still other issues that are making the tests fail randomly.
</details>

If my thoughts here make sense, then we should take the same approach with [waitUntil](https://developer.android.com/jetpack/compose/testing#manual-sync).

Alternatively, we could take advantage of the fact that `ComposeTestRule` [waits](https://developer.android.com/jetpack/compose/testing#sync) for the UI tree to be idle and change the code to:
1) change type of `date` to `mutableStateOf<LocalDate>` so compose tracks when its value is changed
2) display the content of `date` in a `Text` inside `setContent`
3) use `ComposeTestRule` to assert that the text of `Text` was changed to the value of `date`

luizgrp (approved):

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
yschimke: Makes sense.

PR #97, opened 2022-04-29
Add module for Media Data library

luizgrp:
#### WHAT

Add module for Media Data library

#### WHY

In order to include data components related to media and publish them as an individual library.

#### HOW

- Add a new android ui module with same setup as other modules from the project;
- Add maven publish plugin configuration to the module;
- Add annotation `ExperimentalMediaDataApi`;
- Add metalava generated API file;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-data/README.md
+++ b/media-data/README.md
@@ -1,0 +1,1 @@
+# Media Data library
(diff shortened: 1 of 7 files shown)


yschimke:
I wasn't expecting to need this yet, figured it could be a package in the same module.  But sounds fine, let's see where it goes.

yschimke (approved):

PR #99, opened 2022-04-29
Add PlayerUiState

luizgrp:
#### WHAT

Add `PlayerUiState`.

#### WHY

In order to have a class to represent the state of the Media UI components.

#### HOW
- Add `PlayerUiState`;
- Add `MediaItemUiModel`;
- Add `TrackPositionUiModel`;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/MediaItemUiModel.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/MediaItemUiModel.kt
@@ -1,0 +1,22 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.media.ui.state.model
+
+public data class MediaItemUiModel(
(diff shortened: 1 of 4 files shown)


luizgrp:
cc: @dpreussler

media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/MediaItemUiModel.kt
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.media.ui.state.model
+
+public data class MediaItemUiModel(
+    val title: String? = null,
+    val artist: String? = null
```
yschimke, line 21: No problem for now.  But we should consider how this expands.

Do we eventually want fields like album? Is it safe as a data class?

media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/TrackPositionUiModel.kt
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.media.ui.state.model
+
+public data class TrackPositionUiModel(
+    val current: Long,
```
yschimke, line 20: I do wonder whether Long is right type here, although not sure Duration works well for current.

yschimke (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/TrackPositionUiModel.kt
luizgrp: tracked in b/228414169

PR #100, opened 2022-04-29
Add UI mappers

luizgrp:
#### WHAT

Add mappers to map models from the data layer into models from the UI layer.

#### WHY

Reduce amount of code in the view model that would have to map those models across layers.
Make it easier to unit test.

#### HOW

- Add dependency on `media-data` and `media3-common` to `media-ui`;
- Add `PlayerUiStateMapper`;
- Add `MediaItemUiModelMapper`;
- Add `TrackPositionUiModelMapper`;
- Add unit tests;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/state/mapper/PlayerUiStateMapper.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/state/mapper/PlayerUiStateMapper.kt
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.media.ui.state.mapper
+
+import androidx.media3.common.MediaItem
+import androidx.media3.common.Player
+import com.google.android.horologist.media.data.model.TrackPosition
+import com.google.android.horologist.media.ui.state.PlayerUiState
+
+/**
+ * Map [Player.Commands] plus other set of properties into a [PlayerUiState]
+ */
+public object PlayerUiStateMapper {
+
+    public fun map(
+        playerCommands: Player.Commands,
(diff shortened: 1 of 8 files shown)


luizgrp:
cc: @dpreussler

media-ui/src/main/java/com/google/android/horologist/media/ui/state/mapper/PlayerUiStateMapper.kt
```diff
@@ -0,0 +1,53 @@
...
+
+        return PlayerUiState(
+            playEnabled = playPauseCommandAvailable,
+            pauseEnabled = playPauseCommandAvailable,
+            seekBackEnabled = playerCommands.contains(Player.COMMAND_SEEK_BACK),
+            seekForwardEnabled = playerCommands.contains(Player.COMMAND_SEEK_FORWARD),
+            seekToPreviousEnabled = playerCommands.contains(Player.COMMAND_SEEK_TO_PREVIOUS_MEDIA_ITEM),
+            seekToNextEnabled = playerCommands.contains(Player.COMMAND_SEEK_TO_NEXT_MEDIA_ITEM),
+            shuffleEnabled = playerCommands.contains(Player.COMMAND_SET_SHUFFLE_MODE),
+            shuffleOn = shuffleEnabled,
```
yschimke, line 46: this is unfortunate and confusing.  The map param and the constructor param are swapped in meaning.

yschimke (approved):
one comment on naming.

media-ui/src/main/java/com/google/android/horologist/media/ui/state/mapper/PlayerUiStateMapper.kt
luizgrp: agreed, will address it in the next PR 👍 

the map param follows the exposed property in PlayerRepository, so I'm thinking in changing it there to `shuffleModeEnabled` as it aligns to the property in media3.Player.

It will make it clearer but still a bit confusing:
```
            shuffleEnabled = playerCommands.contains(Player.COMMAND_SET_SHUFFLE_MODE),
            shuffleOn = shuffleModeEnabled,
```

Another option would be to change PlayerUiState properties to add a "Button" name in it, e.g. `shuffleButtonEnabled`:
```
            shuffleButtonEnabled = playerCommands.contains(Player.COMMAND_SET_SHUFFLE_MODE),
            shuffleOn = shuffleModeEnabled,
```

Any other suggestions?

yschimke: No, I'm bad at naming.

luizgrp: addressed in https://github.com/google/horologist/pull/101

PR #101, opened 2022-04-29
Add PlayerViewModel

luizgrp:
#### WHAT

Add `PlayerViewModel`

#### WHY

In order to provide a [view model as source of truth](https://developer.android.com/jetpack/compose/state#viewmodels-source-of-truth) to this module's composables.

#### HOW

- Add `PlayerViewModel`;
- Add `StubPlayerRepository` test double;
- Add unit tests;
- Add dependency on `lifecycle-viewmodel-ktx` to `media-ui`;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Nice!

luizgrp:
cc: @dpreussler

PR #104, opened 2022-05-03
Add stateful PauseButton

luizgrp:
#### WHAT

Add stateful `PauseButton`

#### WHY

In order to provide a version of the UI components with default behaviour, using `PlayerViewModel` and `PlayerUiState`.

#### HOW

- Add stateful implementation of `PauseButton`;
- Add tests;
- Add util function `rememberStateWithLifecycle`;
- Add test double `FakePlayerRepository` - it should be moved in another iteration to a separate test fixtures module;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/api/current.api
+++ b/media-ui/api/current.api
@@ -31,6 +31,7 @@ package com.google.android.horologist.media.ui.components.controls {
   }
 
   public final class PauseButtonKt {
+    method @androidx.compose.runtime.Composable @com.google.android.horologist.media.ui.ExperimentalMediaUiApi public static void PauseButton(com.google.android.horologist.media.ui.state.PlayerViewModel playerViewModel, optional androidx.compose.ui.Modifier modifier, optional androidx.wear.compose.material.ButtonColors colors);
     method @androidx.compose.runtime.Composable @com.google.android.horologist.media.ui.ExperimentalMediaUiApi public static void PauseButton(kotlin.jvm.functions.Function0<kotlin.Unit> onClick, optional androidx.compose.ui.Modifier modifier, optional boolean enabled, optional androidx.wear.compose.material.ButtonColors colors);
   }
 
@@ -215,6 +216,13 @@ package com.google.android.horologist.media.ui.state.model {
     property public final long current;
     property public final long duration;
     property public final float percent;
+  }
+
+}
+
+package com.google.android.horologist.media.ui.utils {
+
+  public final class StateWithLifecycleKt {
   }
 
 }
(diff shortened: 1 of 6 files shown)


media-ui/api/current.api
```diff
@@ -219,3 +220,10 @@ package com.google.android.horologist.media.ui.state.model {
 
 }
 
+package com.google.android.horologist.media.ui.utils {
+
+  public final class StateWithLifecycleKt {
```
luizgrp, line 225: hmm maybe we should place that function inside a class marked as internal so this doesn't get exposed

luizgrp:
cc: @dpreussler

media-ui/api/current.api
yschimke: Yep - that makes sense.  We shouldn't accidentally publish util libraries.

media-ui/src/main/java/com/google/android/horologist/media/ui/components/controls/PauseButton.kt
```diff
@@ -19,12 +19,32 @@ package com.google.android.horologist.media.ui.components.controls
...
 import androidx.wear.compose.material.ButtonDefaults
 import com.google.android.horologist.media.ui.ExperimentalMediaUiApi
 import com.google.android.horologist.media.ui.R
+import com.google.android.horologist.media.ui.state.PlayerViewModel
+import com.google.android.horologist.media.ui.utils.rememberStateWithLifecycle
+
+@ExperimentalMediaUiApi
+@Composable
+public fun PauseButton(
+    playerViewModel: PlayerViewModel,
```
yschimke, line 35: This was the bit I wasn't sure about.  Land but, let's see how it works as screens compose. And I realise I was basically arguing for this before, with Repository instead of ViewModel.

Downsides - not sure how you build a stateless screen from these. And we end up with a lot of rememberStateWithLifecycle per screen.

media-ui/src/main/java/com/google/android/horologist/media/ui/utils/StateWithLifecycle.kt
```diff
@@ -0,0 +1,41 @@
...
+import androidx.compose.runtime.State
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.remember
+import androidx.compose.ui.platform.LocalLifecycleOwner
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.flowWithLifecycle
+import kotlinx.coroutines.flow.StateFlow
+
+@Composable
+internal fun <T> rememberStateWithLifecycle(
```
yschimke, line 29: Good idea on internal.

yschimke (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/components/controls/PauseButton.kt
luizgrp: hmm re the stateless screen, the stateless PauseButton defined in this file should be used?

PR #109, opened 2022-05-04
Extract string resources

luizgrp:
#### WHAT

Extract hard coded strings into string resources file.

#### WHY

So they can be maintained independently and alternative resources can be provided.

#### HOW

- Add strings resource file to `audio-ui` module;
- Change `VolumeScreen` to retrieve content description from the resources file;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Thanks.

PR #112, opened 2022-05-05
Fix docs for github pages.

yschimke:
#### WHAT

Github pages docs were missing content.

#### HOW

Define docs in docs/ for each module.


audio-ui/README.md
```diff
@@ -1,14 +1,8 @@
 # Audio Settings UI library
 
-[![Maven Central](https://img.shields.io/maven-central/v/com.google.android.horologist/horologist-compose-layout)](https://search.maven.org/search?q=g:com.google.android.horologist)
+[![Maven Central](https://img.shields.io/maven-central/v/com.google.android.horologist/horologist-audio-ui)](https://search.maven.org/search?q=g:com.google.android.horologist)
```
fstanis, line 3: While we're at it... I wonder if we should link to https://search.maven.org/artifact/com.google.android.horologist/horologist-audio-ui or https://ossindex.sonatype.org/component/pkg:maven/com.google.android.horologist/horologist-audio-ui instead?

yschimke: No opinion. but I'll leave for a follow up. 

I'd link to https://mvnrepository.com/artifact/com.google.android.horologist

PR #117, opened 2022-05-05
Improve stateful PlayScreen

luizgrp:
#### WHAT

Improve stateful version of `PlayScreen`.

#### WHY

In order to provide a version with default implementation of `mediaDisplay` and `controlButtons` slots.

#### HOW

- Add default implementation of `mediaDisplay` slot with `TextMediaDisplay`;
- Add default implementation of `controlButtons` slot with `MediaControlButtons`;
- Add `lifecycle-runtime-ktx` dependency to `media-ui` module;
- Add tests;
- Add util function `rememberStateWithLifecycle`;
- Add test double `FakePlayerRepository` - it should be moved in another iteration to a separate test fixtures module;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp:
cc: @dpreussler

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/PlayScreen.kt
```diff
@@ -28,31 +28,96 @@ import androidx.compose.foundation.layout.fillMaxSize
...
+    buttons: (@Composable RowScope.() -> Unit)? = null,
+    background: @Composable BoxScope.() -> Unit = {}
 ) {
+    val playerUiState by rememberStateWithLifecycle(flow = playerViewModel.playerUiState)
+
     PlayScreen(
-        mediaDisplay = { mediaDisplay(playerViewModel) },
-        controlButtons = { controlButtons(playerViewModel) },
-        buttons = { buttons(playerViewModel) },
+        mediaDisplay = mediaDisplay ?: {
```
yschimke, line 55: Curious about this form, you can't move this default to the param, because it accesses playerUiState.  But that makes it strictly harder to implement an override, should you pass in playerUiState to these composables?

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/PlayScreen.kt
```diff
@@ -28,31 +28,96 @@ import androidx.compose.foundation.layout.fillMaxSize
...
+    onPauseClick: () -> Unit,
+    playPauseEnabled: Boolean,
+    playing: Boolean,
+    percent: Float,
+    onSeekToPreviousButtonClick: () -> Unit,
+    seekToPreviousButtonEnabled: Boolean,
+    onSeekToNextButtonClick: () -> Unit,
+    seekToNextButtonEnabled: Boolean,
+) {
+    if (showProgress) {
```
yschimke, line 95: This feels strange, you branch out here based on showProgress.  But then the private method in MediaControlButtons takes both branches in again.

Why has PlayScreen(showDefault)? why not just choose a default, say with progress, and allow callers to customise the control buttons to remove it.

yschimke (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/PlayScreen.kt
luizgrp: If we pass the `playerUiState` then my understanding is that it won't recompose once a new `PlayerUiState` is emitted by the ViewModel, did you mean to pass a `StateFlow<PlayerUiState>`?

> that makes it strictly harder to implement an override

Could you please elaborate further on this?

yschimke: Changing the param to

```
mediaDisplay: (@Composable ColumnScope.(PlayerUiState) -> Unit)? = null,
```

It should recompose, if it doesn't it's a bug with compose or our state objects.  But usually the bug is reversed, if you compile State classes in modules without compose it doesn't get the tracking right, so it recomposes *too* frequently.

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/PlayScreen.kt
luizgrp: It does feel strange, I'm glad you raised it for discussion. The reason of doing this way is because I initially thought it would feel right to only use the public APIs of `MediaControlButtons` here, but we could change that private method in `MediaControlButtons` to internal, use it, and remove this function completely. I'm keen to do that, if you are also happy with it?

The reason for exposing the `showProgress` in `PlayScreen` is to allow clients to use as:
```kotlin
PlayScreen(
    playerViewModel = playerViewModel,
    showProgress = false,
)
```

Rather than the below for the same result:
```kotlin
val playerUiState by rememberStateWithLifecycle(flow = playerViewModel.playerUiState)

PlayScreen(
    playerViewModel = playerViewModel,
    controlButtons = {
            MediaControlButtons(
                onPlayClick = { playerViewModel.prepareAndPlay() },
                onPauseClick = { playerViewModel.pause() },
                playPauseEnabled = playerUiState.playPauseEnabled,
                playing = playerUiState.playing,
                percent = playerUiState.trackPosition?.percent ?: 0f,
                onSeekToPreviousButtonClick = { playerViewModel.seekToPreviousMediaItem() },
                seekToPreviousButtonEnabled = playerUiState.seekToPreviousEnabled,
                onSeekToNextButtonClick = { playerViewModel.seekToNextMediaItem() },
                seekToNextButtonEnabled = playerUiState.seekToNextEnabled,
                showProgress = false,
            )
    },
)
```
Please let me know if I'm missing anything?

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/PlayScreen.kt
```diff
@@ -28,31 +28,147 @@ import androidx.compose.foundation.layout.fillMaxSize
...
+
+    @Composable
+    public fun content(playerUiState: PlayerUiState): @Composable RowScope.() -> Unit
+}
+
+/**
+ * Represents the content of media display on [PlayScreen].
+ */
+@ExperimentalMediaUiApi
+public interface PlayScreenMediaDisplay {
```
yschimke, line 83: Should these be "fun interface"? to simplify inline usage?

yschimke (approved):
One suggesting to use fun interfaces.

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/PlayScreen.kt
luizgrp: yeah, I tried it but when simplifying the usage I'm facing this weird issue:

> Type inference failed. Expected type mismatch: inferred type is (PlayerUiState) -> @Composable() (ColumnScope.() -> Unit) but (PlayerUiState) -> @Composable() (ColumnScope.() -> Unit) was expected

Please let me know if you know what it is, otherwise I will come back to it at another time

yschimke: Come back to it.

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/PlayScreen.kt
```diff
@@ -28,31 +28,147 @@ import androidx.compose.foundation.layout.fillMaxSize
...
+}
+
+/**
+ * Represents the content of control buttons on [PlayScreen].
+ */
+@ExperimentalMediaUiApi
+public interface PlayScreenControlButtons {
+
+    @Composable
+    public fun content(playerUiState: PlayerUiState): @Composable RowScope.() -> Unit
```
yschimke, line 76: Actually, I think this isn't how I would expect. I wouldn't return a composable here.

Maybe instead of an interface, use 

```
typealias PlayScreenControlButtons = @Composable ((PlayerUiState) -> Unit)
```

luizgrp: thanks, addressed in 151c97617889a8c5871c163a32bf345d93257501

yschimke (approved):
LGTM - I'm still thinking about PlayScreenDefaults.customControlButtons, but not a blocker.

PR #132, opened 2022-05-10
Add stateful PlayScreen guide

luizgrp:
#### WHAT

Add stateful `PlayScreen` guide to docs.

#### Checklist :clipboard:
- [N/A] Run spotless check
- [N/A] Run tests


docs/media-playscreen.md
```diff
@@ -0,0 +1,39 @@
+# Stateful PlayScreen guide
+
+This is a guide on how to use the stateful `PlayScreen` with your own implementation of `PlayerRepository`.
+
+## Basic usage
+
+1. Implement PlayerRepository
```
yschimke, line 7: Maybe describe what is required here typically.

For example you could call out Media3 Player, and how state could be mapped via a listener and a MutableState field?

Operations would be called on the player.

Maybe show implementations of two of these?

```
    var isPlaying = MutableStateFlow(false)

    private val listener = object : Player.Listener {
        override fun onIsPlayingChanged(isPlaying: Boolean) {
            updatePlayingState()
        }
    }
    
    player.addListener(listener)

    fun setPlaybackSpeed(speed: Float) {
        player.setPlaybackSpeed(speed)
    }
```

yschimke (approved):
Looks great!

docs/media-playscreen.md
luizgrp: cool, I will add it in a separate PR 👍

luizgrp: done in https://github.com/google/horologist/pull/134

PR #139, opened 2022-05-10
Add settings buttons into the media sample

yschimke:
#### WHAT

Add settings buttons in.  Just as sample, can delete later once we have more properly supported functionality.

#### WHY

Removes the gap in the PlayerScreen sample.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/sample/src/main/java/com/google/android/horologist/sample/media/SettingsButtons.kt
+++ b/sample/src/main/java/com/google/android/horologist/sample/media/SettingsButtons.kt
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


yschimke:
image

sample/src/main/java/com/google/android/horologist/sample/media/SettingsButtons.kt
```diff
@@ -0,0 +1,54 @@
...
+    val volumeState = volumeViewModel.volumeState.collectAsState().value
+
+    Row(modifier = modifier) {
+        SetVolumeButton(
+            onVolumeClick = onVolumeClick,
+            volumeState = volumeState
+        )
+        Spacer(modifier = Modifier.size(16.dp))
+        AudioOutputButton(
+            onOutputClick = { context.launchBluetoothSettings() }
```
luizgrp, line 51: nit: I think it should be MainActivity's responsibility to navigate to this as I see that's where the navigation of the app is being done

luizgrp (approved):

luizgrp (approved):

PR #140, opened 2022-05-11
Shrink volume chip width

yschimke:
#### WHAT

image

#### WHY

UX Design wants no trailing whitespace.

#### HOW

Use intrinsic width.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


alexvanyo:
Yay screenshot tests!

PR #146, opened 2022-05-12
Androidx RC Updates

yschimke:
#### WHAT

RCs are available for many Android libraries.

#### WHY

We should be looking at adopting the latest versions of dependencies, and we are beta for Compose.  These RCs seem appropriate.

#### HOW

./gradlew versionCatalogUpdate

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


shumelchyk (approved):
LGTM

luizgrp (approved):

PR #147, opened 2022-05-12
Multi previews

yschimke:
#### WHAT

Compose multi device previews.

#### WHY

Simply common preview setup.

#### HOW

New annotation.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
This module will also be a good home for a Theme switcher, to test out the media ui screens.

yschimke:
image

compose-testing/src/main/java/com/google/android/horologist/compose/testing/WearPreviewDevices.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import androidx.compose.ui.tooling.preview.Devices
+import androidx.compose.ui.tooling.preview.Preview
+
+@Preview(
+    device = Devices.WEAR_OS_LARGE_ROUND,
+    showSystemUi = true,
+    backgroundColor = 0xff000000,
+    showBackground = true,
+    name = "Large Round"
```
luizgrp, line 27: should we use the `group` attribute here instead of `name`?

so the users of this annotation can still write their specific name to their previews, but they will be grouped by Large/Small/Square

I'm suggesting but I'm not sure if it works like this as well, would need to check it

luizgrp (approved):
nice one!

compose-testing/src/main/java/com/google/android/horologist/compose/testing/WearPreviewDevices.kt
yschimke: Can you check, and then make a follow up PR.  I can't understand how someone using this annotation would set a name.

yschimke: Actually I'll do it here, Group works really nicely for this.

PR #153, opened 2022-05-13
Add PodcastControlButtons

luizgrp:
#### WHAT

Add `PodcastControlButtons`.

Screen Shot 2022-05-12 at 22 26 30

#### WHY

In order to provide common media components.

#### HOW

- Add `PodcastControlButtons` implementation;
- Add tests;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/components/PodcastControlButtons.kt
```diff
@@ -0,0 +1,184 @@
...
+@ExperimentalHorologistMediaUiApi
+@Composable
+public fun PodcastControlButtons(
+    onPlayButtonClick: () -> Unit,
+    onPauseButtonClick: () -> Unit,
+    playPauseButtonEnabled: Boolean,
+    playing: Boolean,
+    percent: Float,
+    onSeekBackButtonClick: () -> Unit,
+    seekBackButtonIncrement: SeekButtonIncrement,
```
yschimke, line 51: Should we default these to Unknown? for a slightly simpler starting API? likewise the seek forward enabled buttons? I know it's technically wrong, but might be a more gradual adoption path while you work out the API?

Just a suggestion that I'm not sure we want...

media-ui/src/main/java/com/google/android/horologist/media/ui/components/PodcastControlButtons.kt
```diff
@@ -0,0 +1,184 @@
...
+    seekBackButtonEnabled: Boolean,
+    onSeekForwardButtonClick: () -> Unit,
+    seekForwardButtonIncrement: SeekButtonIncrement,
+    seekForwardButtonEnabled: Boolean,
+    showProgress: Boolean,
+    modifier: Modifier = Modifier,
+    percent: Float? = null,
+    colors: ButtonColors = ButtonDefaults.iconButtonColors(),
+    progressColour: Color = MaterialTheme.colors.primaryVariant,
```
yschimke, line 133: Can you check these against the UX spec as we move these? Definite follow up, land first, but avoids rework.

yschimke (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/components/PodcastControlButtons.kt
luizgrp: sounds good, but then it will default to the "not ideal" icon for the `SeekForwardButton` https://github.com/google/horologist/pull/81#pullrequestreview-942031720

media-ui/src/main/java/com/google/android/horologist/media/ui/components/PodcastControlButtons.kt
luizgrp: cool, will do at some point 👍 

I also noticed that this component doesn't set the button size as [done](https://github.com/google/horologist/blob/f2a6eae5a8b8a7b8b6ec5e420a44f500b61e34b4/media-ui/src/main/java/com/google/android/horologist/media/ui/components/MediaControlButtons.kt#L146) for the equivalent component `MediaControlButtons` so I would like to review them and understand

media-ui/src/main/java/com/google/android/horologist/media/ui/components/PodcastControlButtons.kt
luizgrp: done and the above is tracked in https://github.com/google/horologist/issues/157

PR #162, opened 2022-05-13
Picker minor code tidy up

JohnNichol:
#### WHAT
Add missing KDoc
Replace hardcoded strings with stringResources.

#### WHY
To tidy up the picker code

#### Checklist :clipboard:
- [X] Run spotless check
- [ ] Run tests

### DIFF
--- a/composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
@@ -61,6 +61,7 @@ import androidx.wear.compose.material.PickerState
 import androidx.wear.compose.material.Text
 import androidx.wear.compose.material.rememberPickerState
 import java.time.LocalTime
+import java.time.format.DateTimeFormatter
 import java.time.temporal.ChronoField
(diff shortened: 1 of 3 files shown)


yschimke (approved):
Thanks!

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -249,7 +262,8 @@ public fun TimePickerWith12HourClock(
                         horizontalArrangement = Arrangement.Center
                     ) {
                         Text(
-                            text = if (amPm == 0) "AM" else "PM",
+                            text = if (amPm == 0) stringResource(R.string.time_picker_am)
```
yschimke, line 265: Should we avoid needing to localise with

```
    val pmString = remember { LocalTime.of(18, 0).format(DateTimeFormatter.ofPattern("a"))}
```

JohnNichol: Why not - added

PR #169, opened 2022-05-16
Add Media module

luizgrp:
#### WHAT

Add module for Media domain library.

#### WHY

In order to include domain code related to media and publish them as an individual library.

#### HOW
- Add a new java/kotlin module;
- Add maven publish plugin configuration to the module;
- Add annotation `ExperimentalHorologistMediaApi`;
- Add metalava generated API file;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,10 +39,12 @@ buildscript {
 }
 
 plugins {
+
     id "com.diffplug.spotless" version "5.9.0"
     id 'org.jetbrains.kotlin.android' version '1.6.20' apply false
     id "com.github.ben-manes.versions" version "0.41.0"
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -39,10 +39,12 @@ buildscript {
 }
 
 plugins {
+
     id "com.diffplug.spotless" version "5.9.0"
     id 'org.jetbrains.kotlin.android' version '1.6.20' apply false
```
yschimke, line 44: We should bump these to 1.6.21.

In future use alias also https://github.com/google/horologist/issues/170

yschimke (approved):

PR #173, opened 2022-05-16
Add PlayerRepository to media module

luizgrp:
#### WHAT

Add `PlayerRepository` and domain models to `media` module.

#### WHY

In order to have a `PlayerRepository` that is independent of [Media3](https://developer.android.com/jetpack/androidx/releases/media3) dependencies.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media/src/main/java/com/google/android/horologist/media/model/Command.kt
+++ b/media/src/main/java/com/google/android/horologist/media/model/Command.kt
@@ -23,7 +23,7 @@ public enum class Command {
     PlayPause,
     SeekBack,
     SeekForward,
-    SeekToPreviousMediaItem,
-    SeekToNextMediaItem,
+    SkipToPreviousMediaItem,
+    SkipToNextMediaItem,
     SetShuffle
 }
--- a/media/src/main/java/com/google/android/horologist/media/model/Command.kt
+++ b/media/src/main/java/com/google/android/horologist/media/model/Command.kt
@@ -14,21 +14,16 @@
  * limitations under the License.
  */
 
-plugins {
+package com.google.android.horologist.media.model
 
-    id 'java-library'
-    id 'org.jetbrains.kotlin.jvm'
-    id 'org.jetbrains.dokka'
+/**
+ * Commands that can be executed on a player.
+ */
+public enum class Command {
+    PlayPause,
+    SeekBack,
+    SeekForward,
+    SeekToPreviousMediaItem,
+    SeekToNextMediaItem,
+    SetShuffle
 }
-
-java {
-    sourceCompatibility = JavaVersion.VERSION_1_7
-    targetCompatibility = JavaVersion.VERSION_1_7
-}
-
-dependencies {
-
-    implementation libs.kotlinx.coroutines.core
-}
-
-apply plugin: "com.vanniktech.maven.publish"
--- a/media/src/main/java/com/google/android/horologist/media/model/MediaItem.kt
+++ b/media/src/main/java/com/google/android/horologist/media/model/MediaItem.kt
@@ -21,5 +21,5 @@ package com.google.android.horologist.media.model
  */
 public data class MediaItem(
     val title: String? = null,
-    val artist: String? = null
+    val artist: String
 )
--- a/media/src/main/java/com/google/android/horologist/media/model/MediaItem.kt
+++ b/media/src/main/java/com/google/android/horologist/media/model/MediaItem.kt
@@ -14,21 +14,12 @@
  * limitations under the License.
  */
 
-plugins {
+package com.google.android.horologist.media.model
 
-    id 'java-library'
-    id 'org.jetbrains.kotlin.jvm'
-    id 'org.jetbrains.dokka'
-}
-
-java {
-    sourceCompatibility = JavaVersion.VERSION_1_7
-    targetCompatibility = JavaVersion.VERSION_1_7
-}
-
-dependencies {
-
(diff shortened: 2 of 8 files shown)


media/src/main/java/com/google/android/horologist/media/model/MediaItem.kt
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.media.model
+
+/**
+ * Representation of a media item.
+ */
+public data class MediaItem(
```
yschimke, line 22: Should we avoid MediaItem just because of it's existing use? Not sure of the right name here.

media/src/main/java/com/google/android/horologist/media/model/MediaItem.kt
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.media.model
+
+/**
+ * Representation of a media item.
```
yschimke, line 20: This representation seems uniquely focused on the UI. Should we have support for id (for id equality?), url, artwork?

I guess the question here is can the docs clear up the domain constraints for this type? So we can answer the above.

media/src/main/java/com/google/android/horologist/media/model/MediaItem.kt
luizgrp: sorry, not sure I follow what do you mean by "of it's existing use", could you please clarify?

media/src/main/java/com/google/android/horologist/media/model/MediaItem.kt
yschimke: I guess it also presents a question for how to connect a backend source populating fields we didn't consider, like lyrics url, with a frontend that wants to show them.  Should we consider some properties mechanism, or subclasses, or some other design approach?

media/src/main/java/com/google/android/horologist/media/model/MediaItemPosition.kt
```diff
@@ -0,0 +1,28 @@
...
+
+package com.google.android.horologist.media.model
+
+/**
+ * Represents the current [media item][MediaItem] position, duration and percent progress.
+ * Current position and duration are measured in milliseconds.
+ */
+public data class MediaItemPosition(
+    val current: Long,
+    val duration: Long,
```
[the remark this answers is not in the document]
luizgrp: addressed in https://github.com/google/horologist/pull/173/commits/90ec861252c0e8852dd85e9d78b732cdf2afbedd

media/src/main/java/com/google/android/horologist/media/model/Command.kt
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Commands that can be executed on a player.
+ */
+public enum class Command {
+    PlayPause,
+    SeekBack,
+    SeekForward,
+    SeekToPreviousMediaItem,
+    SeekToNextMediaItem,
```
dpreussler, line 27: not happy about the naming. skipping tracks and seek seem to be 2 different things

media/src/main/java/com/google/android/horologist/media/model/MediaItemPosition.kt
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Represents the current [media item][MediaItem] position, duration and percent progress.
+ * Current position and duration are measured in milliseconds.
+ */
+public sealed class MediaItemPosition(
+    public open val current: Long,
+) {
+    public class KnownDuration internal constructor(
+        override val current: Long,
+        public val duration: Long,
```
dpreussler, line 28: why not use Kotlins `Duration` for Duration? ;)

media/src/main/java/com/google/android/horologist/media/model/MediaItem.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+public data class MediaItem(
+    val title: String? = null,
+    val artist: String? = null
```
dpreussler, line 24: what is a media item that has neither title nor artist. Is this sth we want to support?

luizgrp: @yschimke please see the above, I ported this way to follow the implementation in the media toolkit, was there any use case for that in mind?

luizgrp: @dpreussler what would be your suggestion? 

a) non-nullable title?
b) non-nullable title and artist?
c) keep both them nullable but enforce clients to provide one or the other via factory methods?

please let me know if you have a different idea in mind

yschimke: If we think in 95% of apps using this code we should avoid the hassle of nullability and require them.  Maybe just a note that empty "", should be used if there legitimately isn't one.

There can be multiple artists (track artist, album artist) perhaps we should make this clear it's whatever we want for display? displayArtist.  But I don't know, so happy to go with non nullable artist for now.

PR #174, opened 2022-05-17
Picker API review

yschimke:
#### WHAT

Update Pickers API.

#### WHY

Follow idiomatic compose conventions.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

#### Later
- [ ] Confirm a11y for AM/PM
- [ ] Add chevron and next for date picker 
- [ ] CompoundPicker and State for a reusable Picker base.

### DIFF
--- a/composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
@@ -197,12 +197,12 @@ public fun DatePicker(
                 )
                 Button(
                     onClick = {
-                        val date = LocalDate.of(
+                        val confirmedDate = LocalDate.of(
                             yearState.selectedOption + 1,
                             monthState.selectedOption + 1,
                             dayState.selectedOption + 1
                         )
-                        onDateConfirm(date)
+                        onDateConfirm(confirmedDate)
                     },
                 ) {
                     Icon(
--- a/composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
@@ -17,7 +17,6 @@
 package com.google.android.horologist.composables
 
 import androidx.compose.foundation.layout.Arrangement
-import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.BoxWithConstraints
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
@@ -28,6 +27,9 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.offset
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.wrapContentSize
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Check
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.derivedStateOf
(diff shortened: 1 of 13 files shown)


composables/src/main/res/values/strings.xml
```diff
@@ -22,4 +22,5 @@
     <string name="picker_day">Day</string>
     <string name="picker_month">Month</string>
     <string name="picker_year">Year</string>
+    <string name="picker_check_button">Check</string>
```
yschimke, line 25: Is "Check" right? Confirm?

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -201,10 +202,16 @@ public fun DatePicker(
...
                             dayState.selectedOption + 1
                         )
-                        onClick(date)
+                        onValueConfirm(date)
                     },
                 ) {
-                    buttonIcon()
+                    Icon(
+                        imageVector = Icons.Filled.Check,
+                        contentDescription = stringResource(id = R.string.picker_check_button),
```
ssancho14, line 210: The strings namespace is shared between the app and all libraries, right? May make sense to prefix with the library name?

composables/src/main/res/values/strings.xml
ssancho14: Not sure. I think I prefer Confirm, to match `onValueConfirm`. Other alternatives: Accept, Ok

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -73,18 +76,16 @@ import java.time.temporal.ChronoField
  * overrides for MaterialTheme.typography.display3 which is used to display the main picker
  * value.
  *
- * @param buttonIcon the button content.
- * @param onClick the button event handler.
+ * @param onValueConfirm the button event handler.
  * @param modifier the modifiers for the `Box` containing the UI elements.
- * @param initial the initial value to seed the picker with.
+ * @param value the initial value to seed the picker with.
```
ssancho14, line 81: Ok, rereading https://developer.android.com/jetpack/compose/state#state-hoisting , seems that you can use a name that makes sense for the type of data, so if you want to rename this parameter `time`, and the function `onTimeChanges`/`onTimeConfirmed` that will be fine too.
On the other hand, if all these pickers have different names for the parameters may get confusing.

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -60,18 +63,16 @@ import java.time.temporal.TemporalAdjusters
...
  */
 @ExperimentalHorologistComposablesApi
 @Composable
 public fun DatePicker(
-    buttonIcon: @Composable BoxScope.() -> Unit,
-    onClick: (LocalDate) -> Unit,
+    onValueConfirm: (LocalDate) -> Unit,
     modifier: Modifier = Modifier,
-    initial: LocalDate = LocalDate.now()
+    value: LocalDate = LocalDate.now()
```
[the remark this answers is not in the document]
yschimke: Yep, conscious decision. Plus apps should be able to use https://developer.android.com/studio/write/java8-support-table

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
yschimke: Makes sense, I'll follow up in a PR and do all together.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
yschimke: Let's go woth date and time then

luizgrp (approved):

Issue #171, opened 2022-05-16
Bump kotlin gradle plugin to 1.6.21


luizgrp:
https://kotlinlang.org/docs/releases.html#release-details

PR #180, opened 2022-05-18
Picker

JohnNichol:
#### WHAT
Update the kotlin gradle and kapt plugins to take version from kotlin variable. 

#### WHY
Keep the versions in sync - https://github.com/google/horologist/issues/171

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -52,7 +53,17 @@ import java.time.temporal.TemporalAdjusters
...
+ * Includes a slot for a button, typically for submitting.
+ *
+ * This component is designed to take most/all of the screen and utilizes large fonts. In order to
+ * ensure that it will draw properly on smaller screens it does not take account of user font size
+ * overrides for MaterialTheme.typography.display2 which is used to display the main picker
+ * value.
+ *
+ * @param buttonIcon the button content.
+ * @param onClick the button event handler.
+ * @param modifier the modifiers for the `Box` containing the UI elements.
```
luizgrp, line 67: nit: could [link](https://kotlinlang.org/docs/kotlin-doc.html#links-to-elements) with square brackets: `[Box]`

composables/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,25 @@
...
+  -->
+
+<resources>
+    <string name="time_picker_hour">Hour</string>
+    <string name="time_picker_min">Min</string>
+    <string name="time_picker_minute">Minute</string>
+    <string name="time_picker_second">Second</string>
+    <string name="picker_day">Day</string>
+    <string name="picker_month">Month</string>
+    <string name="picker_year">Year</string>
```
luizgrp: nit: we are now adding "horologist" prefix to the string resources as per https://github.com/google/horologist/pull/178

luizgrp (approved):
thanks for the tidy up 👍

PR #184, opened 2022-05-19
Add id and uri to MediaItem

luizgrp:
#### WHAT

Add `id` and `uri` properties to `MediaItem`.

#### WHY

First follow-up from review in https://github.com/google/horologist/pull/173#discussion_r873600662

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media/api/current.api
+++ b/media/api/current.api
@@ -18,17 +18,23 @@ package com.google.android.horologist.media.model {
   }
 
   public final class MediaItem {
-    ctor public MediaItem(optional String? title, String artist);
-    method public String? component1();
+    ctor public MediaItem(String id, String uri, optional String? title, String artist);
+    method public String component1();
     method public String component2();
-    method public com.google.android.horologist.media.model.MediaItem copy(String? title, String artist);
+    method public String? component3();
+    method public String component4();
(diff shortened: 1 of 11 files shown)


media/api/current.api
```diff
@@ -18,17 +18,23 @@ package com.google.android.horologist.media.model {
   }
 
   public final class MediaItem {
-    ctor public MediaItem(optional String? title, String artist);
-    method public String? component1();
-    method public String component2();
-    method public com.google.android.horologist.media.model.MediaItem copy(String? title, String artist);
+    ctor public MediaItem(String id, error.NonExistentClass uri, optional String? title, String artist);
```
yschimke, line 21: NonExistantClass

media/src/main/java/com/google/android/horologist/media/model/MediaItem.kt
```diff
@@ -16,10 +16,14 @@
...
 package com.google.android.horologist.media.model
 
+import java.net.URI
+
 /**
  * Representation of a media item.
  */
 public data class MediaItem(
+    val id: String,
+    val uri: URI,
```
yschimke, line 26: Do we need URI here? Is it simpler to leave as String here?

yschimke (approved):

PR #194, opened 2022-05-21
Make seconds optional on TimePicker

JohnNichol:
#### WHAT
Make the seconds picker optional on TimePicker

#### WHY
Often we only want users to select the hour and minute and removing the seconds picker will make the screen clearer and easier to use

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests
- DatePickerTest is failing - but I can't see how this patch is related.


yschimke (approved):
Nice!

PR #195, opened 2022-05-24
Disable picker tests until beta03

yschimke:
#### WHAT

Tests are flaky with wear compose beta2, even though pickers work in practice.

#### WHY

Race conditions that are unlikely to occur in real world use.

#### HOW

Disable with an annotation

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/composables/src/androidTest/java/com/google/android/horologist/TimePickerTest.kt
+++ b/composables/src/androidTest/java/com/google/android/horologist/TimePickerTest.kt
@@ -32,10 +32,12 @@ import androidx.wear.compose.material.Text
 import com.google.android.horologist.composables.ExperimentalHorologistComposablesApi
 import com.google.android.horologist.composables.TimePicker
 import com.google.android.horologist.composables.TimePickerWith12HourClock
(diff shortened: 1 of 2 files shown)


composables/src/androidTest/java/com/google/android/horologist/TimePickerTest.kt
```diff
@@ -32,10 +32,12 @@ import androidx.wear.compose.material.Text
 import com.google.android.horologist.composables.ExperimentalHorologistComposablesApi
 import com.google.android.horologist.composables.TimePicker
 import com.google.android.horologist.composables.TimePickerWith12HourClock
+import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
 import java.time.LocalTime
 
+@Ignore("Race condition in tests on beta02")
```
JohnNichol, line 40: Shouldn't be needed - but no harm I guess.

JohnNichol (approved):

PR #201, opened 2022-06-03
Add extras property to MediaItem

luizgrp:
#### WHAT

Add `extras` property to `MediaItem`.

#### WHY

In order to provide flexibility for users of the API to pass extra attributes between data and ui layer or use these attributes to implement specific use cases.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp:
cc: @dpreussler

yschimke:
Looks like nice pragmatic option.

PR #215, opened 2022-06-08
Match to Figma designs

yschimke:
Add a sample imported from figma to match against

image
image


media-ui/src/test/java/com/google/android/horologist/media/ui/FigmaPlayerScreenTest.kt
```diff
@@ -0,0 +1,118 @@
...
+import org.junit.Rule
+import org.junit.Test
+
+@Ignore("For interactive use only")
+class FigmaPlayerScreenTest {
+    @get:Rule
+    val paparazzi = Paparazzi(
+        deviceConfig = GALAXY_WATCH4_CLASSIC_LARGE,
+        theme = "android:ThemeOverlay.Material.Dark",
+        maxPercentDifference = 25.0,
```
luizgrp, line 50: question: which criteria was chosen for these values?

luizgrp (approved):
These 2 files should not be added to the source code:
- audio-ui/out/failures/com.google.android.horologist.audio.ui_VolumeScreenThemeTest_volumeScreenThemes[6]_uamp.png
- audio-ui/out/failures/delta-com.google.android.horologist.audio.ui_VolumeScreenThemeTest_volumeScreenThemes[6]_uamp.png

https://github.com/google/horologist/pull/306 should prevent this to happen in the future.

We should also consider the note in https://github.com/cashapp/paparazzi#git-lfs given the amount of images added and modified that we will have.

media-ui/src/test/java/com/google/android/horologist/media/ui/FigmaPlayerScreenTest.kt
yschimke: They don't really matter yet. The differences are too great, but I'm using these visually at the moment.

I'll reduce to 0 though.

PR #216, opened 2022-06-10
Fix preview backgrounds

yschimke:
#### WHAT

Make previews more realistic.

image

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
Was 
image

PR #217, opened 2022-06-10
Tile image loaders based on Coil.

yschimke:
#### WHAT

Loading a network image and creating an InlineImageResource is non trivial.

#### WHY

Simplify image fetching in real apps.

#### HOW

Suspending function that goes from ImageLoader+URL -> Bitmap -> InlineImageResource.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
cc @castedmo

gradle/libs.versions.toml
```diff
@@ -55,6 +55,7 @@ androidx-wear = { module = "androidx.wear:wear", version.ref = "androidxWear" }
 androidx-wear-tiles = { module = "androidx.wear.tiles:tiles", version.ref = "androidxtiles" }
 androidx-wear-tiles-testing = { module = "androidx.wear.tiles:tiles-testing", version.ref = "androidxtiles" }
 com-squareup-okhttp3-okhttp = { module = "com.squareup.okhttp3:okhttp", version.ref = "com-squareup-okhttp3" }
+coil = "io.coil-kt:coil-compose:2.1.0"
```
ataulm, line 58: nit: this file looks like it's a-z, so this should go before com-square

docs/tiles.md
```diff
@@ -16,6 +18,15 @@ class ExampleTileService : CoroutinesTileService() {
 }
 ```
 
+## Coil Image Helpers
+
+Provides a suspending method to load an image from the network, convert to a RGB_565
```
ataulm, line 23:
```suggestion
Provides a suspending method to load an image from the network, convert to an RGB_565
```

tiles/src/androidTest/java/com/google/android/horologist/tiles/ImagesTest.kt
```diff
@@ -0,0 +1,104 @@
...
+        bitmap = BitmapFactory.decodeResource(context.resources, R.drawable.coil)
+    }
+
+    @Test
+    public fun loadImageResource() {
+        runTest {
+            val imageResource = imageLoader.loadImageResource(context, bitmap)
+
+            val inlineResource = imageResource!!.inlineResource!!
+            assertThat(inlineResource).isNotNull()
```
ataulm, line 61: this is redundant because of the !! on the previous line

tiles/src/main/java/com/google/android/horologist/tiles/images.kt
```diff
@@ -0,0 +1,99 @@
...
+public suspend fun ImageLoader.loadImage(
+    context: Context,
+    data: Any?,
+    configurer: ImageRequest.Builder.() -> Unit = {}
+): Bitmap? {
+    val request = ImageRequest.Builder(context)
+        .data(data)
+        .apply(configurer)
+        .allowRgb565(true)
+        .allowHardware(false)
```
ataulm, line 44: what's the reason for disabling hardware?

tiles/src/main/java/com/google/android/horologist/tiles/images.kt
```diff
@@ -0,0 +1,99 @@
...
+        .build()
+    val response = execute(request)
+    return (response.drawable as? BitmapDrawable)?.bitmap
+}
+
+/**
+ * Load a Bitmap from a CoilImage loader.
```
ataulm, line 51:
```suggestion
 * Load an ImageResource from a CoilImage loader.
```

tiles/src/main/java/com/google/android/horologist/tiles/images.kt
```diff
@@ -0,0 +1,99 @@
...
+import java.nio.ByteBuffer
+
+/**
+ * Load a Bitmap from a CoilImage loader.
+ *
+ * @param context the context of the service or activity.
+ * @param data the image to fetch in one of the support Coil formats such as String, HttpUrl.
+ * @param configurer any additional configuration of the ImageRequest being built.
+ */
+public suspend fun ImageLoader.loadImage(
```
ataulm, line 35: Do we need the ones returning `Bitmap` to be public? Which cases do we need Bitmaps (vs. ImageResources)?

tiles/src/main/java/com/google/android/horologist/tiles/images.kt
```diff
@@ -0,0 +1,99 @@
...
+            .setFormat(ResourceBuilders.IMAGE_FORMAT_RGB_565)
+            .build()
+    )
+        .build()
+}
+
+/**
+ * Ensure the Bitmap is configured with RGB_565 required for Tiles.
+ */
+internal fun Bitmap.toRgb565(): Bitmap {
```
ataulm, line 93: Can it be private? It's only used in this file

```suggestion
private fun Bitmap.toRgb565(): Bitmap {
```

tiles/src/main/java/com/google/android/horologist/tiles/images.kt
```diff
@@ -0,0 +1,99 @@
...
+    configurer: ImageRequest.Builder.() -> Unit = {}
+): Bitmap? {
+    val request = ImageRequest.Builder(context)
+        .data(data)
+        .apply(configurer)
+        .allowRgb565(true)
+        .allowHardware(false)
+        .build()
+    val response = execute(request)
+    return (response.drawable as? BitmapDrawable)?.bitmap
```
fstanis, line 47: I think this needs better error handling. If the response is `SuccessResult`, the drawable is non-null, but if it's an `ErrorResult`, this method will just swallow the error and return null. I'd argue an exception is a better idea in the latter case, especially since this is a couroutine.

tiles/src/main/java/com/google/android/horologist/tiles/images.kt
yschimke: You need a Bitmap as a java object, I'll add a comment.

See https://coil-kt.github.io/coil/recipes/#palette

tiles/src/main/java/com/google/android/horologist/tiles/images.kt
yschimke: I think so, I originally had them separate. You may do the first part in the suspending method of the tile service, and the conversion at the last minute.  Also this form might be useful for other situations.

tiles/src/main/java/com/google/android/horologist/tiles/images.kt
yschimke: Removed following later suggestion.

tiles/src/main/java/com/google/android/horologist/tiles/images.kt
yschimke: Ooooffff, this is tough.  

I'm not certain about this.  ImageRequest has placeholder and error images.  If you set an error image you should get that rather than null.

It's also more inline with Coil compose usage.  I think null is the right thing here.  Even logging this isn't very useful, networks a patchy, you probably have to assume requests fail routinely and decide how to handle it.  Logging isn't something you would actually want generally.

I'll add a test for this.

yschimke: Yes, but the result is generally the Drawable.

```
sealed class ImageResult {
    abstract val drawable: Drawable?
    abstract val request: ImageRequest
}
```

It can be null if it fails and you don't set an error image.

yschimke: I don't want an exception on ErrorImage, I want the drawable.

PR #218, opened 2022-06-10
Previews for Tiles

yschimke:
#### WHAT

Provide a simple abstraction for the Render part of 

- [x] Docs

#### WHY

Using Android Studio previews is a big productivity win.

#### HOW

Use TileRenderer inside AndroidView.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
image

yschimke:
cc @ataulm

tiles/src/main/java/com/google/android/horologist/tiles/SingleTileLayoutRenderer.kt
```diff
@@ -0,0 +1,89 @@
...
+import androidx.wear.tiles.DeviceParametersBuilders
+import androidx.wear.tiles.LayoutElementBuilders.Layout
+import androidx.wear.tiles.LayoutElementBuilders.LayoutElement
+import androidx.wear.tiles.RequestBuilders
+import androidx.wear.tiles.ResourceBuilders.Resources
+import androidx.wear.tiles.TileBuilders.Tile
+import androidx.wear.tiles.TimelineBuilders
+import androidx.wear.tiles.material.Colors
+
+public abstract class SingleTileLayoutRenderer<T, R>(
```
ataulm, line 29: :+1: kdoc?

```
[TileLayoutRenderer] for a [Tile] that only has a single [TimelineBuilders.TimelineEntry].
```

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -0,0 +1,158 @@
...
+                /* loadActionListener = */ {}
+            )
+
+            tileRenderer.inflate(it)
+        }
+    )
+}
+
+@Composable
+public fun LayoutPreview(
```
ataulm, line 93: This one doesn't seem to be used by anything in this patch.

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -0,0 +1,158 @@
...
+    val resourceParams =
+        remember(tile.resourcesVersion) { resourceParams(resources, tile.resourcesVersion) }
+    val tileResources =
+        remember(state) { renderer.produceRequestedResources(resourceState, resourceParams) }
+
+    TilePreview(tile, tileResources)
+}
+
+@Composable
+public fun TilePreview(
```
ataulm, line 66: Should this one be private? It's not used outside of this file.

tiles/src/main/java/com/google/android/horologist/tiles/SingleTileLayoutRenderer.kt
yschimke: Yep - that was meant to be the checkbox in the description for docs.  A Todo list, about to commit them.

sample/build.gradle
```diff
@@ -87,6 +87,8 @@ dependencies {
     implementation projects.media
     implementation projects.mediaUi
     implementation projects.networkAwareness
+    implementation projects.tiles
+    implementation projects.composeTools
```
ataulm, line 91: I expected `TilesPreview` stuff to be included in `projects.tiles`

tiles/src/main/java/com/google/android/horologist/tiles/SingleTileLayoutRenderer.kt
```diff
@@ -0,0 +1,89 @@
...
+            )
+            .build()
+
+        return Tile.Builder()
+            .setResourcesVersion(PERMANENT_RESOURCES_VERSION)
+            .setTimeline(singleTileTimeline)
+            .build()
+    }
+
+    internal open fun createTheme(): Colors = Colors.DEFAULT
```
ataulm, line 58: Should this be `protected` rather than `internal`, otherwise implementors won't be able to override this I think.

yschimke: Thanks, good catch.

sample/build.gradle
yschimke: It's awkward because it's using TilesRenderer, and also Compose.  So if we do that you can't have a small tileservice in a views app. You get these extra dependencies also.

yschimke: However it fits quite nicely with the compose tools dependencies.

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
yschimke: No, I think it's useful if you don't want to use TileLayoutRenderer, but have already split out the Tile and Resource into another class. I have to imagine every Tile developer already does this, rather than embedding that logic directly into the service implementation.

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
yschimke: Added an example.

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -0,0 +1,158 @@
...
+import androidx.wear.tiles.StateBuilders.State
+import androidx.wear.tiles.TileBuilders
+import androidx.wear.tiles.TimelineBuilders
+import com.google.android.horologist.tiles.TileLayoutRenderer
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.asExecutor
+import kotlin.math.roundToInt
+
+@Composable
+public fun <T, R> TileLayoutPreview(state: T, resourceState: R, renderer: TileLayoutRenderer<T, R>) {
```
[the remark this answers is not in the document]
yschimke: Have added docs.

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -0,0 +1,178 @@
...
+                        LayoutElementBuilders.Layout.Builder().setRoot(
+                            Box.Builder()
+                                .setHorizontalAlignment(HORIZONTAL_ALIGN_CENTER)
+                                .setVerticalAlignment(VERTICAL_ALIGN_CENTER)
+                                .setHeight(ExpandedDimensionProp.Builder().build())
+                                .setWidth(ExpandedDimensionProp.Builder().build())
+                                .setModifiers(
+                                    Modifiers.Builder().setBackground(
+                                        Background.Builder()
+                                            .setColor(argb(Color.DKGRAY))
```
ataulm, line 130: Could this be passed as a param with a default value of DKGRAY? Since this is for individual components, we don't know that DKGRAY will be an appropriate background to view it properly (but it's a sensible default).

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -0,0 +1,178 @@
...
+                                            .build()
+                                    ).build()
+                                ).addContent(
+                                    Box.Builder()
+                                        .setHorizontalAlignment(HORIZONTAL_ALIGN_CENTER)
+                                        .setVerticalAlignment(VERTICAL_ALIGN_CENTER)
+                                        .setModifiers(
+                                            Modifiers.Builder().setBackground(
+                                                Background.Builder().setColor(
+                                                    argb(Color.BLACK)
```
ataulm, line 140: and/or this one. I can't visualize these in my mind just yet 😂

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
yschimke: Makes sense.

I'll add a comment explaining the choice also, it's that way so that you can see the bounds of the component. But agree some components might want something else.

image

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
yschimke: Can we try some variants of this, see what is the best default.  I'll commit with this and put a PR up showing examples.

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -0,0 +1,178 @@
...
+
+private fun requestParams(resources: Resources) =
+    RequestBuilders.TileRequest.Builder().setDeviceParameters(buildDeviceParameters(resources))
+        .setState(State.Builder().build()).build()
+
+private fun resourceParams(resources: Resources, version: String) =
+    RequestBuilders.ResourcesRequest.Builder().setDeviceParameters(buildDeviceParameters(resources))
+        .setVersion(version).build()
+
+private fun buildDeviceParameters(resources: Resources): DeviceParametersBuilders.DeviceParameters {
```
ataulm, line 168: It would be useful if this was public - e.g. the `Chip.Builder` takes context, clickable and device params, so getting the device parameters from the local context would be helpful.

ataulm: so far - haven't needed the others yet but slowly making my way through the component list.

ataulm: yep, maybe hold off until Weds EOD - that's when I'm planning to have finished going through all of this.

PR #219, opened 2022-06-10
Document version strategy

yschimke:
This versioning is confusing, but comes about because.

sample -> compose-tools -> tiles -> androidx.tiles (26)

We are stopping the version bump to 26 before we get to compose-tools so those libraries can be used in external apps.


yschimke:
cc @ataulm

Issue #200, opened 2022-06-03
Audio output chip on VolumeScreen is not displayed with full width


luizgrp:
The audio output chip on the VolumeScreen is displayed with full width on the screenshot at the page: https://google.github.io/horologist/audio-ui/

However, the chip does not seem to be taking the full width available of the screen:

|physical watch| android studio preview|
|---|---|
|![Screenshot_20220603_094319](https://user-images.githubusercontent.com/878134/171824709-80723990-efe1-470a-b430-3cae7ab2a698.png)|![Screen Shot 2022-06-03 at 10 06 46](https://user-images.githubusercontent.com/878134/171824678-3a2a24fb-870d-4a71-b42f-01a1b7bf6ce5.png)|

PS: I also checked the preview with d29eed95a81c977cd36d9015b06f991861a6e2f5 (before project was updated to wear compose beta02) and the result is the same.

PR #223, opened 2022-06-10
Fix layout of device name

yschimke:
#### WHAT

image

#### WHY

https://github.com/google/horologist/issues/200

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
```diff
@@ -168,44 +169,54 @@ public fun VolumeScreen(
...
+        modifier = Modifier
+            .padding(horizontal = 18.dp)
+            .width(intrinsicSize = IntrinsicSize.Max)
+            .semantics {
+                stateDescription = stateDescriptionText
+                onClick(onClickLabel) { onAudioOutputClick(); true }
+            },
+        label = {
+            Text(
+                modifier = Modifier.fillMaxWidth(),
```
yschimke, line 201: I think this is the fix, was

```
Text(
                        modifier = Modifier.width(IntrinsicSize.Min),
```

Issue #237, opened 2022-06-15
TimePicker's AM/PM label is shrunk after updating to wear-compose 1.0.0-rc01


shumelchyk:
TimePicker's AM/PM label is shrunk after updating to wear-compose 1.0.0-rc01:

![sdk_gwear_x86RWD4 211013 001kshum06152022222823](https://user-images.githubusercontent.com/1803903/173933110-7d00bff1-e1dc-4e9b-a03d-863b61951157.png)

shumelchyk:
Since we've added default vertical padding (8.dp) for enlarging tappable area of `CompatChip`, increasing height from 24 to 40dp (24+8+8) fixes the issue:
```
CompactChip(
                onClick = { amPm = 1 - amPm },
                modifier = Modifier.size(width = 50.dp, height = 40.dp),
                ...
```
@yschimke I can send a PR if needed (have this in local changes).

yschimke:
PR would be great, I'll land the version bumps, but you've already looked into it, so saves me fumbling about.

PR #239, opened 2022-06-16
Fix AM/PM label related to CompactChip chnages in 1.0.0-rc01

shumelchyk:
#### WHAT
TimePicker's AM/PM label looks broken on the recent Compose for Wear OS version

#### WHY
TimePicker's AM/PM label is shrunk after updating to wear-compose 1.0.0-rc01 (https://github.com/google/horologist/issues/237)

#### HOW
Since we've added default vertical padding (8.dp) for enlarging tappable area of CompatChip, increasing height from 24dp to 40dp (24+8+8) fixes the issue.
There is a small visual change in layout due to increased area, but it still looks good:

_Before:_
sdk_gwear_x86RWD4 211013 001kshum06152022225606

_After:_
sdk_gwear_x86RWD4 211013 001kshum06152022224923

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
@@ -30,6 +30,7 @@ import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.offset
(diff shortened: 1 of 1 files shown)


yschimke:
The previews in https://github.com/google/horologist/blob/main/composables/src/debug/java/com/google/android/horologist/composables/TimePicker12hPreview.kt aren't perfect but do show this sort of problem.  Can be useful to evaluate on different devices quickly.

PR #240, opened 2022-06-16
[Tiles] Split LayoutPreview api into element/root variants.

ataulm:
#### WHAT
Splits the `LayoutPreview` preview API for tiles into `LayoutElementPreview` and `LayoutRootPreview`.

#### WHY
The previous LayoutPreview API only worked for non-root elements, like Button. This is because the passed LayoutElement was also wrapped in a BoxLayout so that the passed element could be rendered in the center of the screen. Passing a root element (like PrimaryLayout) to this API which assumes that it has full width and height didn't work.

#### HOW
This PR adds two APIs, one intended for elements and one for root layouts. They're chained together, such that the element one will wrap the element in a BoxLayout (a root layout), and pass that to the root layout API.

image

#### Checklist :clipboard:
[-] Add explicit visibility modifier and explicit return types for public declarations
[-] Run spotless check
[] Run tests
[-] Update metalava's signature text files

### DIFF
--- a/compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
+++ b/compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
@@ -22,6 +22,7 @@ import android.content.res.Resources
 import android.graphics.Color
 import android.util.DisplayMetrics
 import android.widget.FrameLayout
+import androidx.annotation.ColorInt
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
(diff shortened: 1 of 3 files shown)


yschimke:
The instrumentation tests can be flaky, If they are mostly passing then just land it.

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -107,44 +108,49 @@ public fun TilePreview(
  */
 @ExperimentalHorologistComposeToolsApi
 @Composable
-public fun LayoutPreview(
-    layout: LayoutElement,
+public fun LayoutElementPreview(
+    element: LayoutElement,
+    @ColorInt windowBackgroundColor: Int = Color.DKGRAY,
```
yschimke, line 113: I wonder if we should actually prefer Compose colors here in the API? Thoughts? Can be a follow up if so.

yschimke (approved):

PR #241, opened 2022-06-16
Tile preview and some improvements

yschimke:
#### WHAT

Quick preview of material components for a Tile.

image

#### WHY

Want to validate colour choices when building tiles material screens.

#### HOW

Use a Tile with 5 material components and apply the theme.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
cc @ataulm

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -155,6 +157,10 @@ public fun LayoutPreview(
     TilePreview(tile, tileResources)
 }
 
+public val DummyClickable: Clickable = Builder()
```
ataulm, line 160: let's [change this one](https://developers.google.com/style/inclusive-documentation) to `NoOpClickable()`

sample/src/main/java/com/google/android/horologist/tile/SampleTheme.kt
```diff
@@ -0,0 +1,45 @@
...
+    Colors(0xFF981F68.toInt(), 0xFFFFFFFF.toInt(), 0xFF1C1B1F.toInt(), 0xFFFFFFFF.toInt())
+
+@WearLargeRoundDevicePreview
+@Composable
+public fun SampleThemePreview() {
+    ThemePreview(SampleTheme)
+}
+
+@Composable
+public fun ThemePreview(theme: Colors) {
```
ataulm, line 40: nit: Naming here is a little odd because this function isn't previewable. 

I guess we're not inlining it because we're showcasing how to pass a "theme", but how about renaming to `ThemeableTile` since this isn't a preview, the one above is?

tiles/src/main/java/com/google/android/horologist/tiles/preview/ThemePreviewTileRenderer.kt
```diff
@@ -0,0 +1,119 @@
...
+import com.google.android.horologist.tiles.render.SingleTileLayoutRenderer
+
+/**
+ * Tile that renders components with typical layouts and a theme colour.
+ */
+public class ThemePreviewTileRenderer(context: Context, private val thisTheme: Colors) :
+    SingleTileLayoutRenderer<Unit, Unit>(context) {
+    override fun createTheme(): Colors = thisTheme
+
+    internal val DummyClickable: Clickable = Builder()
```
ataulm, line 53: Do we need this or can we use the other one that's public in TilePreview? If we're keeping, let's rename.

tiles/src/main/java/com/google/android/horologist/tiles/render/RendererPreviewTileService.kt
```diff
@@ -0,0 +1,54 @@
...
+        super.onCreate()
+
+        renderer = createTileRenderer()
+    }
+
+    public abstract fun createTileRenderer(): S
+
+    public abstract fun createTileState(): T
+
+    public abstract fun createResources(): R
```
ataulm, line 42: Worth providing a default implementation (empty resources) here that can be overridden?

ataulm (changes requested):
Just a change to the clickable name then LGTM

sample/src/main/java/com/google/android/horologist/tile/SampleTheme.kt
yschimke: Went with ThemePreviewTile

tiles/src/main/java/com/google/android/horologist/tiles/preview/ThemePreviewTileRenderer.kt
yschimke: I'll move NoOpClickable to the tiles module, but it's a dependency issue.

tiles/src/main/java/com/google/android/horologist/tiles/render/RendererPreviewTileService.kt
yschimke: Maybe the name is wrong, you can't create an instance of R.  This is essentially mocking out the suspending work (producing images etc) before you build a Resources instance.

Changed to createResourcesInput() and added suspend.

ataulm (approved):

PR #246, opened 2022-06-17
Avoid setting background for screens.

yschimke:
#### WHAT

image

#### WHY

Screens should sit naturally on any background behind them, setting a constant background makes this impossible without overiding MaterialTheme just for this component.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
Note: the tests don't show that background, they now apply the background inside the test.

PR #256, opened 2022-06-22
Use new factory for VolumeViewModel.

yschimke:
#### WHAT

Use more modern API for ViewModel factory.

#### WHY

Nicer, safer API.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
nice!

PR #264, opened 2022-06-22
Standalone Media Sample

yschimke:
#### WHAT

Standalone Media Sample

#### WHY

It's way bigger than the rest of the samples. And we want to make it even bigger.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


media-sample/src/main/java/com/google/android/horologist/mediasample/components/MediaActivity.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    override fun onResume() {
+        super.onResume()
+
+        lifecycleScope.launchWhenResumed {
+            val playerRepository = viewModelModule.playerRepository
+            val uampService =
+                viewModelModule.mediaApplicationModule.networkModule.uampService
+
+            playerRepository.connected.filter { it }.first()
```
luizgrp, line 70: question: is this line necessary?

media-sample/src/main/java/com/google/android/horologist/mediasample/di/MediaActivityContainer.kt
```diff
@@ -0,0 +1,49 @@
...
+import android.app.Activity
+import com.google.android.horologist.mediasample.components.MediaActivity
+import com.google.android.horologist.mediasample.components.MediaApplication
+
+/**
+ * Simple DI implementation - to be replaced by hilt.
+ */
+class MediaActivityContainer(
+    internal val mediaApplicationModule: MediaApplicationModule,
+    private val activity: MediaActivity,
```
luizgrp, line 28: nit: not in use

media-sample/src/main/java/com/google/android/horologist/mediasample/di/MediaApplicationModule.kt
```diff
@@ -0,0 +1,110 @@
...
+import com.google.android.horologist.networks.data.DataRequestRepository
+import com.google.android.horologist.networks.status.NetworkRepository
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+
+/**
+ * Simple DI implementation - to be replaced by hilt.
+ */
+class MediaApplicationModule(internal val application: MediaApplication) {
```
luizgrp, line 42: nit: sorry if my message was confused in the other PR but I think this class was fine being called MediaApplicationContainer given it has the application as scope? same for PlaybackServiceContainer

media-sample/src/main/java/com/google/android/horologist/mediasample/components/MediaActivity.kt
yschimke: Without this if you call setMediaItems too early it's a no-op.

```
    override fun setMediaItems(mediaItems: List<MediaItem>) {
        checkNotClosed()

        player.value?.let {
```

PR #268, opened 2022-06-22
Hide time text from Volume Screen.

yschimke:
#### WHAT

image

#### WHY

Short visits to dialogs, shouldn't have a time.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
Was 
image

luizgrp (approved):

PR #270, opened 2022-06-23
Fix kdoc and simplify code

luizgrp:
#### WHAT

- Fix kdoc;
- Simplify code;

#### WHY

- Links were broken in KDoc;
- Place logic to change modifier closer to related logic to choose icon;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Thanks

PR #281, opened 2022-06-28
Gapless playback test [in progress]

yschimke:
#### WHAT

Test gapless playback, currently a manual test, since it doesn't confirm playback via events yet.

#### WHY

Confirm gapless playback on emulator and on device.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


media-sample/src/androidTest/java/com/google/android/horologist/mediasample/AudioOffloadTest.kt
```diff
@@ -0,0 +1,183 @@
...
+                browser.waitForPlaying()
+            }
+
+            val format: Format = audioOffloadManager.format.value!!
+
+            val offloadExpected = offloadExpected(output, Build.MODEL, format)
+
+            val flags = checkAdbDumpsys()
+
+            if (offloadExpected) {
```
luizgrp, line 85: nit: I don't think there should be conditionals in our test code, I'd suggest to:
- break this test into 2: one for offload and another without offload
- add annotation `@RequiresOffload(boolean)` so if the current device does not meet the requirements, the test is skipped

media-sample/src/androidTest/java/com/google/android/horologist/mediasample/AudioOffloadTest.kt
```diff
@@ -0,0 +1,183 @@
...
+            .setSampleRate(sampleRate)
+            .setChannelMask(channelConfig)
+            .setEncoding(encoding)
+            .build()
+    }
+
+    private suspend fun checkAdbDumpsys(): List<String> {
+        val adbOutput = dumpSysAudioFlinger().lines()
+
+        val offloadRegex = "offload".toRegex(RegexOption.IGNORE_CASE)
```
luizgrp, line 150: nit: does it need to be converted to a regex?

luizgrp (approved):

media-sample/src/androidTest/java/com/google/android/horologist/mediasample/AudioOffloadTest.kt
yschimke: I can use `it.contains("offload", ignoreCase = true)`

I think I'm still inclined to assume a regex is more optimal in that case. But don't know for sure.

PR #284, opened 2022-06-28
Add Preferences DataStore to allow configuring sample ui for testing.

yschimke:
#### WHAT

image

image

#### WHY

Need to test other UI designs.

#### HOW

Use Preferences DataStore to control UI layout. 

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-sample/src/main/java/com/google/android/horologist/mediasample/ui/app/MediaPlayerAppViewModel.kt
+++ b/media-sample/src/main/java/com/google/android/horologist/mediasample/ui/app/MediaPlayerAppViewModel.kt
@@ -16,8 +16,6 @@
 
 package com.google.android.horologist.mediasample.ui.app
 
-import androidx.datastore.core.DataStore
-import androidx.datastore.preferences.core.Preferences
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import androidx.lifecycle.viewmodel.initializer
@@ -27,8 +25,8 @@ import com.google.android.horologist.media3.offload.AudioOffloadManager
 import com.google.android.horologist.mediasample.AppConfig
 import com.google.android.horologist.mediasample.catalog.UampService
 import com.google.android.horologist.mediasample.di.MediaApplicationContainer
+import com.google.android.horologist.mediasample.domain.SettingsRepository
 import com.google.android.horologist.mediasample.ui.debug.OffloadState
-import com.google.android.horologist.mediasample.ui.settings.Settings
 import com.google.android.horologist.networks.data.DataRequestRepository
 import com.google.android.horologist.networks.data.DataUsageReport
 import com.google.android.horologist.networks.data.Networks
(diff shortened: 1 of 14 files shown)


media-sample/src/main/java/com/google/android/horologist/mediasample/ui/app/MediaPlayerAppViewModel.kt
```diff
@@ -109,10 +114,13 @@ class MediaPlayerAppViewModel(
     }
 
     suspend fun startupSetup(navigateToLibrary: () -> Unit) {
+        val loadAtStartup =
+            prefsDataStore.data.map { Settings(it) }.first().loadItemsAtStartup
```
luizgrp, line 118: nit: `Settings` from package `ui.settings` is being used here in `ui.app` which kind of indicates that `Settings` belongs to a wider package. There is also repetition of `prefsDataStore.data.map { Settings(it) }` in `SettingsScreenViewModel`.

 I'd suggest to:
- move `Settings` to a domain package;
- create a `SettingsRepository` in the domain package - that would abstract how the settings are stored, if we decide to store in another place there will be only one place to change, etc;
- change ViewModels to use `SettingsRepository` instead of `DataStore` directly;
- add a `SettingsUiModel` to `ui.settings` with more explicit properties, e.g. `podcastControlsEnabled: Boolean` in order to avoid interpretation of domain/data models in the view code, like in `enabled = uiState.settings != null`;

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/settings/UampSettingsScreen.kt
```diff
@@ -87,39 +91,49 @@ private fun ActionSetting(
...
                     )
-                }, onCheckedChange = onCheckedChange, label = {
-                Text(text)
-            }, modifier = Modifier.fillMaxWidth()
+                },
+                enabled = enabled,
+                onCheckedChange = onCheckedChange,
+                label = {
+                    Text(text)
+                }, modifier = Modifier.fillMaxWidth()
```
luizgrp, line 136: nit: weird indentation here, the parenthesis of the line below should be aligned with the ToggleChip - same above in line 112

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/settings/Settings.kt
```diff
@@ -0,0 +1,35 @@
...
+package com.google.android.horologist.mediasample.ui.settings
+
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.booleanPreferencesKey
+
+data class Settings(
+    val podcastControls: Boolean = false,
+    val loadItemsAtStartup: Boolean = true
+) {
+    constructor(preferences: Preferences) : this(
```
luizgrp, line 26: nit: I would rather have a separate class to have this mapping from one model to another in order to avoid having the model class coupled to other dependencies. Few issues that I can see with this is:
- as per my comment above, if this is moved to a domain package, the domain layer should not know of external framework dependencies
- if there is a need to create this Setting model from another source, like a database, then another constructor would be added here, and then this class will start to know about how to deal with databases too, evolving to a class that knows too many things

luizgrp (approved):

PR #285, opened 2022-06-28
Remove DefaultPlayerScreenMediaDisplay

luizgrp:
#### WHAT

Remove `DefaultPlayerScreenMediaDisplay`.

#### WHY

It became redundant after adding `DefaultMediaDisplay` in https://github.com/google/horologist/pull/280

#### HOW

- Remove code for `DefaultPlayerScreenMediaDisplay` and replace usages by `DefaultMediaDisplay`;
- Rename function `DefaultMediaDisplay` in `PlayerScreenPreview` file to `DefaultPlayerScreen` as the names were conflicting;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/screens/PlayerScreen.kt
```diff
@@ -81,18 +81,6 @@ public fun PlayerScreen(
     )
 }
 
-/**
- * Default [MediaDisplay] implementation for [PlayerScreen].
- */
-@ExperimentalHorologistMediaUiApi
-@Composable
-public fun DefaultPlayerScreenMediaDisplay(playerUiState: PlayerUiState) {
```
yschimke, line 89: I do wonder whether it's still useful to have this split.  But perhaps, DefaultPlayerScreenMediaDisplay implemented in terms of MediaDisplay.  The reason being is that our designs have content like "Nothing playing".

Does that make sense in a MediaDisplay composable used for a entity screen? Or just for the player screen?

luizgrp: > I do wonder whether it's still useful to have this split

I'm easy about this one, so happy to keep it and close this PR.

---

> But perhaps, DefaultPlayerScreenMediaDisplay implemented in terms of MediaDisplay

We had an implementation in that way, previously. Not with :
- [mediaDisplay](https://github.com/google/horologist/pull/117/files#diff-76a930c4cdf90911c49582d7c22ecd04a09b87e7892963f65693a603851a9653R52) param was of type `PlayScreenMediaDisplay`
- [DefaultPlayScreenMediaDisplay](https://github.com/google/horologist/pull/117/files#diff-76a930c4cdf90911c49582d7c22ecd04a09b87e7892963f65693a603851a9653R133) was of type  `PlayScreenMediaDisplay`

That way it could be extended to also implement the "Nothing playing" scenario.

luizgrp:
Closing as per https://github.com/google/horologist/pull/285#discussion_r908378660

PR #286, opened 2022-06-28
Update UI Display when no item or not connected

yschimke:
#### WHAT

image

image

#### WHY

Avoid blank space, as per UX designs.

#### HOW

DefaultPlayerScreenMediaDisplay includes logic to show other status information, while DefaultMediaDisplay shows the track details only.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/screens/PlayerScreen.kt
```diff
@@ -75,22 +79,37 @@ public fun PlayerScreen(
...
-    TextMediaDisplay(
-        title = playerUiState.mediaItem?.title,
-        artist = playerUiState.mediaItem?.artist
-    )
+public fun DefaultPlayerScreenMediaDisplay(
+    playerUiState: PlayerUiState,
+    modifier: Modifier = Modifier
+) {
+    val mediaItem = playerUiState.mediaItem
+    if (!playerUiState.connected) {
```
luizgrp, line 100: nit: given the UI layer has its own model `PlayerUiState`, it could be changed to attend the specific needs of the UI classes, e.g. `playerUiState.loading` instead of `playerUiState.connected`; or `playerUiState.mediaDisplayStatus` be either of values `Loading`, `Playing` or `NotPlaying`.

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/PlayerScreen.kt
```diff
@@ -75,22 +79,37 @@ public fun PlayerScreen(
...
+        LoadingMediaDisplay(modifier)
+    } else if (mediaItem != null) {
+        DefaultMediaDisplay(
+            mediaItem = mediaItem,
+            modifier = modifier
+        )
+    } else {
+        InfoMediaDisplay(
+            message = stringResource(R.string.horologist_nothing_playing),
```
luizgrp, line 109: nit: I see the benefit of the library providing default content descriptions to the icons it displays (which can be overridden to provide translations) but I feel that displaying a text on the screen in a specific language by default is more invasive and would rather require this as parameter to the component - I might be biased and maybe I should not see any difference in between these two scenarios

audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/actions/SettingsButton.kt
```diff
@@ -0,0 +1,65 @@
...
+import androidx.wear.compose.material.Icon
+import androidx.wear.compose.material.MaterialTheme
+import com.google.android.horologist.audio.VolumeState
+import com.google.android.horologist.audio.ui.ExperimentalHorologistAudioUiApi
+import com.google.android.horologist.audio.ui.semantics.CustomSemanticsProperties.iconImageVector
+
+/**
+ * Button to launch a screen to control the system volume.
+ *
+ * See [VolumeState]
```
luizgrp, line 37: nit: update kdoc

media-ui/src/main/java/com/google/android/horologist/media/ui/state/PlayerViewModel.kt
```diff
@@ -38,9 +38,18 @@ public open class PlayerViewModel(
...
-    ).stateIn(
+        playerRepository.shuffleModeEnabled
+    ) { currentState, availableCommands, mediaItem, mediaItemPosition, shuffleModeEnabled ->
+        PlayerUiStateMapper.map(
+            currentState,
+            availableCommands,
+            mediaItem,
+            mediaItemPosition,
+            shuffleModeEnabled,
+            // TODO work out how to combine more than 5 flows
```
luizgrp, line 49: 😱 

```kotlin
combine(
  combine(f1,f2,f3,f4,f5, map),
  combine(f6,f7,f8,f9,f10, map),
  map
)
```

😅

luizgrp (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/PlayerScreen.kt
yschimke: I'll follow up on this and below as I'm building on this currently.

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/PlayerScreen.kt
yschimke: I'll follow up on this .

PR #289, opened 2022-06-28
Add MediaArtwork

luizgrp:
#### WHAT

Add `MediaArtwork`.

Screen Shot 2022-06-28 at 18 57 17

#### WHY

In order to provide common components.

#### HOW

- Move `MediaArtwork` from `media-sample` to `media-ui`;
- Rename `title` parameter to `contentDescription`;
- Add `placeholder` parameter;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/debug/java/com/google/android/horologist/media/ui/components/MediaArtworkPreview.kt
```diff
@@ -0,0 +1,43 @@
...
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Album
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.graphics.vector.rememberVectorPainter
+import androidx.compose.ui.tooling.preview.Preview
+import com.google.android.horologist.media.ui.ExperimentalHorologistMediaUiApi
+import com.google.android.horologist.media.ui.state.model.MediaItemUiModel
+
+@Preview(
+    backgroundColor = 0x00000000,
```
yschimke, line 30: why transparent here? I guess that's why I prefer named previews, although they have their limitations. But easier to see why you choose a color.

yschimke (approved):

media-ui/src/debug/java/com/google/android/horologist/media/ui/components/MediaArtworkPreview.kt
luizgrp: Fair point and well spotted, I changed it there so the icon can be seen, the icon has black color. Happy to hear any suggestions to keep it the same color as the other previews.

PR #291, opened 2022-06-29
Add MediaChip

luizgrp:
#### WHAT

Add `MediaChip`.

Screen Shot 2022-06-29 at 12 21 35

#### WHY

In order to provide common components.

#### HOW

- Move `MediaChip` from `media-sample` to `media-ui`;
- Add `defaultTitle` and `placeholder` params;
- Add previews;
- Add tests;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
The previews don't really "pop". Rather than the generic placeholder, perhaps

I *think* Coil supports direct android resources as URL Strings.

https://github.com/coil-kt/coil/blob/main/coil-base/src/androidTest/java/coil/fetch/AssetUriFetcherTest.kt

yschimke (approved):

PR #292, opened 2022-06-29
[DO NOT MERGE] Check if paparazzi is running in CI

luizgrp:
I would expect the tests to fail in CI with these changes.


yschimke:
Booo!!!

luizgrp:
This PR should be failing after merging https://github.com/google/horologist/pull/293 but it's not: https://github.com/google/horologist/actions/runs/2585883933.

PR #304, opened 2022-07-01
Avoid running most complex tests in CI automatically.

yschimke:
#### WHAT

-Pandroid.testInstrumentationRunnerArguments.size=small,medium

#### WHY

Android instrumentation tests are flaky

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


scripts/run-tests.sh
```diff
@@ -106,4 +106,4 @@ if [ "$SHARD_COUNT" -gt "0" ]; then
   SHARD_OPTS="$SHARD_OPTS -Pandroid.testInstrumentationRunnerArguments.shardIndex=$SHARD_INDEX"
 fi
 
-./gradlew --scan --continue --no-configuration-cache --stacktrace $TASK $FILTER_OPTS $SHARD_OPTS
+./gradlew --scan --continue --no-configuration-cache --stacktrace $TASK $FILTER_OPTS $SHARD_OPTS -Pandroid.testInstrumentationRunnerArguments.size=small,medium
```
luizgrp, line 109: nit: this certainly does the trick, however, it will be passing this param for every call made using this script, even for unit tests, where these params are irrelevant.

I'd suggest we build something following the structure of this file, adding params like `--run-all-sizes` and `--run-up-to-medium` which would work similarly to [--run-flaky-tests](https://github.com/google/horologist/blob/1d90950cad735f809f295615000240152edc46a8/scripts/run-tests.sh#L51-L54) and add this parameter in the [FILTER_OPTS](https://github.com/google/horologist/blob/1d90950cad735f809f295615000240152edc46a8/scripts/run-tests.sh#L71-L75).

Alternatively we could make unit tests call a separate script (run-unit-tests.sh) as it would reduce the amount of logical conditions of the scripts. I don't see much benefit in the current script of using it for both instrumented and unit tests.

Marking as nit - we can work this out in another iteration.

luizgrp (approved):

PR #307, opened 2022-07-01
Fix seek seconds.

yschimke:
#### WHAT

image

#### WHY

Show seconds

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/components/controls/SeekButtonIncrement.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/components/controls/SeekButtonIncrement.kt
@@ -33,11 +33,12 @@ public sealed class SeekButtonIncrement(
     public data class Other(override val seconds: Int) : SeekButtonIncrement(seconds)
 
     public companion object {
-        public fun SeekButtonIncrement(seconds: Int): SeekButtonIncrement {
+        public fun ofSeconds(seconds: Int?): SeekButtonIncrement {
             return when (seconds) {
                 Five.seconds -> Five
                 Ten.seconds -> Ten
                 Thirty.seconds -> Thirty
+                null -> Unknown
                 else -> Other(seconds)
             }
         }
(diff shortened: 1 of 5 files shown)


media-ui/src/main/java/com/google/android/horologist/media/ui/components/controls/SeekButtonIncrement.kt
```diff
@@ -32,4 +31,15 @@ public sealed class SeekButtonIncrement(
     public object Thirty : SeekButtonIncrement(30)
 
     public data class Other(override val seconds: Int) : SeekButtonIncrement(seconds)
+
+    public companion object {
+        public fun SeekButtonIncrement(seconds: Int): SeekButtonIncrement {
```
luizgrp, line 36: nit: could make seconds param nullable and return Unknown in that scenario?

luizgrp (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/components/controls/SeekButtonIncrement.kt
yschimke: https://github.com/google/horologist/issues/308

luizgrp: hmm I meant this factory function and not the interface itself

btw this function name should start with lower case, it gets confusing after naming so many composable functions that way 😅

luizgrp: weird that code analysis is not picking it up

yschimke: Yep, I understood about the factory taking a nullable, but the problem is the input isn't nullable so doesn't help yet. I've done it now, but will still need to follow up and remove the default of 0.

yschimke: BTW this does follow the conventions

_Exception: factory functions used to create instances of classes can have the same name as the abstract return type:_

```
interface Foo { /*...*/ }

class FooImpl : Foo { /*...*/ }

fun Foo(): Foo { return FooImpl() }
```

luizgrp: thanks!! TIL

PR #318, opened 2022-07-04
Rename preview files to match PlayerScreen class name

luizgrp:
#### WHAT

Rename preview files to match PlayerScreen class name.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Thanks

Issue #323, opened 2022-07-05
Fix VolumeScreenIndividualTest and VolumeScreenThemeTest


luizgrp:
After merging https://github.com/google/horologist/pull/322, `build` step in main branch got [green](https://github.com/google/horologist/runs/7190095284?check_suite_focus=true).

However, the PRs that are up to date with main branch started to [fail](https://github.com/google/horologist/runs/7192510060?check_suite_focus=true) with:

```
com.google.android.horologist.audio.ui.VolumeScreenIndividualTest > volumeScreenAtMaximum FAILED
    java.lang.NullPointerException at VolumeScreenIndividualTest.kt:66
com.google.android.horologist.audio.ui.VolumeScreenIndividualTest > volumeScreenAtMinimum FAILED
    java.lang.NullPointerException at VolumeScreenIndividualTest.kt:49
com.google.android.horologist.audio.ui.VolumeScreenIndividualTest > volumeScreenWithLongTest FAILED
    java.lang.NullPointerException at VolumeScreenIndividualTest.kt:83
```

---

Once again, main is green, but the PRs that are up to date with main branch started to [fail](https://github.com/google/horologist/runs/7195908369?check_suite_focus=true) with:

```
com.google.android.horologist.audio.ui.VolumeScreenThemeTest > volumeScreenThemes[0] FAILED
    java.lang.NullPointerException at VolumeScreenThemeTest.kt:56
com.google.android.horologist.audio.ui.VolumeScreenThemeTest > volumeScreenThemes[1] FAILED
    java.lang.NullPointerException at VolumeScreenThemeTest.kt:56
com.google.android.horologist.audio.ui.VolumeScreenThemeTest > volumeScreenThemes[2] FAILED
    java.lang.NullPointerException at VolumeScreenThemeTest.kt:56
com.google.android.horologist.audio.ui.VolumeScreenThemeTest > volumeScreenThemes[3] FAILED
    java.lang.NullPointerException at VolumeScreenThemeTest.kt:56
com.google.android.horologist.audio.ui.VolumeScreenThemeTest > volumeScreenThemes[4] FAILED
    java.lang.NullPointerException at VolumeScreenThemeTest.kt:56
com.google.android.horologist.audio.ui.VolumeScreenThemeTest > volumeScreenThemes[5] FAILED
    java.lang.NullPointerException at VolumeScreenThemeTest.kt:56
10 tests completed, 7 failed, 3 skipped
[...]

PR #324, opened 2022-07-05
Ignore failing tests

luizgrp:
#### WHAT

https://github.com/google/horologist/issues/323

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
Thanks

PR #327, opened 2022-07-05
Add a Media Collection Tile.

yschimke:
#### WHAT

image

#### WHY

Support standard templates for Media tiles.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/tiles/MediaCollectionsTileRenderer.kt
```diff
@@ -0,0 +1,144 @@
...
+    private fun collectionChip(
+        collection: MediaCollection,
+        deviceParameters: DeviceParameters
+    ) = Chip.Builder(
+        context,
+        Clickable.Builder().setOnClick(collection.action).build(),
+        deviceParameters
+    )
+        .setChipColors(ChipColors.secondaryChipColors(theme))
+        .setPrimaryTextIconContent(collection.name, "collection-${collection.id}")
```
castedmo, line 98: `"$CollectionPrefix${collection.id}"`?

media-sample/src/main/java/com/google/android/horologist/mediasample/components/MediaActivity.kt
```diff
@@ -55,4 +65,8 @@ class MediaActivity : ComponentActivity() {
             viewModelModule.addCreationExtras(this)
         }
     }
+
+    companion object {
+        const val CollectionKey = "collection"
```
luizgrp, line 70: nit: good practice to prefix values of these intent extras keys with the app's package name

media-sample/src/main/java/com/google/android/horologist/mediasample/tile/MediaCollectionsTileService.kt
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Show UAMP as AppIcon, and favourites and podcasts icons.
+     */
+    override suspend fun resourcesRequest(requestParams: ResourcesRequest): Resources {
+        return renderer.produceRequestedResources(
+            MediaCollectionsTileRenderer.ResourceState(
+                R.drawable.ic_uamp,
+                mapOf(
+                    1 to drawableResToImageResource(R.drawable.ic_baseline_queue_music_24),
+                    2 to drawableResToImageResource(R.drawable.ic_baseline_podcasts_24)
```
castedmo, line 108: Would be more clear to use collectionIds here (or some constant which is used on both places).

castedmo (approved):
LGTM for the tile parts

media-sample/src/main/java/com/google/android/horologist/mediasample/components/MediaActivity.kt
```diff
@@ -47,6 +49,14 @@ class MediaActivity : ComponentActivity() {
                 navController = navController,
                 creationExtras = { defaultViewModelCreationExtras }
             )
+
+            LaunchedEffect(key1 = Unit) {
+                val collectionId = intent.getStringExtra(CollectionKey)
+                if (collectionId != null) {
+                    intent.removeExtra(CollectionKey)
+                    navController.navigateToCollection(collectionId.toString())
```
luizgrp, line 57: nit: is the toString needed? if so it can be avoided with that [awkward](https://github.com/google/horologist/pull/317#discussion_r913629814) syntax:

```kotlin
collectionId?.let {
    intent.removeExtra(CollectionKey)
    navController.navigateToCollection(it)
}
```

media-sample/src/main/res/values/strings.xml
```diff
@@ -28,5 +28,8 @@
     <string name="horologist_artwork_gradient">Use Artwork Gradient</string>
     <string name="horologist_show_test_dialog">Show Test Dialog</string>
     <string name="horologist_sample_error">An error occurred</string>
+    <string name="horologist_favorites">Favorites</string>
+    <string name="media_collection">Media Collection</string>
```
luizgrp, line 32: nit: missing the prefix - commenting for the sake of consistency, I only see the benefit in having it for the published libraries and not in the sample app

media-ui/build.gradle
```diff
@@ -92,6 +93,7 @@ project.tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile.class).co
 
 dependencies {
     api projects.media
+    api projects.tiles
```
luizgrp, line 96: any particular reason for adding this as `api` instead of `implementation`?

media-ui/src/main/java/com/google/android/horologist/media/ui/tiles/MaterialTheme.kt
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.media.ui.tiles
+
+import androidx.compose.ui.graphics.toArgb
+import androidx.wear.compose.material.Colors
+
+public fun Colors.toTileColors(): androidx.wear.tiles.material.Colors {
```
luizgrp, line 22: missing experimental annotation?

media-ui/src/main/java/com/google/android/horologist/media/ui/tiles/MediaCollectionsTileRenderer.kt
```diff
@@ -0,0 +1,144 @@
...
+import androidx.wear.tiles.material.Colors
+import androidx.wear.tiles.material.CompactChip
+import androidx.wear.tiles.material.layouts.PrimaryLayout
+import com.google.android.horologist.tiles.ExperimentalHorologistTilesApi
+import com.google.android.horologist.tiles.render.SingleTileLayoutRenderer
+
+/**
+ * A tile template showing two media collections and a primary chip that links to playlists.
+ */
+public class MediaCollectionsTileRenderer(
```
luizgrp, line 46: missing experimental annotation for all classes defined in this file?

tiles/build.gradle
```diff
@@ -101,9 +101,9 @@ dependencies {
     implementation libs.androidx.lifecycle.runtime
     implementation libs.androidx.concurrent.future
 
-    implementation libs.androidx.wear.tiles
-    implementation libs.androidx.wear.tiles.material
-    implementation libs.androidx.lifecycle.service
+    api libs.androidx.wear.tiles
+    api libs.androidx.wear.tiles.material
+    api libs.androidx.lifecycle.service
```
luizgrp, line 106: were they changed so that `media-ui` gets access to them? if so I'd rather see them defined as implementation in `media-ui` so we know explicitly which libs it depends on?

luizgrp (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/tiles/MediaCollectionsTileRenderer.kt
yschimke: Changed this to the media id.

media-sample/src/main/java/com/google/android/horologist/mediasample/tile/MediaCollectionsTileService.kt
yschimke: Fixed. The previews have dummy data still.

media-sample/src/main/java/com/google/android/horologist/mediasample/components/MediaActivity.kt
yschimke: Fixed.

media-ui/build.gradle
yschimke: Because the types are also used by extending projects like media-sample.

media-sample/src/main/res/values/strings.xml
yschimke: Fixed. Maybe we should remove them.

tiles/build.gradle
yschimke: I think they can be both, api here, and implemenation in media-ui. If we expose types with these as public API, then we should have API here.

media-sample/src/main/java/com/google/android/horologist/mediasample/components/MediaActivity.kt
yschimke: I'll follow up on this. Like?

        const val CollectionKey = "com.google.android.horologist.mediasample.tile.collection"

tiles/build.gradle
luizgrp: that's a fair point about our APIs exposing the types of these other libs, but a pain to manually verify them, specially as code changes with time

I wonder if there is a tool out there to flag this for us

media-ui/build.gradle
luizgrp: I'm in agreement with the point in https://github.com/google/horologist/pull/327#discussion_r915713455 rather than the point above, if it's also applicable here?

PR #328, opened 2022-07-05
Add SecondaryPlaceholderChip

luizgrp:
#### WHAT

Add `SecondaryPlaceholderChip`.

Screen Shot 2022-07-05 at 18 24 54

#### WHY

In order to have our base building blocks aligned with specs in Figma.

#### HOW

- Add `material-ripple` as dependency to `media-ui`;
- Add implementation based on [Chip](https://github.com/androidx/androidx/blob/5775cb72bdc32a90328bf39e212fab31c5960316/wear/compose/compose-material/src/commonMain/kotlin/androidx/wear/compose/material/Chip.kt#L294);
- Add preview in the debug folder;
- Add snapshot test;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/components/base/SecondaryPlaceholderChip.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/components/base/SecondaryPlaceholderChip.kt
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.media.ui.components.base
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.interaction.MutableInteractionSource
+import androidx.compose.foundation.layout.Arrangement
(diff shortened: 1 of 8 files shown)


media-ui/src/main/java/com/google/android/horologist/media/ui/components/base/SecondaryPlaceholderChip.kt
```diff
@@ -0,0 +1,127 @@
...
+ * Secondary placeholder chip.
+ */
+@ExperimentalHorologistMediaUiApi
+@Composable
+internal fun SecondaryPlaceholderChip(
+    modifier: Modifier = Modifier,
+    onClick: () -> Unit = {},
+    enabled: Boolean = true,
+) {
+    val backgroundColor = Color(
```
yschimke, line 60: Isn't their a copy version to adapt just alpha

media-ui/src/main/java/com/google/android/horologist/media/ui/components/base/SecondaryPlaceholderChip.kt
```diff
@@ -0,0 +1,127 @@
...
+import androidx.wear.compose.material.ChipDefaults
+import androidx.wear.compose.material.MaterialTheme
+import com.google.android.horologist.media.ui.ExperimentalHorologistMediaUiApi
+
+/**
+ * Secondary placeholder chip.
+ */
+@ExperimentalHorologistMediaUiApi
+@Composable
+internal fun SecondaryPlaceholderChip(
```
yschimke, line 55: If we are reimplementing core material components, we should push back. Find ways to use the existing components with customisation. Or change the designs to follow material components

luizgrp: I started trying to use [Chip](https://github.com/androidx/androidx/blob/5775cb72bdc32a90328bf39e212fab31c5960316/wear/compose/compose-material/src/commonMain/kotlin/androidx/wear/compose/material/Chip.kt#L294) and just pass the labels and icon contents as param but it wasn't straight forward, so then I built from scratch. 
The abstraction we need for our screens is defined here so I will merge and raise an issue to change this implementation to use Chip.

PR #330, opened 2022-07-06
Allow ImageVector to be passed as icon to SecondaryChip

luizgrp:
#### WHAT

Allow `ImageVector` to be passed as icon to `SecondaryChip`.

Screen Shot 2022-07-06 at 12 02 31

<details>
<summary>All previews</summary>

All previews after the changes (in order to check that there were no side effects - which is also checked by snapshot tests).

</details>

#### WHY

In order to allow usage like the below, which was causing Coil to throw exception otherwise:

```kotlin
    SecondaryChip(
        primaryLabel = "Primary label",
        onClick = { },
        icon = Icons.Default.Add
    )
```

#### HOW

- Verify if `icon` param is of type `ImageVector` and create an `Icon`;
- Add preview;
- Add snapshot test;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/components/base/SecondaryChip.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/components/base/SecondaryChip.kt
@@ -27,12 +27,14 @@ import androidx.compose.foundation.layout.width
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.painter.Painter
+import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.text.style.TextAlign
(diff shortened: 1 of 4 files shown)


media-ui/src/main/java/com/google/android/horologist/media/ui/components/base/SecondaryChip.kt
```diff
@@ -93,15 +95,23 @@ internal fun SecondaryChip(
...
-                    Image(
-                        painter = rememberAsyncImagePainter(
-                            model = icon,
-                            placeholder = placeholder
-                        ),
-                        contentDescription = null, // hidden from talkback
-                        modifier = Modifier.size(iconSize),
-                        contentScale = ContentScale.Fit
-                    )
+                    if (icon is ImageVector) {
```
yschimke, line 98: raised https://github.com/coil-kt/coil/discussions/1348 for discussion

PR #331, opened 2022-07-06
Add BrowseScreen

luizgrp:
#### WHAT

Add `BrowseScreen`.

Screen Shot 2022-07-06 at 14 52 37

#### WHY

#### HOW

- Add `DownloadPlaylistUiModel`;
- Add `BrowseScreen` implementation;
- Add previews in `debug` folder;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/test/java/com/google/android/horologist/media/ui/screens/BrowseScreenTest.kt
```diff
@@ -0,0 +1,115 @@
...
+import com.google.android.horologist.media.ui.state.model.DownloadPlaylistUiModel
+import com.google.android.horologist.media.ui.utils.rememberVectorPainter
+import com.google.android.horologist.paparazzi.GALAXY_WATCH4_CLASSIC_LARGE
+import com.google.android.horologist.paparazzi.WearSnapshotHandler
+import com.google.android.horologist.paparazzi.determineHandler
+import org.junit.Ignore
+import org.junit.Rule
+import org.junit.Test
+
+@Ignore("Can't generate golden images - recording task is generating empty ones")
```
luizgrp, line 39: @yschimke is this something you came across before and could point me to the right direction? 

would this be the reason why [MediaPlayerTestCase](https://github.com/google/horologist/blob/c0fed820b2d7154a22104e54f52d19c5aefecda8/media-ui/src/test/java/com/google/android/horologist/media/ui/MediaPlayerTestCase.kt) was added  to place the screen inside Scaffold + Theme + RoundPreview?

media-ui/src/debug/java/com/google/android/horologist/media/ui/screens/browse/BrowseScreenPreview.kt
```diff
@@ -0,0 +1,202 @@
...
+        focusRequester = FocusRequester(),
+        scalingLazyListState = rememberScalingLazyListState(),
+        downloadItemArtworkPlaceholder = rememberVectorPainter(
+            image = Icons.Default.FeaturedPlayList,
+            tintColor = Color.Green,
+        )
+    )
+}
+
+@Preview(
```
yschimke, line 71: Can we avoid all these definitions and use the method name to differentiate?

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/browse/BrowseScreen.kt
```diff
@@ -0,0 +1,172 @@
...
+ * A screen to:
+ * - display user's [downloaded media items][DownloadPlaylistUiModel];
+ * - provide access to libraries;
+ * - provide access to settings;
+ */
+@ExperimentalHorologistMediaUiApi
+@Composable
+public fun BrowseScreen(
+    browseScreenState: BrowseScreenState,
+    onDownloadItemClick: (DownloadPlaylistUiModel) -> Unit,
```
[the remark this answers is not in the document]
yschimke: How much of this comes from the spec for all media apps, vs what we currently have in front of us.

media-ui/src/test/java/com/google/android/horologist/media/ui/screens/BrowseScreenTest.kt
luizgrp: I will delete these test and rethink on the strategy for testing this screen, given it doesn't fit in a full screen I don't think we should be using snapshot tests.

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/browse/BrowseScreen.kt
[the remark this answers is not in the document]
luizgrp: It's a fair point, I agree it should be optional.

The flexibility is already provided in MediaPlayerScaffold, so maybe instead of slot here we could just provide another screen which does not contain the downloads section.

I will leave to iterate over in another PR.

PR #344, opened 2022-07-07
Fix maxPercentDifference discrepancies in tests

luizgrp:
#### WHAT

Fix maxPercentDifference discrepancies in tests

#### WHY

As per https://github.com/google/horologist/pull/338#discussion_r915893640, they were supposed to be the same value.

#### HOW

- Extract a variable in order to try to prevent similar issues in the future;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/audio-ui/src/test/kotlin/com/google/android/horologist/audio/ui/VolumeScreenIndividualTest.kt
+++ b/audio-ui/src/test/kotlin/com/google/android/horologist/audio/ui/VolumeScreenIndividualTest.kt
@@ -32,12 +32,14 @@ import org.junit.Test
 
 @Ignore("https://github.com/google/horologist/issues/323")
 class VolumeScreenIndividualTest {
+    private val maxPercentDifference = 0.1
+
     @get:Rule
     val paparazzi = Paparazzi(
         deviceConfig = GALAXY_WATCH4_CLASSIC_LARGE,
         theme = "android:ThemeOverlay.Material.Dark",
-        maxPercentDifference = 0.0,
-        snapshotHandler = WearSnapshotHandler(determineHandler(0.1))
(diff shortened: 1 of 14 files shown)


audio-ui/src/test/kotlin/com/google/android/horologist/audio/ui/VolumeScreenIndividualTest.kt
```diff
@@ -32,12 +32,14 @@ import org.junit.Test
 
 @Ignore("https://github.com/google/horologist/issues/323")
 class VolumeScreenIndividualTest {
+    private val maxPercentDifference = 0.0
```
luizgrp, line 35: @yschimke please let me know which value is supposed to be used, 0.0 or 0.1

media-ui/src/test/java/com/google/android/horologist/media/ui/FigmaPlayerScreenTest.kt
```diff
@@ -51,15 +51,17 @@ import org.junit.runners.Parameterized
 class FigmaPlayerScreenTest(
     private val deviceConfig: DeviceConfig
 ) {
+    private val maxPercentDifference = 0.1
```
luizgrp, line 54: is this meant to be 0.1?

audio-ui/src/test/kotlin/com/google/android/horologist/audio/ui/VolumeScreenIndividualTest.kt
yschimke: 0.1

yschimke (approved):
Thanks, should be 0.1. I do wonder whether we should have some standard definition that defaults to 0.1?

Issue #339, opened 2022-07-07
False positives by Paparazzi snapshot tests


luizgrp:
Paparazzi snapshot tests does not fail when it should. I don't think it happens 100% of the time, but it's frequently enough.

Previously noticed in https://github.com/google/horologist/pull/292 

Recently noticed after submitting this [commit](https://github.com/google/horologist/commit/aac4dcbc48eba74c07609f2af6b589b377cde04b) which removes **all** the usages of  WearSnapshotHandler. So **all** screenshots went from circle to square. And the snapshots tests [passed](https://github.com/google/horologist/runs/7233660999?check_suite_focus=true). We can see in the logs in CI that most of the tasks are marked as "UP-TO-DATE" when running `verifyPaparazziDebug`:

```
> Task :media-ui:testDebugUnitTest UP-TO-DATE
> Task :media-ui:verifyPaparazziDebug UP-TO-DATE
```

Potentially related to https://github.com/cashapp/paparazzi/issues/388.

luizgrp:
After https://github.com/google/horologist/pull/345 was merged, main got [red](https://github.com/google/horologist/runs/7248667906?check_suite_focus=true), which is a good indication that the fix put in place might have worked. It needs monitoring and this issue should be reopened if the problem persists.

Issue #340, opened 2022-07-07
 NPE in SnapshotVerifier during tests


luizgrp:
Most of the tests are failing in CI with:
```
java.lang.NullPointerException: goldenImage must not be null
	at app.cash.paparazzi.SnapshotVerifier$newFrameHandler$1.handle(SnapshotVerifier.kt:53)
	at com.google.android.horologist.paparazzi.WearSnapshotHandler$newFrameHandler$1.handle(WearSnapshotHandler.kt:57)
	at app.cash.paparazzi.Paparazzi$takeSnapshots$1$2.invoke(Paparazzi.kt:302)
	at app.cash.paparazzi.Paparazzi$takeSnapshots$1$2.invoke(Paparazzi.kt:295)
	at app.cash.paparazzi.Paparazzi.withTime(Paparazzi.kt:336)
	at app.cash.paparazzi.Paparazzi.takeSnapshots(Paparazzi.kt:295)
	at app.cash.paparazzi.Paparazzi.snapshot(Paparazzi.kt:211)
	at app.cash.paparazzi.Paparazzi.snapshot(Paparazzi.kt:204)
	at app.cash.paparazzi.Paparazzi.snapshot$default(Paparazzi.kt:195)
	at com.google.android.horologist.media.ui.FigmaVolumeScreenTest.volumePlayerScreen(FigmaVolumeScreenTest.kt:46)
```

The stacktrace above can be seen in the test reports in the archived artifact. In the actions logs it displays the following:
```
com.google.android.horologist.media.ui.FigmaVolumeScreenTest > volumePlayerScreen FAILED
    java.lang.NullPointerException at FigmaVolumeScreenTest.kt:46

...

53 tests completed, 35 failed
> Task :media-ui:testDebugUnitTest FAILED
```

Some investigation reported [here](https://github.com/google/horologist/issues/323#issuecomment-1177298063).

Usage of `WearSnapshotHandler` was removed in https://github.com/google/horologist/pull/338 for now.

luizgrp:
Potentially fixed by https://github.com/google/horologist/pull/350; it needs monitoring and this issue can be reopened.

PR #345, opened 2022-07-07
Call clean task with verifyPaparazziDebug 

luizgrp:
#### WHAT

Call clean task with verifyPaparazziDebug 

#### WHY

In order to prevent it using cache, in an attempt to fix https://github.com/google/horologist/issues/339

#### HOW

- add clean task to be called with verifyPaparazziDebug;
- move "snapshot test" step to be executed after "upload test reports" step, due to the clean task that would delete the previously generated reports;
- change "upload snapshot tests failures" step to also archive the test results and to run "always";

#### Checklist :clipboard:
- [N/A] Run spotless check
- [x] Run tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,11 +63,6 @@ jobs:
               --unit-tests \
               --affected-base-ref=$BASE_REF
 
-      - name: Snapshot tests
-        run: |
-          ./gradlew --scan --stacktrace \
-              verifyPaparazziDebug
-
       - name: Upload test results and reports
         if: always()
         uses: actions/upload-artifact@v2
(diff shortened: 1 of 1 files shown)


luizgrp:
With this PR, this issue https://github.com/google/horologist/issues/340 is [surfaced](https://github.com/google/horologist/runs/7237417377?check_suite_focus=true) again:

```
> Task :media-ui:testDebugUnitTest
com.google.android.horologist.media.ui.FigmaVolumeScreenTest > volumePlayerScreen FAILED
    java.lang.NullPointerException at FigmaVolumeScreenTest.kt:46
```

luizgrp:
After https://github.com/google/horologist/issues/340 was merged, main got [red](https://github.com/google/horologist/runs/7248667906?check_suite_focus=true), which is a good indication that the fix put in place might have worked. It needs monitoring and this issue should be reopened if the problem persists.

PR #348, opened 2022-07-08
Centralise navigation implementation.

yschimke:
#### WHAT

Centralise navigation details.

#### WHY

Avoid different parts of the code disagreeing on naming.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/navigation/NavigationScreens.kt
```diff
@@ -16,20 +16,112 @@
...
+        public val collectionId: String = "collectionId"
+
+        public fun destination(
+            id: String,
+            collectionId: String? = null
+        ): String = "mediaItem?id=$id" + (if (collectionId != null) "&collectionId=$collectionId" else null)
+
+        override fun deepLinks(deepLinkPrefix: String): List<NavDeepLink> = listOf(
+            navDeepLink {
+                uriPattern = "$deepLinkPrefix/mediaItem?id={id}&collectionId={collectionId}"
```
luizgrp, line 80: nit: could have all the pieces defined as val and reused across? "$deepLinkPrefix/$mediaItemPath?$idParam={id}&$collectionIdParam={collectionId}"

luizgrp (approved):
nice one!

media-ui/src/main/java/com/google/android/horologist/media/ui/navigation/NavigationScreens.kt
yschimke: Maybe, but they are coincidentally the same here.  Let's try this for a bit, I suspect we will refactor this, maybe make it not a singleton and baked in, but allow others to change it?

PR #350, opened 2022-07-08
Fix Git LFS

yschimke:
#### WHAT

Files were not checked out.

#### WHY

Needs extra param.  See https://github.com/actions/checkout

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):
:shipit:

PR #352, opened 2022-07-08
Add PlaylistScreen

luizgrp:
#### WHAT

Add `PlaylistScreen`.

Screen Shot 2022-07-08 at 13 25 33

#### WHY

#### HOW

- Add `PlaylistUiModel`;
- Change `DownloadPlaylistUiModel` to contain a `PlaylistUiModel`;
- Add `PlaylistScreen` implementation;
- Add previews in debug folder;
- Change `BrowseScreen` to pass `largeIcon = true` to `SecondaryChip`;
- Change `SecondaryChip` param name to `label` in order to align with param name from material Chip;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/playlist/PlaylistScreen.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/playlist/PlaylistScreen.kt
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.media.ui.screens.playlist
+
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.graphics.painter.Painter
+import androidx.compose.ui.unit.dp
+import androidx.wear.compose.material.ScalingLazyColumn
+import androidx.wear.compose.material.ScalingLazyListState
+import com.google.android.horologist.compose.navscaffold.scrollableColumn
+import com.google.android.horologist.media.ui.ExperimentalHorologistMediaUiApi
+import com.google.android.horologist.media.ui.R
+import com.google.android.horologist.media.ui.components.base.SecondaryChip
+import com.google.android.horologist.media.ui.components.base.SecondaryPlaceholderChip
+import com.google.android.horologist.media.ui.components.base.Title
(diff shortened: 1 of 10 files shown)


media-ui/src/main/java/com/google/android/horologist/media/ui/screens/playlist/PlaylistScreen.kt
```diff
@@ -0,0 +1,92 @@
...
+                    SecondaryChip(
+                        label = it.title,
+                        onClick = { onPlaylistItemClick(it) },
+                        icon = it.artworkUri,
+                        largeIcon = true,
+                        placeholder = playlistItemArtworkPlaceholder
+                    )
+                }
+            }
+        } else if (playlistScreenState is PlaylistScreenState.Loading) {
```
yschimke, line 71: Nice!

media-ui/src/main/res/values/strings.xml
```diff
@@ -31,4 +31,5 @@
     <string name="horologist_browse_library_title">Library</string>
     <string name="horologist_browse_library_playlists">Playlists</string>
     <string name="horologist_browse_library_settings">Settings</string>
+    <string name="horologist_browse_playlist_title">Playlist</string>
```
yschimke, line 34: Shouldn't this be a playlist name?

yschimke (approved):

media-ui/src/main/res/values/strings.xml
luizgrp: This is the screen to display a list of playlists, not the entity screen. Are we talking about the same screen? Otherwise I don't follow the question, sorry!

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/playlist/PlaylistScreen.kt
```diff
@@ -0,0 +1,92 @@
...
+        item {
+            Title(
+                textId = R.string.horologist_browse_playlist_title,
+                modifier = Modifier.padding(bottom = 12.dp),
+            )
+        }
+
+        if (playlistScreenState is PlaylistScreenState.Loaded) {
+            playlistScreenState.playlistList.forEach {
+                item {
```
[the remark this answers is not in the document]
luizgrp: nice, thanks!

yschimke (commented):
Shouldn't this be playlists (plural) in the code?

This also gets into naming, might be worth checking with Danny. Maybe its Collection or some other term.

luizgrp:
> Shouldn't this be playlists (plural) in the code?

asked in Figma

> This also gets into naming, might be worth checking with Danny. Maybe its Collection or some other term.

this is a broader strategy to be defined for the whole lib and not just in this PR - needs to be defined and iterated over all the default texts

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/playlist/PlaylistScreen.kt
```diff
@@ -0,0 +1,91 @@
...
+                modifier = Modifier.padding(bottom = 12.dp),
+            )
+        }
+
+        if (playlistScreenState is PlaylistScreenState.Loaded) {
+            items(count = playlistScreenState.playlistList.size) { index ->
+
+                val playlist = playlistScreenState.playlistList[index]
+
+                SecondaryChip(
```
yschimke, line 64: I think we should pull this up into a param, with our chip as the default. But let them use other chip designs.

yschimke (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/playlist/PlaylistScreen.kt
luizgrp: could do as a slot param, could also make the model generic, add the title as param so it's fully customisable - will raise an issue to discuss it and improve it once this is landed

PR #353, opened 2022-07-08
Change media sample to use PlaylistScreen

luizgrp:
#### WHAT

Change media sample to use `PlaylistScreen`.

#### WHY

As per specs in Figma.

#### HOW

- Change `UampPlaylistsScreen` to use `PlaylistsScreen`;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-sample/src/main/java/com/google/android/horologist/mediasample/ui/library/PlaylistUiModelMapper.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import com.google.android.horologist.media.ui.state.model.MediaItemUiModel
+import com.google.android.horologist.media.ui.state.model.PlaylistUiModel
+
+object PlaylistUiModelMapper {
+
+    fun map(
+        mediaItemUiModel: MediaItemUiModel,
+        defaultTitle: String = "",
+        shouldMapArtworkUri: Boolean = true,
```
yschimke, line 27: Nice, I like this.

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/library/UampPlaylistsScreenViewModel.kt
```diff
@@ -71,12 +71,16 @@ class UampPlaylistsScreenViewModel(
...
-        UiState(items?.map { MediaItemUiModelMapper.map(it) })
-    }.stateIn(viewModelScope, started = SharingStarted.Eagerly, initialValue = UiState())
+        UiState.Loaded(items?.let { items.map { MediaItemUiModelMapper.map(it) } } ?: emptyList())
+    }.stateIn(viewModelScope, started = SharingStarted.Eagerly, initialValue = UiState.Loading)
 
-    data class UiState(
-        val items: List<MediaItemUiModel>? = null,
-    )
+    sealed class UiState {
+        object Loading : UiState()
```
yschimke, line 78: Do we need some failure state?

yschimke (approved):

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/library/UampPlaylistsScreenViewModel.kt
luizgrp: I think so, requested in Figma 👍

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/library/UampPlaylistsScreen.kt
```diff
@@ -16,64 +16,53 @@
...
             )
         }
-        if (items != null) {
-            items(items) {
-                val mediaItem =
-                    if (settingsState?.showArtworkOnChip == true) it else it.copy(artworkUri = null)
-                MediaChip(
-                    mediaItem = mediaItem,
-                    onClick = {
-                        uampPlaylistsScreenViewModel.play(it)
```
yschimke, line 67: Without this, playing is not possible. I don't want to add it back while other in progress changes.  Any suggestions?

luizgrp: apologies for missing that, I tested and I thought the app was still playing? maybe I missed the fact that is playing the wrong song 🤔 will check that tomorrow

for now I guess this call can be added to this lamba in [UampWearApp](https://github.com/google/horologist/blob/c5fb2c1461dd04ecf1c4aba27d00a50cd8be6441/media-sample/src/main/java/com/google/android/horologist/mediasample/ui/app/UampWearApp.kt#L124-L126), changing the type of [onPlaylistItemClick](https://github.com/google/horologist/blob/c5fb2c1461dd04ecf1c4aba27d00a50cd8be6441/media-sample/src/main/java/com/google/android/horologist/mediasample/ui/library/UampPlaylistsScreen.kt#L34) in `UampPlaylistsScreen` accordingly. Will put some more thoughts on this solution tomorrow if you think it's not the right approach.

luizgrp: should be fixed in https://github.com/google/horologist/pull/376

PR #355, opened 2022-07-08
Rename SecondaryChip to StandardChip

luizgrp:
#### WHAT

Rename `SecondaryChip` to `StandardChip`.

Screen Shot 2022-07-08 at 16 59 40

Screen Shot 2022-07-08 at 17 10 16

#### WHY

In order to allow it to also be used for primary chip.

#### HOW

- Rename `SecondaryChip` to `StandardChip`;
- Add `StandardChipType` enum;
- Add `chipType: StandardChipType` param to `StandardChip`;
- Add preview and snapshot tests for primary chip type;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/api/current.api
+++ b/media-ui/api/current.api
@@ -98,10 +98,10 @@ package com.google.android.horologist.media.ui.components.background {
 
 package com.google.android.horologist.media.ui.components.base {
 
-  public final class SecondaryChipKt {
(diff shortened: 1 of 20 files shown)


yschimke (approved):
Assuming a mostly mechanical change

Issue #66, opened 2022-04-06
Failing SnackbarHost test


yschimke:
```
com.google.android.horologist.compose.snackbar.SnackbarHostTest > snackbarHost_returnedResult[test(AVD) - 11] FAILED 
	androidx.compose.ui.test.ComposeTimeoutException: Condition still not satisfied after 5000 ms
	at androidx.compose.ui.test.AndroidComposeTest.waitUntil(ComposeTest.android.kt:218)
```

Have disabled for now in https://github.com/google/horologist/pull/65

MohitMandalia:
SnackbarHost `snackbarHost_returnedResult` issue seems to be resolved I ran the test by removing the `@Ignore` annotation it seems to be passing.

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/36087806/178152568-4e4a5c93-e75d-4d03-a51e-3c0fa431a83a.png">

yschimke:
Great, want to put up a PR?

yschimke:
Please complete https://github.com/google/horologist/blob/main/CONTRIBUTING.md#contributor-license-agreement also

MohitMandalia:
sure
This issue was already fixed so i'll just remove the `@Ignore` annotation and make a PR for it.

yschimke:
Tracking in other bug

PR #358, opened 2022-07-10
Remove Ignore annotation from SnackbarHostTest (#66)

MohitMandalia:
#### WHAT
Remove Ignore annotation from SnackbarHostTest

#### WHY
As the test `snackbarHost_returnedResult` is not failing `@Ignore` annotation should be removed

#### HOW

- Remove `@Ignore` annotation from `snackbarHost_returnedResult()` and `SnackbarHostTest`

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


MohitMandalia:
I have signed the License.

Issue #356, opened 2022-07-10
Overly strict check on position


yschimke:
Seems we can't guarantee this.

```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.google.android.horologist.mediasample.debug, PID: 31837
    java.lang.IllegalStateException: Current position has to be greater than duration [current: 5.003s] [duration: 4.999s].
```

PR #361, opened 2022-07-11
Handle positions beyond duration.

yschimke:
#### WHAT

#fixes https://github.com/google/horologist/issues/356

#### WHY

Media Metadata may lie.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/media-data/src/test/java/com/google/android/horologist/media/data/PlayerRepositoryStateTest.kt
+++ b/media-data/src/test/java/com/google/android/horologist/media/data/PlayerRepositoryStateTest.kt
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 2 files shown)


media-data/src/test/java/com/google/android/horologist/media/data/PlayerRepositoryStateTest.kt
```diff
@@ -0,0 +1,84 @@
...
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import kotlin.time.Duration.Companion.milliseconds
+
+@RunWith(RobolectricTestRunner::class)
+@Config(manifest = Config.NONE)
+class PlayerRepositoryStateTest {
```
luizgrp, line 33: nit: it looks like this test is not actually testing the class `PlayerRepositoryImpl` but the `Player?.readPosition` extension function.

In order to be a `PlayerRepositoryImplTest`:
- the class under test should be a test double of `PlayerRepositoryImpl` and not of `Player` (here it uses `StubPlayer`);
-  the `updatePosition` function and `mediaItemPosition` property should be used instead of the `readPosition` extension function;

The other "benefit" of doing this way is that the `readPosition` extension function can be marked as private again, and it won't be marked as "internal" just for the sake of testing.

luizgrp (approved):

media-data/src/test/java/com/google/android/horologist/media/data/PlayerRepositoryStateTest.kt
yschimke: Won't I have to implement a lot of stubplayer?

I first looked at changing existing tests, bit looked onerous

yschimke: You were right, was pretty simple.

PR #364, opened 2022-07-11
[DO NOT MERGE] Add paparazzi diff for comparison

luizgrp:


yschimke:
Was a 4 pixel difference.

PR #367, opened 2022-07-12
Correct the playing status logic.

yschimke:
#### WHAT

Correct logic from https://github.com/google/horologist/pull/360

See https://github.com/google/horologist/pull/360

#### WHY

Loading state is a broken proxy for Buffering.  So use the correct value.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/media-data/src/main/java/com/google/android/horologist/media/data/PlayerRepositoryImpl.kt
+++ b/media-data/src/main/java/com/google/android/horologist/media/data/PlayerRepositoryImpl.kt
@@ -97,6 +97,7 @@ public class PlayerRepositoryImpl : PlayerRepository, Closeable {
 
             if (events.contains(Player.EVENT_MEDIA_ITEM_TRANSITION)) {
                 _currentMediaItem.value = player.currentMediaItem?.let(MediaItemMapper::map)
+                updatePosition()
             }
 
             // Reason for handling these events here, instead of using individual callbacks
@@ -109,7 +110,7 @@ public class PlayerRepositoryImpl : PlayerRepository, Closeable {
                     Player.EVENT_IS_LOADING_CHANGED,
                     Player.EVENT_IS_PLAYING_CHANGED,
                     Player.EVENT_PLAYBACK_STATE_CHANGED,
-                    Player.EVENT_PLAY_WHEN_READY_CHANGED
+                    Player.EVENT_PLAY_WHEN_READY_CHANGED,
                 )
             ) {
                 updateState(player)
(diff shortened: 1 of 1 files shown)


luizgrp (approved):

media-data/src/main/java/com/google/android/horologist/media/data/PlayerRepositoryImpl.kt
```diff
@@ -130,13 +131,14 @@ public class PlayerRepositoryImpl : PlayerRepository, Closeable {
...
     private fun updateState(player: Player) {
-        _currentState.value = if ((player.isPlaying || player.isLoading) && player.playWhenReady) {
-            PlayerState.Playing
-        } else if (player.isLoading) {
-            PlayerState.Loading
-        } else {
-            PlayerStateMapper.map(player.playbackState)
-        }
+        _currentState.value =
+            if ((player.playbackState == Player.STATE_BUFFERING || player.playbackState == Player.STATE_READY) && player.playWhenReady) {
```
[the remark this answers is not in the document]
yschimke: playing is strictly whether time is moving forward.  It's not relevant for the UI components, because it should be a subset of this logic.

media-data/src/test/java/com/google/android/horologist/media/data/PlayerStateMapperTest.kt
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package com.google.android.horologist.media.data
+
+import androidx.media3.common.Player
+import com.google.android.horologist.media.model.PlayerState
+import com.google.common.truth.Truth.assertThat
+import org.junit.Test
+
+class PlayerStateMapperTest {
```
luizgrp, line 24: nit: the tests in this file don't look like unit tests for `PlayerStateMapper` itself, they look like integration tests for `media3.Player` and `PlayerStateMapper` in the `FakePlayer` implementation. Which is odd given it looks like it is testing a test double implementation. 

I'd question if this is necessary. If yes, then I would consider moving the test double to an individual module ("media-data-fixtures"), where the implementation is in main folder and this test is in the test folder. 

Alternatively, as a quick win would be to just rename this file and move to another package (suggestion, prefix with `integration` so it is clear: `integration.player.PlayerTest`).

luizgrp (approved):

media-data/src/test/java/com/google/android/horologist/media/data/PlayerStateMapperTest.kt
luizgrp: to clarify my previous comment, I think it's valid to have the test for the test double, if it is complex enough and we would like to provide it as fixture, where I suggested to check if it's "necessary" I meant to check if it's intended

yschimke: To me it's focused on PlayerStateMapper. I've introduced PlayerStateMapper and MediaItemPositionMapper, because they are evidently things that are discussion worthy, even a doc for one of them.  Pulling it out because they don't appear to be just minor implementation details of PlayerRepositoryImpl, but standalone business logic.

I don't understand the comment about it testing the test double, it's hard to use the TestExoPlayer to test specific states like this (or if it's simple, then we can follow up on this PR). The test double just extends StubPlayer to return specific values, and I really want a test that documents the various cases for PlayerStateMapper. It's not a general purpose one, just enough to get specific player states.

luizgrp: totally agree with the benefits of extracting out those classes.

re "testing the test double", my apologies, I missed the call to `PlayerStateMapper.map` in line 52 (or this was added now after my comment) so I had the impression that this test class was only calling functions from FakePlayer class

nit: suggestion for clarity:
```kotlin
@Test
    fun `check playback state while playing`() {
        val player = fakeStatePlayer.overrideState(
            playbackState = Player.STATE_READY,
            playWhenReady = true
        )

        val state = PlayerStateMapper.map(fakeStatePlayer)

        assertThat(state).isEqualTo(PlayerState.Playing)
    }
```

it adds two extra lines to each test, but it makes clearer which function is being tested, specially now that the class has more than one public function

PR #368, opened 2022-07-12
Smaller images

yschimke:
#### WHAT

Avoid large images that cause visual jank when scrolling.

#### WHY

Music can be interrupted by visual jank.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp:
https://github.com/google/horologist/pull/369 has been merged and you can proceed with this PR, thanks for your patience!

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/library/UampRepository.kt
```diff
@@ -0,0 +1,72 @@
...
+import coil.request.ImageRequest
+import com.google.android.horologist.mediasample.catalog.UampService
+import com.google.android.horologist.mediasample.catalog.model.Music
+
+/**
+ * Uamp repository for loading media items.
+ *
+ * TODO move this into a DB, with initial load from API.
+ */
+class UampRepository(
```
luizgrp, line 30: "uamp" repository implies that it's a repository for the "uamp" app, which is very broad. Classes like this tend to become huge as everything is added to it.

I'd rather go with one repository per entity, I see that you tried to limit the scope specifying "media items" in the kdoc, so maybe MusicRepository would be more suitable.

Now Music is a model specific to the retrofit API, I'd have a repository for our own domain models.

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/library/UampRepository.kt
```diff
@@ -0,0 +1,72 @@
...
+            "https://storage.googleapis.com/uamp/Kai_Engel_-_Irsens_Tale/art.jpg" -> "$GithubPrefix/kai.jpg"
+            "https://storage.googleapis.com/automotive-media/album_art.jpg" -> "$GithubPrefix/album_art.jpg"
+            "https://storage.googleapis.com/automotive-media/album_art_2.jpg" -> "$GithubPrefix/album_art_2.jpg"
+            "https://storage.googleapis.com/automotive-media/album_art_3.jpg" -> "$GithubPrefix/album_art_3.jpg"
+            else -> image
+        }
+    }
+
+    companion object {
+        val GithubPrefix = "https://media.githubusercontent.com/media/google/horologist/main/media-sample/src/main/res/drawable-mdpi"
```
luizgrp, line 70: why hitting the network to load these images if they are already in the apk? why not load the file from the disk?

yschimke: To be a realistic reference app. We could do the same for media, but it doesn't help us prove out the libraries.

luizgrp: I see, does that also explain why [this](https://github.com/google/horologist/pull/360#discussion_r917726378) wasn't a concern?

if these images were added to be used for this purpose then no point in shipping them with the apk, maybe place them outside the android project? or are there any other plans for them in the future?

luizgrp (approved):

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/library/UampRepository.kt
yschimke: Yeah, this was being expedient.  I'll add a task to set up a storage server.

yschimke: https://github.com/google/horologist/issues/377

luizgrp: cool, just to clarify when I say outside the android project, I meant you could still have them in this repo, but not in the `res` folder in `media-sample` (which makes them be shipped with the apk), but another folder in the root of the project for example

yschimke: OK, I'll move them.

media-sample/src/main/java/com/google/android/horologist/mediasample/di/MediaApplicationContainer.kt
```diff
@@ -195,21 +195,18 @@ class MediaApplicationContainer(
         DataUpdates(updater)
     }
 
-    fun getPlaylistRepository(playlistRepositoryImpl: PlaylistRepositoryImpl): PlaylistRepository =
-        playlistRepositoryImpl
-
-    fun getPlaylistRepositoryImpl(playlistRemoteDataSource: PlaylistRemoteDataSource): PlaylistRepositoryImpl =
-        PlaylistRepositoryImpl(playlistRemoteDataSource = playlistRemoteDataSource)
+    val playlistRepository: PlaylistRepository by lazy {
```
yschimke, line 198: @luizgrp I'm happy to replace this janky manual DI with a framework, but until we bite that bullet, I brought them more inline with the existing definitions.

luizgrp: no worries but I thought you wanted something more aligned with existing DI libs, the way I had built is more easily reused when introducing dagger/hilt

yschimke: OK, I understand now. Yep, when we make the change. But a partial change was confusing and not adding much right now.

Should we define other factory function in a reusable place? These types here probably wouldn't raise to that level as they are tie to this app.

luizgrp (approved):

media-sample/src/main/java/com/google/android/horologist/mediasample/di/MediaApplicationContainer.kt
yschimke: It's the partial bit that worries me, if we clean this up, we should do it with a clear target of what we want. if we just change a few as we touch things, I think it will be confusing.

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/library/UampRepository.kt
yschimke: I'll follow up with this, hard to do in the same PR.

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/library/UampRepository.kt
yschimke: removed this

media-sample/src/main/java/com/google/android/horologist/mediasample/di/MediaApplicationContainer.kt
luizgrp: Most of these factory functions are not needed with dagger/hilt, given you just need to add the Inject annotation to the constructor and the Provider implementation is auto generated.

The bindings, like the line above would be simply changed to a signature without implementation:
```
@Binds
fun bindPlaylistRepository(playlistRepositoryImpl: PlaylistRepositoryImpl): PlaylistRepository
```

Which you can see is more aligned with the code added.

The Dispatchers would need a proper named annotation and a factory function or class.

yschimke: Yep, good example, then we can skip those functions for these types.  Just have them for anything with real choices, like a ExoPlayer instances.

luizgrp: If you are concerning about making this DI implementation less confusing, allow me to list one the things that could be improved: this [function](https://github.com/google/horologist/blob/b11174473cd7d30a735b47a002cce24638ea9b9a/media-sample/src/main/java/com/google/android/horologist/mediasample/di/ViewModelModule.kt#L73).

Think of a dagger module, it wouldn't be a place to build a helper function like that, that adds dependencies to an extra class. Check the media toolkit project and see the classes annotated with `@Module`. Do you see that function being placed there? Do you see one of those classes being [injected](https://github.com/google/horologist/blob/b11174473cd7d30a735b47a002cce24638ea9b9a/media-sample/src/main/java/com/google/android/horologist/mediasample/components/MediaActivity.kt#L35) in an activity and being used directly?

I would suggest to start improving the DI solution from there, thinking in how it would be done with a DI lib.

In case a missed the point and that class intention wasn't to be a like a dagger module, then that [provider](https://github.com/google/horologist/blob/b11174473cd7d30a735b47a002cce24638ea9b9a/media-sample/src/main/java/com/google/android/horologist/mediasample/di/ViewModelModule.kt#L54) function shouldn't be there. And maybe that class should be renamed and moved out of the DI package.

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/library/UampRepository.kt
luizgrp: my comment was addressed in https://github.com/google/horologist/pull/492

PR #369, opened 2022-07-12
Add PlaylistRepository

luizgrp:
#### WHAT

Add `PlaylistRepository` to `media-sample`.

#### WHY

- In order to have a better abstraction to support caching and offline first;
- In order to follow the Android architecture guidelines.

#### HOW

- Rename `catalog` package to `data.api`;
- Rename `Catalog` and `Music` to `CatalogUiModel` and `MusicUiModel`;
- Add `Playlist` domain model;
- Add `PlaylistRepository` to `domain` package and `PlaylistRepositoryImpl` to `data` package;
- Add `PlaylistRemoteDataSource`;
- Add mappers;
- Add tests;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-sample/src/main/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSource.kt
+++ b/media-sample/src/main/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSource.kt
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.mediasample.data.datasource
+
+import com.google.android.horologist.mediasample.data.api.UampService
+import com.google.android.horologist.mediasample.data.mapper.PlaylistMapper
+import com.google.android.horologist.mediasample.domain.model.Playlist
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.flowOn
+
+class PlaylistRemoteDataSource(
+    private val ioDispatcher: CoroutineDispatcher,
+    private val uampService: UampService,
+) {
+
+    fun getPlaylists(): Flow<List<Playlist>> = flow {
+        emit(PlaylistMapper.map(uampService.catalog()))
+    }.flowOn(ioDispatcher)
+}
--- a/media-sample/src/main/java/com/google/android/horologist/mediasample/domain/PlaylistRepository.kt
+++ b/media-sample/src/main/java/com/google/android/horologist/mediasample/domain/PlaylistRepository.kt
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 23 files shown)


media-sample/src/test/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSourceTest.kt
```diff
@@ -0,0 +1,55 @@
...
+    @Before
+    fun setUp() {
+        sut = PlaylistRemoteDataSource(
+            ioDispatcher = testDispatcher,
+            uampService = uampService
+        )
+    }
+
+    @Test
+    fun getPlaylists_backedByUampServiceCatalog() = runTest(testDispatcher) {
```
yschimke, line 45: I'm not expert enough in kotlin runTest, but I thought runTest, creates a TestScope, which has a TestDispatcher, for you.

Is it deliberate to create another?

Or should we use TestScope, and then testScope.runTest?

I'm trying to clarify the correct approach, not sure what it is.

media-sample/src/main/java/com/google/android/horologist/mediasample/domain/PlaylistRepository.kt
```diff
@@ -0,0 +1,28 @@
...
+
+package com.google.android.horologist.mediasample.domain
+
+import com.google.android.horologist.mediasample.domain.model.Playlist
+import kotlinx.coroutines.flow.Flow
+
+/**
+ * A repository of [Playlist].
+ */
+interface PlaylistRepository {
```
yschimke, line 25: nit: This seems very specific to one screen, is this the MAD approach, a repository per screen even if they share the same types? Or is this a MediaRepository? 

Not sure of naming, but will we have a different repository for the Media Entity page?

media-sample/src/main/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSource.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.google.android.horologist.mediasample.data.api.UampService
+import com.google.android.horologist.mediasample.data.mapper.PlaylistMapper
+import com.google.android.horologist.mediasample.domain.model.Playlist
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.flowOn
+
+class PlaylistRemoteDataSource(
+    private val ioDispatcher: CoroutineDispatcher,
```
yschimke, line 28: nit: I'm curious about ioDispatcher here. The network layer, retrofit should be executing these correctly on a background thread. This would seem mainly to affect the JSON parsing, is that the intent? If so is the default dispatcher better here?

yschimke (approved):

media-sample/src/main/java/com/google/android/horologist/mediasample/domain/PlaylistRepository.kt
luizgrp: just took a look and the documentation doesn't seem to be specific about the repositories, but I'm usually aligned with [this](https://developer.android.com/topic/architecture/domain-layer#combine-repositories): create one repository per entity and combine them with use cases

let me know what do you think

yschimke: Go for it.

media-sample/src/main/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSource.kt
luizgrp: good question, wouldn't retrofit also take care of executing the JSON parsing in a background thread?
if so, here the dispatcher would be just for the mapper, which we could move to default as suggested

media-sample/src/test/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSourceTest.kt
luizgrp: neither do I, will review it and work on it in another iteration so I don't hold your other prs for too long 👍

media-sample/src/main/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSource.kt
yschimke: Yeah, I suspect you are right. But I will confirm later, since there is some evidence that it isn't always the case.  https://github.com/square/retrofit/pull/3693/files

Issue #276, opened 2022-06-27
App crashes after going back from Audio output screen


luizgrp:
Media sample app freezes and crashes after going back (swipe back without selecting any output) from Audio output screen. This only happens if the Audio output screen is launched after pushing the "play" button on the player screen. If the Audio output button is pushed instead, the issue does NOT happen.

Video:

https://user-images.githubusercontent.com/878134/175915551-57c0af8a-3836-433e-b6ea-a2137dc357b1.mp4

Logs:
```
2022-06-27 10:42:51.090  6644-6644  InputMethodManager      com....horologist.mediasample.debug  I  startInputInner - mService.startInputOrWindowGainedFocus
2022-06-27 10:42:53.000   578-7953  ActivityManager         pid-578                              E  ANR in com.google.android.horologist.mediasample.debug (com.google.android.horologist.mediasample.debug/com.google.android.horologist.mediasample.components.MediaActivity)
                                                                                                    PID: 6644
                                                                                                    Reason: Input dispatching timed out (48d7fbd com.google.android.horologist.mediasample.debug/com.google.android.horologist.mediasample.components.MediaActivity (server) is not responding. Waited 10004ms for FocusEvent(hasFocus=false))
                                                                                                    Parent: com.google.android.horologist.mediasample.debug/com.google.android.horologist.mediasample.components.MediaActivity
                                                                                                    Load: 22.88 / 22.51 / 11.03
```

Device: Galaxy Watch4 Classic (4W2J) - SM-R895F

#### Steps to reproduce

1) Make sure the watch does not have any bluetooth device connected 
2) Launch app - player screen is displayed with a media
3) Push play - bluetooth settings screen is displayed
4) Swipe back

#### Expected result

Player screen is displayed, in an idle/paused state, and app is not frozen.

#### Actual result

Player screen is displayed, in an idle/paused state, but app is frozen (nothing happens when tapping on any buttons) and after few seconds the app closes (crashes).

PR #378, opened 2022-07-13
Fix playback.

yschimke:
#### WHAT

I am not sure why I was using runBlocking here, assuming I was just testing at
first, and left that in.

#### WHY

Playback can't block main thread.  Must return immediately.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
Fixes https://github.com/google/horologist/issues/276

PR #387, opened 2022-07-15
Remove experimental from some fairly stable APIs.

yschimke:
#### WHAT

This stops callers from opting in, so remove experimental from
APIs that have been used enough to have come confidence.

#### WHY

Let callers opt-in to APIs like Accompanist Pager, but also our own, but only when actually changing.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/audio-ui/src/main/java/com/google/android/horologist/audio/ui/semantics/CustomSemanticsProperties.kt
+++ b/audio-ui/src/main/java/com/google/android/horologist/audio/ui/semantics/CustomSemanticsProperties.kt
@@ -19,10 +19,12 @@ package com.google.android.horologist.audio.ui.semantics
 import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.semantics.SemanticsPropertyKey
 import androidx.compose.ui.semantics.SemanticsPropertyReceiver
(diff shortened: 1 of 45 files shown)


audio-ui/src/main/java/com/google/android/horologist/audio/ui/semantics/CustomSemanticsProperties.kt
```diff
@@ -19,12 +19,10 @@ package com.google.android.horologist.audio.ui.semantics
 import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.semantics.SemanticsPropertyKey
 import androidx.compose.ui.semantics.SemanticsPropertyReceiver
-import com.google.android.horologist.audio.ui.ExperimentalHorologistAudioUiApi
 
 /**
  * Custom semantic properties, mainly used for accessibility and testing.
  */
-@ExperimentalHorologistAudioUiApi
```
yschimke, line 27: reverting this

yschimke:
@shumelchyk will land for a release, let me know if you have any concerns.

shumelchyk (commented):
@yschimke I was in the middle of reviewing when you merged it :)
LGTM

PR #388, opened 2022-07-15
Change MarqueeTest to match spec, pause and speed.

yschimke:
#### WHAT

4 second pause time, correct rate.

https://user-images.githubusercontent.com/231923/179220281-2f15b5be-6da3-4d8f-ae4e-1261e002bccc.mp4

#### WHY

Match the UX spec.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
```diff
@@ -47,24 +54,31 @@ private enum class MarqueeComponents {
...
     color: Color = Color.Unspecified,
     style: TextStyle = LocalTextStyle.current,
     textAlign: TextAlign = TextAlign.Left,
-    spacing: Int = 50,
-    edgeGradientWidth: Int = 30,
-    animationTime: Duration = 3.seconds
+    followGap: Dp = 96.dp,
+    edgeGradientWidth: Dp = 10.dp,
+    marqueeDpPerSecond: Dp = 64.dp,
+    pauseTime: Duration = 4.seconds
```
BowerSteve, line 81: It would be good to add some KDocs for these parameters - for example, is pauseTime the delay before marquee animation starts? Does it repeatedly pause and then start again?

composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
```diff
@@ -76,88 +90,121 @@ public fun MarqueeText(
...
+        if (state == AnimationState.Marquee) {
+            edgeGradientWidth - (measuredWidths.text + followGap)
+        } else {
+            edgeGradientWidth
+        }
+    }
 
-    val actualOffset = offset.value.toInt()
+    // Reset animation
+    LaunchedEffect(text) {
```
BowerSteve, line 149: when is this supposed to happen? For the initial animation kick-off, use SideEffect. Seems this would happen every time text changes - so then the animation would never restart? Would you not want the marquee to restart after a text change, in which case reset would be to Pause state.

composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
```diff
@@ -76,88 +90,121 @@ public fun MarqueeText(
...
+            delay(pauseTime)
+            transitionState.targetState = AnimationState.Marquee
+        }
+    }
+
+    fun ContentDrawScope.drawFadeGradient(
+        leftEdge: Boolean,
+    ) {
+        drawRect(
+            size = Size(edgeGradientWidth.toPx(), size.height),
```
BowerSteve, line 172: Create a temporary variable? edgeGradientWidth.toPx() is called 4 times here.

composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
```diff
@@ -76,88 +90,121 @@ public fun MarqueeText(
...
+        }
+
+        if (measuredWidths.isScrollRequired) {
             val secondTextPlaceable = subcompose(MarqueeComponents.Second) {
                 textFn()
             }.first().measure(Constraints())
 
-            val firstElementWidth = textPlaceable.width + spacing
-            val secondTextOffset = firstElementWidth + actualOffset
+            val firstTextOffset = offset
```
BowerSteve, line 219: Not sure about the naming here - isn't the firstOffset/secondOffset more 'offsetTextStart/offsetTextEnd' for the text region as it moves along?

BowerSteve (approved):
Just some minor comments :-)

composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
yschimke: Doesn't SideEffect happen on every recompose? I want this on every text change. I think it's right and it works visually, take a look at the extra code comments I've added in.

composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
yschimke: I've renamed, take a look.

PR #392, opened 2022-07-15
Add EntityScreen

luizgrp:
#### WHAT

Add `EntityScreen`.

Screen Shot 2022-07-15 at 17 45 03

<details>

</details>

<details>

</details>

**Pending** to be implemented:
- Allow customisation (slot param for title / custom chip layouts / custom buttons);
- Tests (snapshots / instrumentation tests);

#### WHY

#### HOW
- Add `DownloadMediaItemUiModel`;
- Add `EntityScreen` implementation;
- Add previews in `debug` folder;
- Add unit tests for `DownloadsState`;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/entity/EntityScreen.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/entity/EntityScreen.kt
@@ -1,0 +1,272 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.media.ui.screens.entity
+
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxSize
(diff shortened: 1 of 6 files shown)


yschimke (approved):
I've raised https://github.com/google/horologist/issues/395 to work out an issue with paparazzi previews for ScalingLazyColumn.

My main thoughts were around the design supporting different content (the title, the buttons, the list) and how you can customise those without abandoning the overall layout structure.  But those are out of scope for the review.  

So LGTM!

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/entity/EntityScreen.kt
```diff
@@ -0,0 +1,272 @@
...
+import com.google.android.horologist.media.ui.components.base.StandardButton
+import com.google.android.horologist.media.ui.components.base.StandardButtonType
+import com.google.android.horologist.media.ui.components.base.StandardChip
+import com.google.android.horologist.media.ui.components.base.Title
+import com.google.android.horologist.media.ui.state.model.DownloadMediaItemUiModel
+import com.google.android.horologist.media.ui.state.model.PlaylistUiModel
+
+@ExperimentalHorologistMediaUiApi
+@Composable
+public fun EntityScreen(
```
yschimke, line 51: nit: This seems more specific than just EntityScreen, but not a problem if we want to change later we we make it more configurable.  Same with things like downloadItemArtworkPlaceholder. Since that is only needed in a StandardChip, if we make that a slot, then we don't need a specific param for it.

yschimke (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/entity/EntityScreen.kt
luizgrp: agreed, I should added these screens in `media-sample` and only move to `media-ui` once they are fully customisable. I will come back to them soon after updating `media-sample` to use them.

Issue #333, opened 2022-07-06
Do not run tests with LargeTest annotation in CI


luizgrp:
#### WHAT

Do not run tests with LargeTest annotation in CI.

#### WHY

It was the intention of https://github.com/google/horologist/pull/304

However, it looks like they are still running as this recent PR [failed](https://github.com/google/horologist/runs/7216301069?check_suite_focus=true) with:
```
com.google.android.horologist.mediasample.playback.PlaybackNotificationTest > testCausesNotification[test(AVD) - 9] FAILED 
	expected to be true
	at com.google.android.horologist.mediasample.playback.PlaybackNotificationTest$testCausesNotification$1$1.invokeSuspend(PlaybackNotificationTest.kt:63)
Tests on test(AVD) - 9 failed: There was 1 failure(s).
```

And the test that failed is [annotated](https://github.com/google/horologist/blob/c0fed820b2d7154a22104e54f52d19c5aefecda8/media-sample/src/androidTest/java/com/google/android/horologist/mediasample/playback/PlaybackNotificationTest.kt#L35-L38) with LargeTest.

luizgrp:
Added a test annotated with LargeTest in https://github.com/google/horologist/pull/401 and confirmed that they are running in CI.
After cherry-picking commit from https://github.com/google/horologist/pull/402 it got green, so confirmed that the solution works and large tests will not run in CI after that commit.

luizgrp:
Reopening as this [build](https://github.com/google/horologist/runs/7424842986?check_suite_focus=true) failed for a test annotated with `LargeTest`: 

```
com.google.android.horologist.media.ui.screens.PlayerScreenTest > givenCustomControlButtons_thenCustomIsDisplayed[test(AVD) - 11] FAILED 
	java.lang.AssertionError: Activity never becomes requested state "[DESTROYED]" (last lifecycle transition = "RESUMED")
	at androidx.test.core.app.ActivityScenario.waitForActivityToBecomeAnyOf(ActivityScenario.java:338)
```

Can see in the logs that the `IGNORE_LARGE_TESTS` is `true` but `SIZE_OPTS` is not being set, so the resulting `gradlew` command does not contain the size filters:
```
IGNORE_LARGE_TESTS=true

+ SIZE_OPTS=
+ [[ -z true ]]
+ [[ ! -z true ]]

[...]

PR #402, opened 2022-07-19
Do not run tests with LargeTest annotation in CI

luizgrp:
#### WHAT

Do not run tests with LargeTest annotation in CI

#### WHY

Fixes https://github.com/google/horologist/issues/333

#### HOW

Pass two params for "size" argument, instead of a single one with two values.

#### Checklist :clipboard:
- [N/A] Run spotless check
- [x] Run tests


yschimke (approved):
Uggghhhh!!!!!

PR #404, opened 2022-07-19
[Not to land] Test with AVD caching.

yschimke:
#### WHAT

Add AVD caching as defined here 

https://github.com/ReactiveCircus/android-emulator-runner

Also in use here https://github.com/coil-kt/coil/blob/b36f9aa53fa3bd1b451abf958e1d3c5aac4013b9/.github/workflows/ci.yml#L71

#### WHY

Speed up CI builds.

#### HOW

Generate AVD and cache across builds.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
Closing, no clear evidence it is faster as is.

PR #412, opened 2022-07-19
Move to Ksp.

yschimke:
#### WHAT

Advised by DevRel to move to Ksp, which would allow a Kotlin 1.7.0 upgrade at some point.
Working for Room, Moshi, but not for Dagger/Hilt.

#### WHY

https://issuetracker.google.com/issues/236612358

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
@shumelchyk I'll follow up with any build comments, but mostly a mechanical switch.

PR #423, opened 2022-07-21
Avoid crashes on network

yschimke:
#### WHAT

Catch exceptions in the Uamp service and show a snackbar.

#### WHY

Avoid crashing when no connection available.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-sample/src/main/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSource.kt
+++ b/media-sample/src/main/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSource.kt
@@ -33,11 +33,13 @@ class PlaylistRemoteDataSource(
     private val errorReporter: ErrorReporter
 ) {
 
-    fun getPlaylists(): Flow<List<Playlist>> = flow {
-        try {
-            emit(PlaylistMapper.map(uampService.catalog()))
+    fun getPlaylists(): Flow<Result<List<Playlist>>> = flow {
+        val result = try {
+            Result.success(PlaylistMapper.map(uampService.catalog()))
         } catch (ioe: IOException) {
             errorReporter.showMessage(R.string.horologist_sample_network_error)
+            Result.failure(ioe)
         }
+        emit(result)
     }.flowOn(ioDispatcher)
 }
--- a/media-sample/src/main/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSource.kt
+++ b/media-sample/src/main/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSource.kt
@@ -16,6 +16,8 @@
 
 package com.google.android.horologist.mediasample.data.datasource
 
+import com.google.android.horologist.media3.logging.ErrorReporter
+import com.google.android.horologist.mediasample.R
 import com.google.android.horologist.mediasample.data.api.UampService
 import com.google.android.horologist.mediasample.data.mapper.PlaylistMapper
 import com.google.android.horologist.mediasample.domain.model.Playlist
@@ -23,13 +25,19 @@ import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.flowOn
+import java.io.IOException
 
 class PlaylistRemoteDataSource(
     private val ioDispatcher: CoroutineDispatcher,
     private val uampService: UampService,
+    private val errorReporter: ErrorReporter
 ) {
 
     fun getPlaylists(): Flow<List<Playlist>> = flow {
-        emit(PlaylistMapper.map(uampService.catalog()))
+        try {
+            emit(PlaylistMapper.map(uampService.catalog()))
+        } catch (ioe: IOException) {
+            errorReporter.showMessage(R.string.horologist_sample_network_error)
+        }
     }.flowOn(ioDispatcher)
 }
--- a/media-sample/src/main/java/com/google/android/horologist/mediasample/system/Logging.kt
+++ b/media-sample/src/main/java/com/google/android/horologist/mediasample/system/Logging.kt
@@ -25,7 +25,7 @@ import com.google.android.horologist.media3.logging.ErrorReporter
 public class Logging(
     private val res: Resources,
     private val snackbarManager: SnackbarManager
-    ) : ErrorReporter {
+) : ErrorReporter {
     override fun showMessage(@StringRes message: Int) {
         val messageString = res.getString(message)
         Log.i("ErrorReporter", messageString)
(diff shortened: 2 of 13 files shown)


media-sample/src/main/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSource.kt
```diff
@@ -16,20 +16,28 @@
...
     private val uampService: UampService,
+    private val errorReporter: ErrorReporter
 ) {
 
     fun getPlaylists(): Flow<List<Playlist>> = flow {
-        emit(PlaylistMapper.map(uampService.catalog()))
+        try {
+            emit(PlaylistMapper.map(uampService.catalog()))
+        } catch (ioe: IOException) {
+            errorReporter.showMessage(R.string.horologist_sample_network_error)
```
luizgrp, line 40: I'm very confused with this ErrorReporter class, it seems to be responsible for logging and for displaying a snackbar.
Assuming that it displays a snackbar, it should not be the responsibility of the data layer to make changes to the UI. The ui state in the viewmodel should be the source of truth of the screen. If we want a snackbar to be displayed, it should be modelled in the ui state, and the view should react to that state. 


Screen Shot 2022-07-21 at 09 24 02

- https://developer.android.com/topic/architecture/ui-layer/events#consuming-trigger-updates
- https://youtu.be/lwGtp0Yr0PE

media-sample/src/main/java/com/google/android/horologist/mediasample/system/Logging.kt
```diff
@@ -19,12 +19,17 @@ package com.google.android.horologist.mediasample.system
...
+public class Logging(
+    private val res: Resources,
+    private val snackbarManager: SnackbarManager
+    ) : ErrorReporter {
     override fun showMessage(@StringRes message: Int) {
-        // TODO show on screen
-        Log.i("ErrorReporter", res.getString(message))
+        val messageString = res.getString(message)
+        Log.i("ErrorReporter", messageString)
+        snackbarManager.showMessage(messageString)
```
luizgrp, line 32: same as previous comment, snackbar should not be called from here, only from the view

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/app/MediaPlayerAppViewModel.kt
```diff
@@ -111,7 +114,7 @@ class MediaPlayerAppViewModel @Inject constructor(
                 playerRepository.setMediaItems(mediaItems)
                 playerRepository.prepare()
             } catch (ioe: IOException) {
-                // Nothing
+                errorReporter.showMessage(R.string.horologist_sample_network_error)
```
luizgrp, line 117: same as previous message, viewmodel should not make a call to display a snackbar directly, it should change the state and let the screen decide how that error message is displayed, either snackbar, text on screen, etc

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/app/MediaPlayerAppViewModel.kt
```diff
@@ -134,17 +137,21 @@ class MediaPlayerAppViewModel @Inject constructor(
...
+            waitForConnection()
 
-        playerRepository.setMediaItems(mediaItems)
-        playerRepository.prepare()
-        playerRepository.play(mediaItemIndex = index)
+            playerRepository.setMediaItems(mediaItems)
+            playerRepository.prepare()
+            playerRepository.play(mediaItemIndex = index)
+        } catch (ioe: IOException) {
+            errorReporter.showMessage(R.string.horologist_sample_network_error)
```
luizgrp, line 153: same as previous message

yschimke:
My go to for reference on good design is Tivi, seems similar here.  https://github.com/chrisbanes/tivi/blob/5c9f8bd5404f213ef5785ccbc42a711c49715b93/ui-search/src/main/java/app/tivi/home/search/SearchViewModel.kt#L69

Do you mind if I land this then open a discussion with the architecture team?  I'll create a tracking bug for it.

To me the Snackbar is designed to allow bridging backend errors with a centralised place to handle displaying errors in the UI.  Surfacing this in each individual Composable that could have an error seems like a lot of additional code, so I'd like to confirm this first?

media-sample/src/main/java/com/google/android/horologist/mediasample/data/datasource/PlaylistRemoteDataSource.kt
yschimke: Also these errors are currently not shown in the screen.  They can be, but we need error states, which we haven't implemented.  Currently they are shown in the Scaffold, which is above this composable or viewmodel.

luizgrp: I don't think the data layer should throw messages that are displayed directly to the user, they should produce an specific exception or return a "result" class which can be success or failure. If there is a message, it should be targeted to the developer, maybe even just an error code. It's up to the UI to decide which message to display and deal with translations etc.

yschimke: Same here https://github.com/android/compose-samples/blob/4ae43c166b33ab82b3e4154e94e6576ec4a0d148/Jetsnack/app/src/main/java/com/example/jetsnack/ui/home/cart/CartViewModel.kt

https://github.com/android/compose-samples/blob/081721ad44dfb29b55b1bc34f83d693b6b8dc9dd/Jetsnack/app/src/main/java/com/example/jetsnack/ui/JetsnackAppState.kt

yschimke: I agree, these errors should eventually be handled inside this screen, without a dialog.

But there are always other errors, say during playback in a background service, that we need to present.  I can add TODOs to say this is temporary until we handle them inside the screens.

But app is broken currently, this makes it functionally better, while still following architecture guidelines.  But it just isn't the final state.

yschimke:
Follow up in https://github.com/google/horologist/issues/425

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/app/MediaPlayerAppViewModel.kt
```diff
@@ -111,7 +117,12 @@ class MediaPlayerAppViewModel @Inject constructor(
                 playerRepository.setMediaItems(mediaItems)
                 playerRepository.prepare()
             } catch (ioe: IOException) {
-                // Nothing
+                snackbarManager.showMessage(
+                    UiMessage(
+                        message = resourceProvider.getString(R.string.horologist_sample_network_error),
```
luizgrp, line 122: nit: there is a [recommendation](https://medium.com/androiddevelopers/locale-changes-and-the-androidviewmodel-antipattern-84eb677660d9) to not retrieve the string from the resources from a viewmodel. The implementation seems ok here but it would be better to avoid as code can be refactored and this matter can be easily forgotten when doing so

PR #424, opened 2022-07-21
Fix player state after resume

yschimke:
#### WHAT

Player State was not being synced correctly after resume.  

TODO - test to confirm the issue

#### WHY

When ExoPlayer was already running, it didn't send events for current state.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-data/src/main/java/com/google/android/horologist/media/data/PlayerRepositoryImpl.kt
+++ b/media-data/src/main/java/com/google/android/horologist/media/data/PlayerRepositoryImpl.kt
@@ -17,7 +17,6 @@
 package com.google.android.horologist.media.data
 
 import android.util.Log
-import androidx.media3.common.PlaybackParameters
 import androidx.media3.common.Player
 import com.google.android.horologist.media.model.Command
 import com.google.android.horologist.media.model.MediaItem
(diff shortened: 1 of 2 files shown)


media-data/src/test/java/com/google/android/horologist/media/data/PlayerRepositoryImplTest.kt
```diff
@@ -147,6 +147,34 @@ class PlayerRepositoryImplTest {
         )
     }
 
+    @Test
+    fun `given is connected after prepared when play until position then state is correct`() {
```
luizgrp, line 151: nit: given is NOT connected and played until position when connect then state is correct 

and maybe just leave the call to `sut.connect` under `// when` block as it is what this test is checking

media-data/src/main/java/com/google/android/horologist/media/data/PlayerRepositoryImpl.kt
```diff
@@ -107,17 +105,29 @@ public class PlayerRepositoryImpl : PlayerRepository, Closeable {
...
+    }
+
+    private fun updateShuffleMode(player: Player) {
+        _shuffleModeEnabled.value = player.shuffleModeEnabled
+    }
+
+    private fun updateCurrentMediaItem(player: Player) {
+        _currentMediaItem.value = player.currentMediaItem?.let(MediaItemMapper::map)
+    }
+
```
luizgrp, line 130: 👍

luizgrp (approved):

media-data/src/test/java/com/google/android/horologist/media/data/PlayerRepositoryImplTest.kt
```diff
@@ -148,7 +148,7 @@ class PlayerRepositoryImplTest {
     }
 
     @Test
-    fun `given is connected after prepared when play until position then state is correct`() {
+    fun `given is NOT connected and played until position when connect then state is correct`() {
```
luizgrp, line 151: Maybe there was a misunderstanding from my previous message, I meant to change to the code to:

```
        player.play()
        playUntilPosition(player, 0, 5.seconds.inWholeMilliseconds)

        // when
        sut.connect(player) {}
```

As it reflects to my suggestion name for the test: "given is NOT connected and played until position when connect then state is correct".

With the changes made, now it looks like the correct name would be: "given is NOT connected when connect AND play until position then state is correct".

PR #434, opened 2022-07-26
Media3 local checkout

yschimke:
#### WHAT

See https://github.com/androidx/media#locally

#### WHY

Use a local checkout when testing with modifications or fixes from main.

#### HOW

Uncomment flags in settings.gradle to import media3.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


settings.gradle
```diff
@@ -45,3 +45,7 @@ include ':network-awareness'
 enableFeaturePreview("VERSION_CATALOGS")
 // https://docs.gradle.org/7.4/userguide/declaring_dependencies.html#sec:type-safe-project-accessors
 enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
+
+// Uncomment to use a local checkout
```
luizgrp, line 49: nit: could have use a flag from gradle.properties to decide to apply this or not. Flag could also be used for function projectOrDependency

luizgrp (approved):

settings.gradle
yschimke: Nice idea. I'm not sure how to do the latter (set a static field?), but first makes sense.

luizgrp: you can refer to it directly in the code:
```groovy
def projectOrDependency(module, dependency = null) {
    if (FLAG) {
        return findProject(module)
    } else {
        return dependency
    }
}
```

if you have defined `FLAG` in `gradle.properties`

yschimke: gradle.properties is checked in though, so easy to accidentally commit the change.  What about from local.properties?

luizgrp: local.properties is specific to android studio as far as I know and is cumbersome to load as you probably noticed, seeing your last commit. Trade-offs, maybe leave a comment on top of the flag to not push to remote in a hope it's spotted by code reviewers when changed? other options [here](https://docs.gradle.org/current/userguide/build_environment.html), like environment variables. Or back to original code? happy with whatever you decide

yschimke: I've used local.properties for similar things, so I'll keep that change for now. It avoids editing files under source control.

PR #441, opened 2022-07-28
Version bumps and Tiles fixes.

yschimke:
#### WHAT

Bump to release or new versions.

#### WHY

More stable dependencies.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
Issue with the Media Tile

yschimke:
I need to look at a layout issue with the tiles still, but will follow up with that.

ataulm (approved):

PR #442, opened 2022-07-28
Fix missed version bumps.

yschimke:
#### WHAT

Bump a couple of dependencies including wear-compose.

#### WHY

Use fianl versions.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


ataulm (approved):

yschimke:
Landing relatively safe version bump.

PR #446, opened 2022-07-29
Refactor Audio Offload Manager

yschimke:
#### WHAT

Allow changing the offload mode without restarting.
Also improve the logging information.
Introduces offload modes Never, Always, Background.

#### WHY

Validate compressed audio offload in testing on different devices. 

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
@kul3r4 it's mostly debugging changes in sample app.  Main changes in library code are offload modes, Never/Always/Background.

kul3r4 (approved):

PR #448, opened 2022-07-29
Fix snapshot images

luizgrp:
#### WHAT

Fix snapshot images

#### WHY

Seeing the following warning:
```
Encountered 3 files that should have been pointers, but weren't:
	media-ui/src/test/snapshots/images/com.google.android.horologist.media.ui.tiles_MediaCollectionsTileTest_mediaCollectionsTile[0]_square.png
	media-ui/src/test/snapshots/images/com.google.android.horologist.media.ui.tiles_MediaCollectionsTileTest_mediaCollectionsTile[1]_small_round.png
	media-ui/src/test/snapshots/images/com.google.android.horologist.media.ui.tiles_MediaCollectionsTileTest_mediaCollectionsTile[2]_large_round.png
```

#### HOW

Not sure if this is the correct approach but just added the automatic changes made locally.

#### Checklist :clipboard:
- [N/A] Run spotless check
- [x] Run tests


yschimke (approved):
Thanks, likely caused by doing operations on my new laptop without lfs installed. Thanks!

PR #450, opened 2022-08-01
Change MediaCollectionsTileService to use PlaylistRepository

luizgrp:
#### WHAT

Change `MediaCollectionsTileService` to use `PlaylistRepository` instead of `UampService` directly.

It now also displays the album name as previously intended:
Screenshot_20220801_102405

#### WHY

In order to follow architecture.
Also it would benefit from not hitting the API if the playlist were previously downloaded - offline first strategy.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Nice, looks great cleanup.

PR #456, opened 2022-08-01
Avoid staying in playing state at end of track.

yschimke:
#### WHAT

Avoid staying in playing state at end of track.

I'm not 100% confident of the product spec, or where to put it in, so for discussion.

#### WHY

Player should stop.

#### HOW

Detect playing and state == ENDED, and move to the start.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


kul3r4 (commented):
Often when the last song from a playlist plays, a music app will start another song from the chosen playlist. 
This is mostly an app level choice, suggestion would be to be able to configure it from the sample app with this default in place: if an app does not implement anything in specific it should not show the wrong state.

yschimke:
Will move this logic to the sample app.

PR #459, opened 2022-08-02
Adjust the tap target for buttons

yschimke:
#### WHAT

Adjust the tap target for buttons. 48x60 for the control buttons.

TODO check with designers.

#### WHY

Avoid being too close to the edge.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/components/animated/AnimatedSeekToPreviousButton.kt
```diff
@@ -57,11 +63,14 @@ public fun AnimatedSeekToPreviousButton(
             AnimatedMediaButton(
                 modifier = modifier,
                 onClick = onClick,
-                contentDescription = stringResource(id = R.string.horologist_seek_to_next_button_content_description),
+                contentDescription = stringResource(id = R.string.horologist_seek_to_previous_button_content_description),
```
luizgrp, line 66: 👍

luizgrp (approved):

PR #460, opened 2022-08-02
Add experimental annotation to classes in media-data

luizgrp:
#### WHAT

Add experimental annotation to classes in `media-data`.

#### WHY

In order to warn developers that they are still in experimental phase.

#### HOW

Add `ExperimentalHorologistMediaDataApi` annotation to classes and opt-in for their usages.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Good catch.

PR #462, opened 2022-08-02
Playback and Nav fixes.

yschimke:
#### WHAT

Make sure play can start from various flows through the app.
Remove PlayerRepository.play(index) and make an explicit call.

#### WHY

Fix playback.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-sample/src/main/java/com/google/android/horologist/mediasample/ui/entity/UampEntityScreen.kt
+++ b/media-sample/src/main/java/com/google/android/horologist/mediasample/ui/entity/UampEntityScreen.kt
@@ -41,7 +41,10 @@ fun UampEntityScreen(
         onDownloadClick = {
             uampEntityScreenViewModel.download()
         },
-        onDownloadItemClick = onDownloadItemClick,
+        onDownloadItemClick = {
+            uampEntityScreenViewModel.play(it.mediaUiModel.id)
+            onDownloadItemClick(it)
+        },
         onShuffleClick = {
             uampEntityScreenViewModel.shufflePlay()
             onShuffleClick(it)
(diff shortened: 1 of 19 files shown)


media/src/main/java/com/google/android/horologist/media/repository/PlayerRepository.kt
```diff
@@ -68,9 +68,11 @@ public interface PlayerRepository {
     public fun play()
 
     /**
-     * Play [Media] at given index as soon as player is ready.
+     * Seeks to the default position associated with the specified MediaItem.
```
luizgrp, line 71: the domain should be agnostic to the implementation details, so there should be no mention to "MediaItem" here, there should be no explanation on how it works but just what's the expected behaviour

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/playlist/PlaylistsScreen.kt
```diff
@@ -37,7 +37,7 @@ import com.google.android.horologist.media.ui.state.model.PlaylistUiModel
 
 @ExperimentalHorologistMediaUiApi
 @Composable
-public fun PlaylistScreen(
+public fun PlaylistsScreen(
```
luizgrp, line 40: the class PlaylistScreenState should also be renamed to PlaylistsScreenState to match

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/entity/UampEntityScreen.kt
```diff
@@ -41,7 +41,10 @@ fun UampEntityScreen(
         onDownloadClick = {
             uampEntityScreenViewModel.download()
         },
-        onDownloadItemClick = onDownloadItemClick,
+        onDownloadItemClick = {
+            uampEntityScreenViewModel.play(it.mediaUiModel.id)
+            onDownloadItemClick(it)
+        },
```
luizgrp, line 47: thanks 👍

luizgrp (approved):

PR #466, opened 2022-08-04
Offload mode testing

yschimke:
#### WHAT

Offload mode testing. Shows a SW/HW in settings screen and also logs to confirm whether offload is enabled.

#### WHY

Make testing offload easier without spelunking through logs to confirm.

#### HOW

Relies on a pending CL for ExoPlayer. Commented out for now.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-sample/src/main/java/com/google/android/horologist/mediasample/di/MediaApplicationModule.kt
+++ b/media-sample/src/main/java/com/google/android/horologist/mediasample/di/MediaApplicationModule.kt
@@ -35,6 +35,7 @@ import com.google.android.horologist.media3.logging.ErrorReporter
 import com.google.android.horologist.media3.navigation.IntentBuilder
 import com.google.android.horologist.media3.navigation.NavDeepLinkIntentBuilder
 import com.google.android.horologist.media3.offload.AudioOffloadManager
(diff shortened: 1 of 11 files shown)


media-sample/src/main/java/com/google/android/horologist/mediasample/di/MediaApplicationModule.kt
```diff
@@ -105,12 +109,31 @@ object MediaApplicationModule {
...
-    ): DefaultAudioSink =
-        wearMedia3Factory.audioSink(
-            attemptOffload = appConfig.offloadEnabled,
-            offloadMode = appConfig.offloadMode
+        wearMedia3Factory: WearMedia3Factory,
+        audioOffloadListener: AudioOffloadListener,
+        settingsRepository: SettingsRepository
+    ): DefaultAudioSink {
+        // TODO check this is basically free at this point
+        val offloadEnabled = runBlocking {
```
yschimke, line 117: This is ugly, but I think required.  Reading the config must happen before this class, which unfortunately happens on the main thread.  But should be already available, most times.

wrkben (approved):

wrkben (approved):

PR #467, opened 2022-08-04
Network Aware Download Listener

yschimke:
#### WHAT

Network Aware Download Listener. Copied from internal project.

#### WHY

Need to request the right network before downloads happen, or they could fail depending on the NetworkRules.

#### HOW

Add a DownloadListener that makes the right calls for DownloadManager.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media3-backend/build.gradle
```diff
@@ -25,7 +25,7 @@ android {
     compileSdkVersion 32
 
     defaultConfig {
-        minSdk 25
+        minSdk 26
```
yschimke, line 28: This bump seems ok, and better than fudging in the manifest.

media3-backend/src/main/java/com/google/android/horologist/media3/service/NetworkAwareDownloadListener.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    override fun onDownloadChanged(
+        downloadManager: DownloadManager,
+        download: Download,
+        finalException: Exception?
+    ) {
+        appEventLogger.logMessage(
+            "download ${download.request.uri.lastPathSegment} ${
+            (download.percentDownloaded).toInt().coerceAtLeast(0)
+            }% ${finalException?.message.orEmpty()}",
```
luizgrp, line 60: nit: I would try to minimise code inside string templates in order to make it more readable

```kotlin
        val name = download.request.uri.lastPathSegment
        val percent = (download.percentDownloaded).toInt().coerceAtLeast(0)
        val error = finalException?.message.orEmpty()
        appEventLogger.logMessage(
            "download $name $percent% $error",
```

luizgrp (approved):

PR #469, opened 2022-08-04
Remove duplicated dependency

luizgrp:
#### WHAT

Remove duplicated dependency

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media3-backend/build.gradle
```diff
@@ -103,7 +103,7 @@ dependencies {
     api projectOrDependency(":media-lib-session", libs.androidx.media3.session)
     implementation libs.androidx.lifecycle.process
     implementation libs.kotlinx.coroutines.guava
-    implementation libs.androidx.corektx
+    implementation libs.androidx.core
```
yschimke, line 106: nit: should the first one be ":core".  Should this be libs.androidx.corektx?

But not against the single dependency.

yschimke (approved):

PR #470, opened 2022-08-04
Add Dispatcher annotation

luizgrp:
#### WHAT

Add `Dispatcher` annotation.

#### WHY

In order to allow the dispatchers injected in our code to be replaced by a dispatcher built with an appropriate executor that we control.

#### HOW

- Add `Dispatcher` annotation and `UampDispatchers` enum;
- Replace direct usages of `Dispatchers.IO` to use injected `CoroutineDispatcher` annotated with `UampDispatcher.IO`

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-sample/src/main/java/com/google/android/horologist/mediasample/di/annotation/Dispatcher.kt
+++ b/media-sample/src/main/java/com/google/android/horologist/mediasample/di/annotation/Dispatcher.kt
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


media-sample/src/main/java/com/google/android/horologist/mediasample/di/annotation/Dispatcher.kt
```diff
@@ -0,0 +1,27 @@
...
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.mediasample.di.annotation
+
+import javax.inject.Qualifier
+
+@Qualifier
+@Retention(AnnotationRetention.RUNTIME)
+annotation class Dispatcher(val uampDispatchers: UampDispatchers)
```
yschimke, line 23: TIL!

yschimke (approved):

Issue #474, opened 2022-08-05
Update metalava for kotlin 1.7


yschimke:
Current errors

```
e: /home/runner/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib/1.7.0/a5f42c684ad9003160ef0d0f693ecf0ba7b13549/kotlin-stdlib-1.7.0.jar!/META-INF/kotlin-stdlib.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.7.1, expected version is 1.5.1.
e: /home/runner/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-stdlib-common/1.7.0/51736992f422993a1e741051bdf3c12801bc1ca1/kotlin-stdlib-common-1.7.0.jar!/META-INF/kotlin-stdlib-common.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.7.1, expected version is 1.5.1.
e: /home/runner/work/horologist/horologist/audio/build/tmp/kotlin-classes/debug/META-INF/audio_debug.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.7.1, expected version is 1.5.1.
```

We held off because of plugin config changed, without obvious migration.

PR #475, opened 2022-08-05
Add media-sync module

luizgrp:
#### WHAT

Add `media-sync` module.

#### WHY

In order to provide a module with common sync mechanism for media apps in Wear.

#### HOW

Adapt code from Now in Android [repository][1], making it more generic given their [model][2] is not open for extension.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

[1]: https://github.com/android/nowinandroid/tree/4a0fbf99b2518384df1960e0045facd456a51539/sync/src/main/java/com/google/samples/apps/nowinandroid/sync

[2]: https://github.com/android/nowinandroid/blob/4a0fbf99b2518384df1960e0045facd456a51539/core-datastore/src/main/java/com/google/samples/apps/nowinandroid/core/datastore/ChangeListVersions.kt#L23-L26


luizgrp:
Failing in [CI](https://github.com/google/horologist/runs/7689505384?check_suite_focus=true) due to https://github.com/google/horologist/issues/474:

```
e: /home/runner/.gradle/caches/transforms-3/b4f2eda692c73bf07eae0853aa41082c/transformed/room-ktx-2.4.3-api.jar!/META-INF/room-ktx_release.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.7.1, expected version is 1.5.1.
```

yschimke:
> Failing in [CI](https://github.com/google/horologist/runs/7689505384?check_suite_focus=true) due to #474:
> 
> ```
> e: /home/runner/.gradle/caches/transforms-3/b4f2eda692c73bf07eae0853aa41082c/transformed/room-ktx-2.4.3-api.jar!/META-INF/room-ktx_release.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.7.1, expected version is 1.5.1.
> ```

Think it's this

```
  - Type 'org.gradle.api.tasks.JavaExec' property '$1' specifies file '/home/runner/work/horologist/horologist/sync/api/current.api' which doesn't exist.
```

Did you check in the api file, or want to skip it?

sync/README.MD
```diff
@@ -0,0 +1,91 @@
...
+        repositoryImpl1,
+        repositoryImpl2,
+    )
+```
+
+#### ChangeListVersionRepository
+
+An implementation of `ChangeListVersionRepository` which can provide a way of retrieving and 
+updating the local change list version for a specific model. A typical implementation will use 
+Android's `SharedPreferences` to store this information.
```
yschimke, line 43: We should probably not point to SharedPreferences, instead maybe DataStore? Discuss offline.

sync/README.MD
```diff
@@ -0,0 +1,91 @@
...
+### Implement Syncable
+
+Repositories that should be synchronized with a remote source should implement interface `Syncable`,
+using helper function `changeListSync` in the `syncWith` function.
+
+The diff between local and remote data should be provided via `NetworkChangeList` based on the 
+current and remote change list versions. If your database does not provide such capability, both the
+diff and `NetworkChangeList` should be computed and built locally.
+
+### Provide configurations via Dagger
```
yschimke, line 18: Is Dagger the actual API, or just the easiest way to configure in a real app?

sync/README.MD
```diff
@@ -0,0 +1,91 @@
...
+
+#### CoroutineDispatcherProvider
+
+An implementation of `CoroutineDispatcherProvider` in order to provide dispatchers of your choice to
+be executing this module's work.
+
+```kotlin
+    @Singleton
+    @Provides
+    fun coroutineDispatcherProvider(): CoroutineDispatcherProvider =
```
yschimke, line 68: Why the indirection of CoroutineDispatcherProvider?

sync/src/main/java/com/google/android/horologist/sync/api/NetworkChangeList.kt
```diff
@@ -0,0 +1,40 @@
...
+ * Network representation of a change list for a model.
+ *
+ * Change lists are a representation of a server-side map like data structure of model ids to
+ * metadata about that model. In a single change list, a given model id can only show up once.
+ */
+public data class NetworkChangeList(
+    /**
+     * The id of the model that was changed
+     */
+    val id: String,
```
yschimke, line 29: Does this assume ids are global, user with id = 4 and track with id = 4 can't exist?

sync/src/main/java/com/google/android/horologist/sync/workers/SyncWorker.kt
```diff
@@ -0,0 +1,89 @@
...
+    private val coroutineDispatcherProvider: CoroutineDispatcherProvider,
+    private val changeListVersionRepository: ChangeListVersionRepository,
+    private val syncables: Array<Syncable>,
+) : CoroutineWorker(appContext, workerParams), Synchronizer {
+
+    override suspend fun getForegroundInfo(): ForegroundInfo =
+        appContext.syncForegroundInfo()
+
+    override suspend fun doWork(): Result = withContext(coroutineDispatcherProvider.getIODispatcher()) {
+        traceAsync("Sync", 0) {
```
yschimke, line 55: nice!

yschimke (commented):
Looks good. It makes more sense now why it might be something we want to offer, particularly if it can make more advanced decisions based on battery, network etc.

I'll try it out, and see how it works for me.

sync/README.MD
luizgrp: It could require more than one `CoroutineDispatcher`, then `CoroutineDispatcherProvider` would have more than one function.

It might feel an overkill as it only requires one dispatcher now, but the function `getIODispatcher` also "documents" via code that the required `CoroutineDispatcher` will be used for IO operations.

I'm inclined to leave it, but no strong feelings if you'd rather simplify it.

sync/src/main/java/com/google/android/horologist/sync/api/NetworkChangeList.kt
luizgrp: the change lists are per "model", see in https://github.com/google/horologist/pull/475/files#diff-d3c791955391f5665c87a7ebbe67c663255928833e0a9b670a3aa3f1635e3dd2R79-R80, so the list will either have "users" or "tracks"

sync/src/main/java/com/google/android/horologist/sync/workers/SyncWorker.kt
luizgrp: [credits](https://github.com/android/nowinandroid/blob/4a0fbf99b2518384df1960e0045facd456a51539/sync/src/main/java/com/google/samples/apps/nowinandroid/sync/workers/SyncWorker.kt#L63) to NiA devs

sync/README.MD
yschimke: Not sure yet.

If every library did this, I'd get annoyed fast.  I think I prefer sensible defaults, Dispatcher.IO is global for a deliberate reason and not considered an anti-pattern.  But I guess you need someway to override for tests?

sync/README.MD
luizgrp: It's the way it's implemented in NiA, [DelegatingWorker](https://github.com/android/nowinandroid/blob/4a0fbf99b2518384df1960e0045facd456a51539/sync/src/main/java/com/google/samples/apps/nowinandroid/sync/workers/DelegatingWorker.kt#L52) delegates to [SyncWorker](https://github.com/android/nowinandroid/blob/4a0fbf99b2518384df1960e0045facd456a51539/sync/src/main/java/com/google/samples/apps/nowinandroid/sync/workers/SyncWorker.kt#L52-L56) which gets all its dependencies injected via `HiltWorkerFactory`.

we could potentially explore the possibility of changing the code and all the configuration passed via the [initialize](https://github.com/android/nowinandroid/blob/4a0fbf99b2518384df1960e0045facd456a51539/sync/src/main/java/com/google/samples/apps/nowinandroid/sync/initializers/SyncInitializer.kt#L31) function; however as per [comments](https://github.com/android/nowinandroid/blob/4a0fbf99b2518384df1960e0045facd456a51539/sync/src/main/java/com/google/samples/apps/nowinandroid/sync/initializers/SyncInitializer.kt#L28-L30), the initialize function is a workaround as they are having [issues](https://github.com/android/nowinandroid/blob/4a0fbf99b2518384df1960e0045facd456a51539/sync/src/main/AndroidManifest.xml#L27) with instrumented tests. The original idea seems to be to have the module [initialized automatically](https://github.com/android/nowinandroid/blob/4a0fbf99b2518384df1960e0045facd456a51539/sync/src/main/AndroidManifest.xml#L23) via [jetpack's InitializationProvider](https://developer.android.com/topic/libraries/app-startup).

yschimke:
It would be really interesting to hang secondary jobs off this.  For example we could preload artwork after a sync, while charging overnight on fast wifi.  

Or download tracks in your favourite playlists.

yschimke:
Or update tiles and complications after sync finishes?

sync/README.MD
yschimke: Not blocking, but I've gone off considering Hilt/Dagger in anyway the API of a library. Be interesting to try fixing init, or just externalising this part of it.  But I don't fully understand the implications yet.

luizgrp: I dislike the idea, but would be too much of change for a v1. Should definitely consider changing it if we decide to publish for external use.

yschimke:
I really like where this is heading, I think keeping it media focused for now allows us to be more opinionated with implementing niceties for media apps.

yschimke (approved):
With planned changes, hopefully including rename to media-sync.

PR #477, opened 2022-08-05
Bump audit plugins, fix issues and run cleanup.

yschimke:
#### WHAT

Fix for metalava, but involves a lot of cleanup when updating dependencies.

#### WHY

https://github.com/google/horologist/issues/474

#### HOW

Updated all audit plugins, and run a clean, and fix issues like filenames.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
I'm tempted to format back with trailing ','.  But not sure.

yschimke:
Removing trailing commas is annoying, but it seems like a firm requirement now.

luizgrp (approved):
thanks!

PR #479, opened 2022-08-09
Settings from Proto DataStore

yschimke:
#### WHAT

Settings from Proto DataStore.

#### WHY

Remove redundant code for dealing with Preferences and instead build on Proto which will become a more important thing.

#### HOW

Adopt protoc in gradle build.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-sample/build.gradle
```diff
@@ -132,6 +156,10 @@ dependencies {
     implementation libs.androidx.wear.tiles
     implementation libs.androidx.wear.tiles.material
 
+    api libs.androidx.datastore.preferences
+    api libs.androidx.datastore
```
luizgrp, line 160: nit: not need to be api as this module is not meant to be used as a dependency?

media-sample/src/main/java/com/google/android/horologist/mediasample/domain/model/UserPreferencesSerializer.kt
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.mediasample.domain.model
+
+import android.content.Context
+import androidx.datastore.core.CorruptionException
+import androidx.datastore.core.DataStore
+import androidx.datastore.core.Serializer
+import androidx.datastore.dataStore
```
luizgrp, line 23: domain layer should be free of framework dependencies, this class should be placed in the data package

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/settings/SettingsScreenViewModel.kt
```diff
@@ -61,43 +62,55 @@ class SettingsScreenViewModel @Inject constructor(
...
         val animated: Boolean = true,
         val debugOffload: Boolean = false,
-        val offloadMode: Settings.OffloadMode = Settings.OffloadMode.Background,
+        val offloadMode: OffloadMode = OffloadMode.BACKGROUND,
         val writable: Boolean = false
     )
 
     fun setShowTimeTextInfo(enabled: Boolean) {
         viewModelScope.launch {
-            settingsRepository.writeShowTimeTextInfo(enabled)
```
luizgrp, line 70: The previous functions were a nicer abstraction for the UI layer.
Now the UI layer knows too much about how the settings are stored, having to make a copy of the settings and change one property. Those details should be done in the data layer. This PR didn't need changes to any UI classes if the layers are done appropriately.

media-sample/src/main/java/com/google/android/horologist/mediasample/domain/model/Settings.kt
```diff
@@ -1,35 +0,0 @@
...
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-package com.google.android.horologist.mediasample.domain.model
-
-import com.google.android.horologist.media3.offload.AudioOffloadStrategy
-import com.google.android.horologist.media3.offload.BackgroundAudioOffloadStrategy
-
-data class Settings(
```
luizgrp, line 22: this class should still exist if we want a domain layer, if we'd rather not have a domain layer, so maybe move SettingsRepository to data package.

media-sample/build.gradle
```diff
@@ -94,23 +95,46 @@ android {
...
+protobuf {
+    protoc {
+        artifact = 'com.google.protobuf:protoc:3.21.4'
+    }
+    plugins {
+        javalite {
+            artifact = 'com.google.protobuf:protoc-gen-javalite:3.0.0'
+        }
+    }
+    generateProtoTasks {
```
luizgrp, line 107: Are there any new tasks that I need to be aware now when working on the project? Just asking before digging in the datastore/protobuf documentation.

luizgrp (approved):
nice one! 

left comments regarding architecture, if we are aiming to keep consistency

media-sample/build.gradle
yschimke: No, should all just build as normal.

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/settings/SettingsScreenViewModel.kt
yschimke: The previous form was awkward to change, too much knowledge about the list of options spread out around the app.  Every new setting required making the same change in multiple places.  

I think we could decide to spend time to introduce some Settings concept, and make it cleaner to wire through. But we didn't have that, and it's not this PR either. We currently don't have any Settings abstraction outside a specific sample app.

But for the purpose of the sample app, this change reduces maintenance burden, while sacrificing a small amount of architectural purity. I think it's worth this tradeoff here.

media-sample/src/main/java/com/google/android/horologist/mediasample/domain/model/Settings.kt
yschimke: If I keep it, it's just a mirror of the existing Settings proto and jsut another thing to keep in sync for little actual benefit here.  Let's refactor this when/if we have some reason like a shared settings concept.

PR #480, opened 2022-08-09
Remove horologist prefix from string resources in media sample

luizgrp:
#### WHAT

Remove horologist prefix from string resources in media sample.

#### WHY

No benefits given the module is not a library.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Good idea!

PR #481, opened 2022-08-10
Add SectionedScreen

luizgrp:
#### WHAT

Add `SectionedScreen`.

Screen Shot 2022-08-10 at 11 25 56

<details>

</details>

<details>

</details>

#### WHY

In order to provide screens that contain sections with individual states.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/sectioned/SectionedScreen.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/sectioned/SectionedScreen.kt
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.media.ui.screens.sectioned
+
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusRequester
+import androidx.wear.compose.material.ScalingLazyColumn
+import androidx.wear.compose.material.ScalingLazyListItemScope
+import androidx.wear.compose.material.ScalingLazyListState
+import com.google.android.horologist.compose.navscaffold.scrollableColumn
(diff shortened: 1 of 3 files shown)


media-ui/src/main/java/com/google/android/horologist/media/ui/screens/sectioned/SectionedScreen.kt
```diff
@@ -0,0 +1,92 @@
...
+        state = scalingLazyListState
+    ) {
+        sections.forEach { section ->
+            section.headerContent?.let { content ->
+                item { content() }
+            }
+
+            when (val sectionState = section.sectionState) {
+                is SectionedScreenSectionState.Loading -> {
+                    item { sectionState.content(this) }
```
yschimke, line 50: I'm a bit concerned if we use each section as an individual `item`, if it spans the whole screen height or more then scaling might be weird.  I don't have a better abstraction to point to, it's possible that this level of reuse isn't well supported.  But it feels worth exploring.

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/sectioned/SectionedScreen.kt
```diff
@@ -0,0 +1,92 @@
...
+@ExperimentalHorologistMediaUiApi
+public data class SectionedScreenSection(
+    val headerContent: (@Composable ScalingLazyListItemScope.() -> Unit)?,
+    val sectionState: SectionedScreenSectionState
+)
+
+@ExperimentalHorologistMediaUiApi
+public sealed class SectionedScreenSectionState {
+
+    public data class Loading(
```
yschimke, line 77: Putting these into data classes doesn't feel like a idiomatic compose approach.  I'm not sure about effect on recompositions etc, whether the equality of a Composable breaks things.  I think they'd tend to expose slots, maybe something to noodle over with the compose team? They may say that this level of reuse is not supported for lazy lists?

Looking at LazyListIntervalContent it uses a non data class, but uses keys, which might help with positional memoization, if we did split each section into multiple items.

yschimke (commented):
No clear comments from me, just some questions to noodle over.

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/sectioned/SectionedScreen.kt
luizgrp: Not sure I understand the issue of having an item in `ScalingLazyColumn` that takes the space of the whole screen height? If the content height is greater than the screen height then it would provide a scroll. Unless you think the scrolling behaviour is undesirable, then we could change the implementation to only use `ScalingLazyColumn` for "loaded" state and another component (e.g. `Box`) taking full screen for the other states?

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/sectioned/SectionedScreen.kt
luizgrp: Drafted a more idiomatic version in https://github.com/google/horologist/pull/487

PR #482, opened 2022-08-10
Datalayer module with Proto and Preferences DataStore

yschimke:
#### WHAT

Datalayer module with Proto and Preferences DataStore.

#### WHY

Familiar abstraction of Androidx DataStore, with a DataLayer implementation.

#### HOW

An early API design likely to change.  Some big TODOs after we have tried this in a few places.

- Use CapabilityClient instead of NodeClient
- Consider switching local/remote to owner/reader
- Support WearableListenerService to handle changes to data.
- Review typical uses and see whether API needs to change.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


datalayer/build.gradle
```diff
@@ -0,0 +1,93 @@
...
+}
+
+dependencies {
+    implementation projects.media
+
+    implementation libs.kotlin.stdlib
+    implementation libs.kotlinx.coroutines.core
+    implementation libs.androidx.wear
+
+    implementation 'com.google.android.gms:play-services-wearable:17.1.0'
```
yschimke, line 76: Will update.

datalayer/src/main/java/com/google/android/horologist/data/store/prefs/PreferencesSerializer.kt
```diff
@@ -0,0 +1,122 @@
...
+            is Boolean -> Value.newBuilder().setBoolean(value).build()
+            is Float -> Value.newBuilder().setFloat(value).build()
+            is Double -> Value.newBuilder().setDouble(value).build()
+            is Int -> Value.newBuilder().setInteger(value).build()
+            is Long -> Value.newBuilder().setLong(value).build()
+            is String -> Value.newBuilder().setString(value).build()
+            is Set<*> -> Value.newBuilder().setStringSet(
+                StringSet.newBuilder().addAllStrings(value as Set<String>)
+            ).build()
+            else -> throw IllegalStateException(
```
fstanis, line 94: Nit: `IllegalArgumentException`

datalayer/src/main/java/com/google/android/horologist/data/store/impl/WearLocalDataStore.kt
```diff
@@ -0,0 +1,104 @@
...
+    private val path: String
+) : DataStore<T> {
+    private val nodeIdFlow = nodeIdFlow().shareIn(coroutineScope, started = started, replay = 1)
+
+    private fun nodeIdFlow() = flow {
+        val nodeId = TargetNodeId.ThisNodeId.evaluate(wearDataLayerRegistry)
+        emit(
+            NodeIdAndPath(
+                nodeId = nodeId,
+                fullPath = buildUri(nodeId, path)
```
fstanis, line 60: Perhaps a bit overly pedantic, but I wonder if `fullPath` should be a getter and `path` a field in the data class.

fstanis (approved):

datalayer/src/main/java/com/google/android/horologist/data/store/prefs/PreferencesSerializer.kt
yschimke: Comes from here, so won't change.

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:datastore/datastore-preferences-core/src/jvmMain/kotlin/androidx/datastore/preferences/core/PreferencesSerializer.kt;l=34

datalayer/src/main/java/com/google/android/horologist/data/store/impl/WearLocalDataStore.kt
yschimke: Let me think it over.  I might end up refactoring this anyway.

datalayer/README.md
```diff
@@ -0,0 +1,31 @@
+# Data Layer library
+
+[![Maven Central](https://img.shields.io/maven-central/v/com.google.android.horologist/horologist-media-data)](https://search.maven.org/search?q=g:com.google.android.horologist)
+
+For more information, visit the documentation: https://google.github.io/horologist/media-data
```
LouisCAD, line 5: This was copy pasted, right?
I can see it's not been updated on the main branch yet.

datalayer/README.md
```diff
@@ -0,0 +1,31 @@
...
+repositories {
+    maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
+}
+
+dependencies {
+    implementation "com.google.android.horologist:horologist-datalayer:<version>-SNAPSHOT"
+}
+```
+
+[snap]: https://oss.sonatype.org/content/repositories/snapshots/com/google/android/horologist/horologist-media-data/
```
LouisCAD, line 31: Same as above.

datalayer/README.md
LouisCAD, line 5: This was copy pasted, right?
I can see it's not been updated on the main branch yet.

LouisCAD, line 5: This was copy pasted, right?
I can see it's not been updated on the main branch yet.

LouisCAD:
I can see the doc website isn't referencing this, despite it being released in 0.1.11. Is that intentional?

datalayer/README.md
yschimke: Thanks.

datalayer/README.md
yschimke: https://github.com/google/horologist/pull/599

PR #483, opened 2022-08-10
Fixup metalava

yschimke:
#### WHAT

Fixup metalava

#### WHY

Needed to upgrade for Kotlin 1.7, but not running with 0.2.3.

#### HOW

Workaround until issue is fixed.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/network-awareness/api/current.api
+++ b/network-awareness/api/current.api
@@ -1,7 +1,7 @@
 // Signature format: 4.0
 package com.google.android.horologist.networks {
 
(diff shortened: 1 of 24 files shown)


network-awareness/api/current.api
```diff
@@ -333,6 +333,10 @@ package com.google.android.horologist.networks.okhttp {
     field public static final com.google.android.horologist.networks.okhttp.AlwaysHttpsInterceptor INSTANCE;
   }
 
+  public final class AwaitKt {
```
yschimke, line 336: Im ok with these changes as they won't have source compatibility issues.

PR #489, opened 2022-08-12
Fix package name for PlaylistsScreen

luizgrp:
#### WHAT

Fix package name for PlaylistsScreen

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (commented):
LGTM

PR #491, opened 2022-08-15
Reorganise screen packages

luizgrp:
#### WHAT

Reorganise screen packages

#### WHY

In order to keep it consistent with the package organisation of other screens.

#### HOW

- Move `PlayerScreen` and `PlayerLibraryPagerScreen` into their own package;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (commented):
LGTM

PR #492, opened 2022-08-15
Add backend folder

luizgrp:
#### WHAT

Add backend folder to media-sample module.

#### WHY

In order to make explicit that these assets are not used by the app code but just used as a replacement for a lack of functionality in our backend.

#### HOW

- Add backend folder outside source folder;
- Move files from source folder to backend folder;
- Add 3 extra images that were added recently in the backend;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (commented):
LGTM

PR #494, opened 2022-08-16
Workaround to fix play item by index

luizgrp:
#### WHAT

Workaround to fix play item by index until issue https://github.com/androidx/media/issues/85 is fixed in media3.

Fixes https://github.com/google/horologist/issues/495

#### WHY

So the app can play the specific song selected by the user.

#### HOW

- Add a delay after call to `setMediaList`

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (commented):
Looks like this is no longer relevant.

Issue #495, opened 2022-08-16
Media sample doesn't play song at specific index


luizgrp:
Due to https://github.com/androidx/media/issues/85

PR #497, opened 2022-08-16
Add setMediaListAndPlay to PlayerRepository

luizgrp:
#### WHAT

Add `setMediaListAndPlay` to `PlayerRepository` with a workaround implementation for #495.

#### WHY

In order to be able to play an item at specific index.
In order to be able to set the media list and seek to a specific index without having an intermediate emission from `PlayerRepository.mediaPosition`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-data/src/test/java/com/google/android/horologist/media/data/repository/PlayerRepositoryImplTest.kt
+++ b/media-data/src/test/java/com/google/android/horologist/media/data/repository/PlayerRepositoryImplTest.kt
@@ -39,6 +39,7 @@ import com.google.common.truth.Truth.assertThat
 import org.junit.Assert.assertThrows
 import org.junit.Assert.assertTrue
 import org.junit.Before
+import org.junit.Ignore
 import org.junit.Test
 import org.junit.runner.RunWith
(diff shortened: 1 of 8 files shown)


media-data/src/test/java/com/google/android/horologist/media/data/repository/PlayerRepositoryImplTest.kt
```diff
@@ -564,6 +565,36 @@ class PlayerRepositoryImplTest {
...
+        val player = TestExoPlayerBuilder(context).build()
+        sut.connect(player) {}
+
+        val media1 = getStubMedia("id1")
+        val media2 = getStubMedia("id2")
+        val mediaList = listOf(media1, media2)
+
+        // when
+        sut.setMediaListAndPlay(mediaList, 1)
+        runUntilPendingCommandsAreFullyHandled(player)
```
yschimke, line 581: How do we know we can rely on this still? Does runUntilPendingCommandsAreFullyHandled, wait until the items are resolved and then we get onTimelineChangedcalled on the main thread?

PR #503, opened 2022-08-17
Improve loadingContent param from EntityScreen

luizgrp:
#### WHAT

Improve `loadingContent` param from `EntityScreen`.

Screen Shot 2022-08-17 at 14 57 35 Screen Shot 2022-08-17 at 14 57 29

#### WHY

In order to allow further customisation. 
The previous implementation was calling the `mediaLoadingContent` param twice, now it's up to the developer to decide how it should look like.

#### HOW

- Rename `mediaLoadingContent` to `loadingContent` in `EntityScreen` and change type to be a lambda executed with `ScalingLazyListScope`;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (commented):
Nice, looks good.

PR #505, opened 2022-08-19
Display progress of media downloads

luizgrp:
#### WHAT

Display progress of media downloads in `PlaylistDownloadScreen`.

https://user-images.githubusercontent.com/878134/185630854-0d641585-6f69-4a32-a215-4fd4d5dc6344.mp4

#### WHY

In order to align with Figma specs.

#### HOW

- Add `progress` and `size` to `MediaDownloadEntity`;
- Add `DownloadProgressMonitor` to query `DownloadManager` every specified interval and update the download progress with `MediaDownloadLocalDataSource`;
- Improve `DownloadManagerListener` to start and stop `DownloadProgressMonitor`;

#### Checklist :clipboard:
- [x] Add explicit visibility modifier and explicit return types for public declarations
- [x] Run spotless check
- [x] Run tests
- [x] Update metalava's signature text files

PR #502, opened 2022-08-17
Make EntityScreen generic and add PlaylistDownloadScreen

luizgrp:
#### WHAT

Make `EntityScreen` more customizable;

Screen Shot 2022-08-17 at 10 17 56

<details>
<summary>Small rounded devices preview</summary>

</details>

<details>
<summary>Square device previews</summary>

</details>

Add `PlaylistDownloadScreen`, a specific implementation of `EntityScreen`.

Screen Shot 2022-08-16 at 13 28 40

#### WHY

In order to provide common screens.

#### WHAT

- Add a stateless and stateful implementation of `EntityScreen`;
- Add `PlaylistDownloadScreen`;
- Add previews in debug folder;

#### Checklist :clipboard:
- [x] Add explicit visibility modifier and explicit return types for public declarations
- [x] Run spotless check
- [x] Run tests
- [x] Update metalava's signature text files

PR #506, opened 2022-08-19
Rename string resources to match screen name

luizgrp:
#### WHAT

1. Rename string resources to match screen name when `PlaylistDownloadScreen` was extracted out from `EntityScreen` in https://github.com/google/horologist/pull/502.

2. Push down logic from `UampEntityScreenViewModel` to `createPlaylistDownloadScreenStateLoaded` function for checking if there is any download in progress - follow up from https://github.com/google/horologist/pull/505

Screen Shot 2022-08-19 at 15 25 09

#### WHY

1. Avoid confusion.

2. Small optimisation.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (commented):
Good idea!

PR #507, opened 2022-08-21
Update segmented progress indicator to use canvas api directly

oas004:
#### WHAT
In this PR I have update the SegmentedProgressIndicator to use the Canvas APIs directly.

#### WHY
The reasoning behind this is that we don't want to draw multiple canvases on top of each other.

#### Screenshots

Here the total weight is 6
Screenshot 2022-08-21 at 17 42 38

Here the total weight is 3
Screenshot 2022-08-21 at 17 44 08

#### Checklist :clipboard:
- [x] Add explicit visibility modifier and explicit return types for public declarations
- [x] Run spotless check
- [x] Run tests
- [x] Update metalava's signature text files

Issue #431, opened 2022-07-22
SquareSegmentedProgressIndicator


garanj:
Create a `SquareSegmentedProgressIndicator` (or `RectangularSegmentedProgressIndicator`)

yschimke:
Would we use Compose Canvas for this?

See https://google.github.io/horologist/composables/#segmented-progress-indicator

garanj:
Yes, and I think it probably worth moving the existing segmented one to using canvas as opposed to multiple progress indicators as it currently does.

oas004:
Hi, is there a design on this? 😄  I tried to implement this to be like a rectangular version of the `SegmentedProgressIndicator`, but it looked a bit strange. Furthermore, is there a plan to also have a `LinearSegmentedProgressIndicator` or am I misunderstanding something here? 😅

yschimke:
Hey, no official design. The idea was to copy the style of th Circular design closely. It likely won't get attention at the moment as an official design, but I can check.  Can you attach a screenshot and put up a PR with your current attempt?

If it really won't work with the current design, we should probably close this.

oas004:
Gave this another shot, and I managed to get to this here:


|   progress = 0f	|   progress = 0.375f	|   progress = 0.6875f	| 
|---	|---	|---	|
|   ![Screenshot 2022-10-16 at 19 00 56](https://user-images.githubusercontent.com/54936943/196048363-ca820a9a-a613-4096-aaa4-8b0ce9ade8d6.png)	|   ![Screenshot 2022-10-16 at 19 01 15](https://user-images.githubusercontent.com/54936943/196048368-d18396e7-1a21-489d-9f61-a144b6e09f8d.png)	|   ![Screenshot 2022-10-16 at 19 02 00](https://user-images.githubusercontent.com/54936943/196048370-e53ee49f-c1be-4b8c-bc62-b9ae8dc87797.png)	| 

I choose to expose a progress as a float value and divide the section based on the length of the circumference instead of dealing with the angles as I first tried. 
[...]

PR #508, opened 2022-08-21
[WIP] Make SquareSegmentedProgressIndicator

oas004:
#### WHAT
I have tried to make the Square version of the (Circular)SegmentedProgressIndicator

#### WHY
Because of #431 

#### HOW
I have been trying to make the SquareSegmentedProgressIndicator, but I am kinda stuck.  I think it will work with a rounded rectangular shape, but if we are going to expose the corner radius, and the consumer would set that to 0 I think we will have an issue.

I tried to follow the same approach as the (Circular)SegmentedProgressIndicator. However, the drawRoundedRectangle  function does not take a start nor a sweep angle. So what I did was to use the drawPath  functionality of the Canvas. So based on the currentStartAngle and the sweepAngle I’m calculating if I should draw an arch or a line. For the line starting point and ending point, I’m using Square to Disc interpolation to map the coordinates as by described in [this](https://arxiv.org/abs/1509.06344). I’m having some problems getting it to work correctly with the endDelta for the breaking point between the lines and to get the calculation correct.

I would really like some guidance on this if I'm on the correct track here.

If this can't work, feel free to close this.

Also, I tried to make this a nested PR from #507, but it seems that I can't change the target branch to that from the forked repository 🤔 Is there a way to do that in open source repos?

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
@@ -1,0 +1,348 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.composables
+
+import androidx.annotation.VisibleForTesting
+import androidx.compose.foundation.Canvas
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.progressSemantics
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.geometry.Rect
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Path
+import androidx.compose.ui.graphics.StrokeCap
+import androidx.compose.ui.graphics.drawscope.DrawScope
+import androidx.compose.ui.graphics.drawscope.Stroke
+import androidx.compose.ui.layout.onSizeChanged
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.tooling.preview.Devices
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import androidx.wear.compose.material.MaterialTheme
+import androidx.wear.compose.material.ProgressIndicatorDefaults
+import java.lang.Float.max
+import kotlin.math.PI
+import kotlin.math.abs
+import kotlin.math.asin
+import kotlin.math.cos
+import kotlin.math.min
+import kotlin.math.sin
+import kotlin.math.sqrt
+
+/**
+ * Represents a square segmented progress indicator.
+ *
+ * @param modifier [Modifier] to be applied to the [SegmentedProgressIndicator]
+ * @param trackSegments A list of [ProgressIndicatorSegment] definitions, specifying the properties
+ * of each segment.
+ * @param progress The progress of this progress indicator where 0.0 represents no progress and 1.0
+ * represents completion. Values outside of this range are coerced into the range 0..1.
+ * @param startAngle The starting position of the progress arc,
+ * measured clockwise in degrees (0 to 360) from the 3 o'clock position. For example, 0 and 360
+ * represent 3 o'clock, 90 and 180 represent 6 o'clock and 9 o'clock respectively.
+ * Default is -90 degrees (top of the screen)
+ * @param endAngle The ending position of the progress arc,
+ * measured clockwise in degrees (0 to 360) from the 3 o'clock position. For example, 0 and 360
+ * represent 3 o'clock, 90 and 180 represent 6 o'clock and 9 o'clock respectively.
+ * By default equal to 270 degrees.
+ * @param strokeWidth The stroke width for the progress indicator.
+ * @param paddingAngle The gap to place between segments. Defaults to 0 degrees.
+ * @param trackColor The background track color. If a segment specifies [trackColor] then the
+ * segment value takes preference. Defaults to [Color.Black]
+ */
+@ExperimentalHorologistComposablesApi
+@Composable
+public fun SquareSegmentedProgressIndicator(
+    modifier: Modifier = Modifier,
+    trackSegments: List<ProgressIndicatorSegment>,
+    progress: Float,
+    startAngle: Float = -90.0f,
+    endAngle: Float = 270.0f,
+    strokeWidth: Dp = ProgressIndicatorDefaults.StrokeWidth,
+    paddingAngle: Float = 0.0f,
+    trackColor: Color = MaterialTheme.colors.onBackground.copy(alpha = 0.1f)
+) {
+    var boxMinDimension by remember { mutableStateOf(0) }
+    Box(
+        contentAlignment = Alignment.Center,
+        modifier = Modifier.onSizeChanged { boxMinDimension = min(it.width, it.height) }
+    ) {
+        val localDensity = LocalDensity.current
+
+        val boxMinDimensionDp = with(localDensity) {
+            boxMinDimension.toDp()
+        }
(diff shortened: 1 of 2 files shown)


composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
```diff
@@ -0,0 +1,348 @@
...
+            if ((225 + cornerRadius / 2..315 - cornerRadius / 2).isAnElementOf(angleSection) || (-135 + cornerRadius / 2..-45 - cornerRadius / 2).isAnElementOf(angleSection)) {
+
+                // The angle within we are defining the line
+                val lineAngleSection = (315 - cornerRadius / 2) - (225 + cornerRadius / 2)
+                val actualLineSection = min(lineAngleSection.toFloat(), backgroundSweep)
+
+                val actualLineSectionStart = max(abs(currentStartAngle), 225 + cornerRadius / 2f)
+
+                moveTo(left + (cornerRadius / 2f) + transformAngleToCoordinate(angle = actualLineSectionStart, width = width, height = height).first , top)
+                lineTo(left + (cornerRadius / 2f) + transformAngleToCoordinate(angle = actualLineSection, width = width, height = height).first, top)
```
oas004, line 217: I would have to do this for each line on the square as well I think

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
```diff
@@ -0,0 +1,348 @@
...
+                    Rect(
+                        left = left,
+                        top = top,
+                        right = left + cornerRadius,
+                        bottom = top + cornerRadius
+                    ),
+                    180f,
+                    min(
+                        backgroundSweep, 90f
+                    ),
```
oas004, line 203: This I'm very unsure of would work 🤔

yschimke:
Can you add a paparazzi test, and checkin the file, it helps with understanding the change.

Similar to https://github.com/google/horologist/pull/507#issuecomment-1221558496

yschimke:
Instructions for updating files are here https://github.com/google/horologist/blob/main/CONTRIBUTING.md

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
```diff
@@ -0,0 +1,348 @@
...
+        val endDelta = remember(strokeWidth, boxMinDimensionDp) {
+            (
+                asin(
+                    strokeWidth.value /
+                        (boxMinDimensionDp.value - strokeWidth.value)
+                ) * 180 / Math.PI
+                ).toFloat()
+        }
+        val totalWeight = remember(trackSegments) {
+            trackSegments.sumOf { it.weight.toDouble() }.toFloat()
```
yschimke, line 109: why toDouble then toFloat?

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
```diff
@@ -0,0 +1,348 @@
...
+                remainingProgress -= segmentAngle
+            }
+        }
+    }
+}
+
+private fun DrawScope.drawRectangularProgressIndicator(
+    segment: ProgressIndicatorSegment,
+    currentStartAngle: Float,
+    endDelta: Float,
```
yschimke, line 165: Assume this is very much in progress, not currently using endDelta.

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
```diff
@@ -0,0 +1,348 @@
...
+ * @param trackColor The background track color. If a segment specifies [trackColor] then the
+ * segment value takes preference. Defaults to [Color.Black]
+ */
+@ExperimentalHorologistComposablesApi
+@Composable
+public fun SquareSegmentedProgressIndicator(
+    modifier: Modifier = Modifier,
+    trackSegments: List<ProgressIndicatorSegment>,
+    progress: Float,
+    startAngle: Float = -90.0f,
```
yschimke, line 82: I wonder if angle is the most natural way to express the start and end points? A percent would be natural for a square, but may be problematic for rectangle.

I'm not sure of the answer, but do wonder based on how we expect it to be used, whether there are simpler ways?

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
```diff
@@ -0,0 +1,348 @@
...
+ * represent 3 o'clock, 90 and 180 represent 6 o'clock and 9 o'clock respectively.
+ * By default equal to 270 degrees.
+ * @param strokeWidth The stroke width for the progress indicator.
+ * @param paddingAngle The gap to place between segments. Defaults to 0 degrees.
+ * @param trackColor The background track color. If a segment specifies [trackColor] then the
+ * segment value takes preference. Defaults to [Color.Black]
+ */
+@ExperimentalHorologistComposablesApi
+@Composable
+public fun SquareSegmentedProgressIndicator(
```
yschimke, line 78: Does it complicate things greatly if we make it RectangularSegmentedProgressIndicator instead?

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
oas004: Yes, my plan was to calculate if the progress angle if that is within the angle of the section and then adding or subtracting endDelta to the starting and ending point of the line. I guess I would do some magic if it is an arch as well 🤔

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
oas004: The disc to square interpolation of the starting and ending point of the square should take the width and height into account, so I don't think so, but It might be some aspect I don't see here 🤔 

Feels a bit like Im over complicating things in this solution

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
oas004: Yeah I have that feeling as well that there should be an easier way. Maybe a percentage solution would be better 🤔

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
oas004: yeah, sorry my bad. Will remove that

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
```diff
@@ -0,0 +1,348 @@
...
+                lineTo(left, bottom - (cornerRadius / 2f) - diameter + cornerRadius)
+            //}
+
+        },
+        style = stroke,
+    )
+}
+
+@VisibleForTesting
+fun IntRange.isAnElementOf(section: IntRange): Boolean {
```
yschimke, line 294: Can this be simpler?

`(this.start <= section.endInclusive) && (section.start <= this.endInclusive)`

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
```diff
@@ -0,0 +1,348 @@
...
+    val angleSection = currentStartAngleInt..endAngle
+
+    drawPath(
+        color = segment.trackColor ?: trackColor,
+        path = Path().apply {
+
+            // top left
+            moveTo(left, (top + cornerRadius))
+            if ((225 - cornerRadius / 2..225 + cornerRadius / 2).isAnElementOf(angleSection)) {
+                arcTo(
```
yschimke, line 193: I suspect instead of arcTo with angles, we might have a simpler solution just treating the progress as vertical or horizontal distances, and handling the corners carefully. But I don't know for sure.

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
oas004: Yeah that looks way better!

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
oas004: Will there then always be a distance between the vertical and horizontal lines then? 🤔 Just thinking since then because of the rounded strokes it might be weird if they meet? Or what do you think?

yschimke: Not sure, I can try a few ideas.  Maybe drawing each line as either 1 to 3 sections (side, corner, side), with some minimums to draw them, and caps only on the end pieces?

oas004: Yeah maybe that could work 🤔 😄

oas004:
I'm not getting anywhere with this. I can't get it to work. Sorry about this. I it okay if I close it @yschimke ? I can probably re-open it if come up with a major breakthrough or something? Again, sorry for wasting your time.

yschimke:
Of course. No need to apologise.

PR #509, opened 2022-08-21
Use Material Chip for the secondary placeholder chip

### LINKED ISSUE #332: Change implementation of SecondaryPlaceholderChip to use Chip
#### WHAT

Change implementation of [SecondaryPlaceholderChip](https://github.com/google/horologist/blob/c0fed820b2d7154a22104e54f52d19c5aefecda8/media-ui/src/main/java/com/google/android/horologist/media/ui/components/base/SecondaryPlaceholderChip.kt) to use [Chip](https://github.com/androidx/androidx/blob/5775cb72bdc32a90328bf39e212fab31c5960316/wear/compose/compose-material/src/commonMain/kotlin/androidx/wear/compose/material/Chip.kt#L282) instead of building everything from scratch, and pass the grey shapes as params to `label`, `secondaryLabel`, and `icon`.

PS: remember to remove this [dependency](https://github.com/google/horologist/blob/c0fed820b2d7154a22104e54f52d19c5aefecda8/media-ui/build.gradle#L107) from the module when done.

#### WHY

https://github.com/google/horologist/pull/328#discussion_r914674450

#### Expected outcome from this task

- Implementation of `SecondaryPlaceholderChip` is changed to use `Chip`;
- Preview for `SecondaryPlaceholderChip` is not affected;
- Snapshot test for `SecondaryPlaceholderChip` is not affected;

oas004:
First of all, thank you so much for this cool library! I really like it! 😃 🤩 

#### WHAT
In this PR I have started to use the material chip apis for the secondary placeholder chip instead of making it from basic composable as it was before 😃 

#### WHY
fixes #332 

#### HOW
I have changed the implementation and seen that the screenshot test that was used on the old implementation still passes.

Screenshot 2022-08-22 at 15 16 07

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/build.gradle
+++ b/media-ui/build.gradle
@@ -115,9 +115,6 @@ dependencies {
     implementation libs.compose.ui.tooling
     implementation libs.compose.material.iconscore
     implementation libs.compose.material.iconsext
-    implementation(libs.compose.material.ripple) {
-        because "it is needed for SecondaryPlaceholderChip. Remove it once this is done: https://github.com/google/horologist/issues/332"
-    }
     implementation libs.androidx.lifecycle.viewmodel.compose
 
     implementation libs.coil
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/components/base/SecondaryPlaceholderChip.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/components/base/SecondaryPlaceholderChip.kt
@@ -17,31 +17,22 @@
 package com.google.android.horologist.media.ui.components.base
 
 import androidx.compose.foundation.background
-import androidx.compose.foundation.clickable
-import androidx.compose.foundation.interaction.MutableInteractionSource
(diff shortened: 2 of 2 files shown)


oas004:
I don't quite understand why this test now fails. Locally, the test passes with 0.1% difference, but on CI it fails with the same amount of difference 🤔 The test also states that 0.1% difference is max

yschimke:
The diff seems to show a real issue

image

This is with the Android Studio "Image Diff" plugin https://plugins.jetbrains.com/plugin/12691-image-diff

oas004:
> The diff seems to show a real issue
> 
> image
> 
> This is with the Android Studio "Image Diff" plugin https://plugins.jetbrains.com/plugin/12691-image-diff

Oo thanks! Have to use that plugin! Converting to draft until I get it perfect 😄

luizgrp (approved):
LGTM - thanks for fixing this! 👍

PR #511, opened 2022-08-22
Add SectionedList

luizgrp:
#### WHAT

Add `SectionedList`.

Screen Shot 2022-08-12 at 10 48 10

<details>

</details>

<details>

</details>

#### WHY

In order to provide a list component that contain sections with individual states.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/components/list/sectioned/SectionedList.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/components/list/sectioned/SectionedList.kt
@@ -1,0 +1,218 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


media-ui/src/main/java/com/google/android/horologist/media/ui/components/list/sectioned/SectionedList.kt
```diff
@@ -0,0 +1,218 @@
...
+     * A state of a [Section].
+     */
+    public sealed class State {
+        public object Loading : State()
+
+        public data class Loaded<T>(
+            val list: List<T>
+        ) : State()
+
+        public object Failed : State()
```
yschimke, line 114: Should failed have some detail? An exception? or resource string id?

yschimke (approved):
Hard to review in PR, land and let's throw some things at it.

media-ui/src/main/java/com/google/android/horologist/media/ui/components/list/sectioned/SectionedList.kt
luizgrp: the content for this state is totally open to the dev, see usage [here](https://github.com/google/horologist/pull/511/files#diff-fa0ba882cae23e52395390f8f9bcfb38bf9753e13b767f9d728445adee795035R105) as example, so there is no need for params

the "Loaded" state has the list param as the content provided by the user is per item 

hope that clarifies, otherwise happy to explain over a chat

PR #513, opened 2022-08-23
Change PlaylistsScreen to use SectionedList

luizgrp:
#### WHAT

Change `PlaylistsScreen` to use `SectionedList`.

Preview after changes:

Screen Shot 2022-08-23 at 10 37 05

#### WHY

In order to reuse functionality implemented in `SectionedList`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/playlists/PlaylistsScreen.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/playlists/PlaylistsScreen.kt
@@ -16,22 +16,22 @@
 
 package com.google.android.horologist.media.ui.screens.playlists
 
-import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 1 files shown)


media-ui/src/main/java/com/google/android/horologist/media/ui/screens/playlists/PlaylistsScreen.kt
```diff
@@ -61,34 +61,37 @@ public fun <T> PlaylistsScreen(
...
+                    textId = R.string.horologist_browse_playlist_title,
+                    modifier = Modifier.padding(bottom = 12.dp)
+                )
             }
-            is PlaylistsScreenState.Loading<T> -> {
-                items(count = 4) {
+
+            loaded { playlistContent(it) }
+
+            loading {
```
yschimke, line 87: I do wonder whether loading should have some possible state, like completion percent?

luizgrp: it's up to the dev to pass the full content, adding that would make the API less flexible, IMO

yschimke (approved):

PR #521, opened 2022-08-25
Update the DataStore functionality

yschimke:
#### WHAT

Update the DataStore functionality.

Improve to support any proto, and also wake up via WearableListenerService.

#### WHY

Simplify using Data Layer for critical events such as updating a tile.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
@fstanis Going to land this as is, not currently being used.  I'm building on top of it. Let me know any feedback and I'll follow up.

PR #527, opened 2022-08-29
Move Playlist, PlaylistDownload and repositories to media module

luizgrp:
#### WHAT

Move `Playlist`, `PlaylistDownload` and repositories to media module.

#### WHY

In order to be part of the media library.

#### HOW

Move classes, add experimental annotation and improve kdocs.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-sample/src/main/java/com/google/android/horologist/mediasample/ui/mapper/DownloadMediaUiModelMapper.kt
```diff
@@ -37,16 +37,20 @@ object DownloadMediaUiModelMapper {
         }
 
         is MediaDownload.Status.InProgress -> {
+            val mediaDownloadStatus = (mediaDownload.status as MediaDownload.Status.InProgress)
```
yschimke, line 40: Why do you need this cast? Shouldn't type inference work?

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/mapper/DownloadMediaUiModelMapper.kt
```diff
@@ -37,16 +37,20 @@ object DownloadMediaUiModelMapper {
...
+                progress = if (mediaDownloadStatus.progress == PROGRESS_WAITING) {
                     DownloadMediaUiModel.Progress.Waiting
                 } else {
-                    DownloadMediaUiModel.Progress.InProgress(PROGRESS_FORMAT.format(mediaDownload.status.progress))
+                    DownloadMediaUiModel.Progress.InProgress(PROGRESS_FORMAT.format(mediaDownloadStatus.progress))
                 },
                 size = when (mediaDownload.size) {
-                    is MediaDownload.Size.Known -> DownloadMediaUiModel.Size.Known(mediaDownload.size.sizeInBytes)
+                    is MediaDownload.Size.Known -> {
+                        val mediaDownloadSize = mediaDownload.size as MediaDownload.Size.Known
```
yschimke, line 51: Same, why is the cast needed?

yschimke (approved):

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/mapper/DownloadMediaUiModelMapper.kt
luizgrp: unfortunately kotlin can't do a smart cast when the class is defined in another module, more info in this thread:
https://discuss.kotlinlang.org/t/what-is-the-reason-behind-smart-cast-being-impossible-to-perform-when-referenced-class-is-in-another-module/2201/2

PR #543, opened 2022-09-01
Updates SegmentedProgressIndicator to fix re-composition issues

garanj:
Issues in rendering could be seen on emulator or device in some cases where the Canvas was being recomposed, but the starting calculations of initial angles not recalculated.

Tested on device and emulator

#### WHAT

SegmentedProgressIndicator could in some cases render incorrectly if the Canvas was recomposed but the outer SegmentedProgressIndicator was not

#### WHY

Values used in the rendering of each segment were initialized outside of the Canvas call, and updated as each segment was drawn. Subsequent recomposition did not reinitialize these.

Moved the initialization into the Canvas call, to ensure that calls to it have no side effects.

#### Checklist :clipboard:
- [ n/a] Add explicit visibility modifier and explicit return types for public declarations
- [ x] Run spotless check
- [ x] Run tests
- [ n/a] Update metalava's signature text files

### DIFF
--- a/composables/src/main/java/com/google/android/horologist/composables/SegmentedProgressIndicator.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/SegmentedProgressIndicator.kt
@@ -17,12 +17,10 @@
 package com.google.android.horologist.composables
 
 import androidx.compose.foundation.Canvas
-import androidx.compose.foundation.layout.BoxWithConstraints
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.progressSemantics
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
-import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.geometry.Size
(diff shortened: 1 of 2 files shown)


composables/src/main/java/com/google/android/horologist/composables/SegmentedProgressIndicator.kt
```diff
@@ -85,76 +82,64 @@ public fun SegmentedProgressIndicator(
...
-            trackSegments.sumOf { it.weight.toDouble() }.toFloat()
-        }
-        // The degrees of the circle that are available for progress indication, once any padding
-        // between segments is accounted for. This will be shared by weighting across the segments.
-        val segmentableAngle = (endAngle - startAngle) - trackSegments.size * paddingAngle
-
+        val endDelta =
+            (asin(
+                strokeWidth.toPx()
+                    / (size.minDimension - strokeWidth.toPx())
```
yschimke, line 110: Nice, this is much better.

From 

```
    var boxMinDimension by remember { mutableStateOf(0) }
    Box(
        contentAlignment = Alignment.Center,
        modifier = Modifier.onSizeChanged { boxMinDimension = min(it.width, it.height) }
    ) {
        val boxMinDimensionDp = with(LocalDensity.current) {
            boxMinDimension.toDp()
        }
```

To BoxWithConstraints

To Canvas & DrawScope.size.

yschimke (approved):

PR #547, opened 2022-09-01
Updates SegmentedProgressIndicator to allow for in-progress background tracks

garanj:
ProgressIndicatorSegment now allows the developer to specify [inProgressTrackColor], which will be used as the background track color for the given segment if it is the one currently in progress.

#### WHAT

Adds optional inProgressTrackColor parameter to ProgressIndicatorSegment

#### WHY

Allows developer to more clearly show the in-progress track

#### Checklist :clipboard:
- [n/a ] Add explicit visibility modifier and explicit return types for public declarations
- [x ] Run spotless check
- [ x] Run tests
- [x ] Update metalava's signature text files


yschimke:
Can you add it to the screenshot test?

yschimke:
LGTM

The only argument that I can think of is that I'd imagine in a typical app, the progress as a percent follows on from the progress of individual items.  So I expect it's easier to change the color of the segment you are currently in directly via the existing ProgressIndicatorSegment.trackColor.

Should indicatorColor have an inprogress color also?

garanj:
Agree, can see a use for having indicatorColor be different whilst in progress too.

garanj:
> Can you add it to the screenshot test?

Added

yschimke (approved):

PR #490, opened 2022-08-15
Add failed dialog to UampPlaylistsScreen

luizgrp:
#### WHAT

Add failed dialog to `UampPlaylistsScreen`.

Screenshot_20220815_104238

#### WHY

In order to have design closer to Figma specs.
In order to have user journey aligned with Figma specs.

#### HOW

- Change ViewModel to emit a "failed" state for any exception;
- Change screen to display a dialog on Failed state;
- Add `onErrorDialogCancelClick` param to screen so it can navigate back when the dialog is dismissed;

#### Checklist :clipboard:
- [N/A] Add explicit visibility modifier and explicit return types for public declarations
- [x] Run spotless check
- [x] Run tests
- [N/A] Update metalava's signature text files

PR #551, opened 2022-09-02
Display failed dialog if no playlists are available

luizgrp:
#### WHAT

Display failed dialog if no playlists are available.

Same dialog as implemented in #490 - no retry functionality implemented yet.

#### WHY

In order to align with Figma specs.

#### HOW

Emit failed state if data returned from repository is empty.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Nice and simple

PR #552, opened 2022-09-02
Change BrowseScreen to use SectionedList

luizgrp:
#### WHAT

Change `BrowseScreen` to use `SectionedList`.

Preview after changes:

Screen Shot 2022-09-02 at 16 53 19

#### WHY

In order to reuse functionality provided by SectionedList.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/browse/BrowseScreen.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/browse/BrowseScreen.kt
@@ -16,7 +16,6 @@
 
 package com.google.android.horologist.media.ui.screens.browse
 
-import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Downloading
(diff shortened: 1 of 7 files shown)


media-ui/src/main/java/com/google/android/horologist/media/ui/screens/browse/BrowseScreen.kt
```diff
@@ -63,90 +63,102 @@ public fun BrowseScreen(
...
-        item {
-            StandardChip(
-                label = stringResource(id = R.string.horologist_browse_library_playlists),
-                icon = Icons.Default.PlaylistPlay,
-                onClick = onPlaylistsClick,
-                chipType = StandardChipType.Secondary
-            )
-        }
+            loaded {
+                Column {
```
yschimke, line 145: This doesn't seem ideal.  They will scale together instead of naturally.  Is this intentional?

yschimke (approved):
One question about the Column with two items.

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/browse/BrowseScreen.kt
luizgrp: I was tempted to lean towards simplicity given there are only two Chips in this section, and in practice this content could have been something as big or bigger than the height of two Chips, so in the end it wouldn't matter.

One side-effect can be seen on the a11y test image diff, which now displays the talk balk of the second Chip - curious to know if I add an item with height bigger than the screen size in a ScalingLazyColumn if it would have the same effect, something to try at some point.

I will make the changes on the BrowseScreen side, but will raise a ticket to evolve SectionedList to expose the `item` scope of ScalingLazyColumn in order to allow it to be used for these scenarios.

luizgrp: re side-effect, for future reference, there were some padding changes, that might have made the second chip "inside" the screen and listed in the talkback

Issue #418, opened 2022-07-20
Play screen in wrong state after going back from Audio output screen


luizgrp:
Player screen in media sample goes to wrong state after going back (swipe back without selecting any output) from Audio output screen. 

Video:

https://user-images.githubusercontent.com/878134/179996574-cb9b4ec3-bca9-4c07-a21d-b1de7c09428e.mp4

Device: Galaxy Watch4 Classic (4W2J) - SM-R895F

#### Steps to reproduce

1) Make sure the watch does not have any bluetooth device connected 
2) Launch app - player screen is displayed with a media
3) Push play - bluetooth settings screen is displayed
4) Swipe back

#### Expected result

Player screen is displayed in an idle/paused state.

#### Actual result

Player screen is displayed in a play state with progress going forward 2 times and looping back.

yschimke:
Should be fixed in https://github.com/google/horologist/pull/424

yschimke:
I'll take another look, this was the first thing I hit, but I haven't tested on device yet. Only emulator.

yschimke:
I was able to reproduce this 1 time out of 3.  But it actually looks like the player is playing.  The state isn't wrong.  The ongoing notification is there.  Sound is just going to the wrong place.

yschimke:
I'll add some more debug for this, make it easier to debug without a special release build.

luizgrp:
Still happening with latest build. 

#### Scenario:
1) song was playing;
2) headphones disconnected for whatever reason;
3) tried to reconnect to headphones and it failed;

![Screenshot_20220906_110921_settings](https://user-images.githubusercontent.com/878134/188610520-ad8769b6-0d41-441c-aedc-f0e93b6e408e.png)

4) swiped back;

##### Expected behaviour 

Player screen is displayed in an idle/paused state.

#### Actual behaviour 

Player screen is displayed in a play state with progress going forward and looping back.

yschimke:
Disabled test to land, struggling with test coroutine dispatchers.

PR #558, opened 2022-09-06
Fix playback state after two unsuccessful BT attempts

yschimke:
#### WHAT

Emit a onPlayWhenReadyChanged when we launch the BT prompts instead of playing. 

#### WHY

MediaController optimistically updates playWhenReadyState, so we need something to trigger MediaSession to flush an update.

#fixes https://github.com/google/horologist/issues/418

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):

yschimke:
Raised https://github.com/androidx/media/issues/163

Issue #553, opened 2022-09-05
Rotary scrolling not working in Entity screen


luizgrp:
Rotary scrolling via bezel or crown not working in Entity screen.

![Screenshot_20220905_122245](https://user-images.githubusercontent.com/878134/188438444-054438bf-2f9e-499a-adbe-d0a78f6ea587.png)

#### Steps to reproduce
1. Launch media sample app;
2. Navigate to Playlists screen;
3. Choose a playlist with many songs (e.g. Electronic);
4. Scroll via bezel or crown;

##### Expected behaviour

Screen should scroll up or down accordingly.

##### Current behaviour

Screen does not scroll up or down.

##### Recovery steps

1. With the Entity screen on display, place the app on background;
2. Launch the app again. It should display the Entity screen;
3. Scroll via bezel or crown - now it works;

---

PS: rotary scrolling can be [tested using an emulator](https://developer.android.com/training/wearables/user-input/rotary-input#emulator).

PR #559, opened 2022-09-07
Re-request focus after Entity Screen state change

yschimke:
#### WHAT

Re-request focus when the state of the EntityScreen changes.

#### WHY

#fixes https://github.com/google/horologist/issues/553

When the whole ScalingLazyList is replaced, the focus is lost.

#### HOW

What for state changes and if we are the active screen then requestFocus()

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
> When the whole ScalingLazyList is replaced, the focus is lost.

I wonder if the screen can be improved then to keep the SLL, something for another PR.

Thanks for fixing it!

yschimke:
I wonder whether this re-implementation of Focus Control by Sergio might remove this hassle.

PR #560, opened 2022-09-07
Add screen samples using SectionedList

luizgrp:
#### WHAT

Add screen samples using `SectionedList`.

https://user-images.githubusercontent.com/878134/188883704-f13f2c22-77d2-46b9-a378-6af5aaae2e15.mp4

#### WHY

In order to have different samples of how the component can be used.

#### HOW

- Add "SectionedList Samples" menu item to `sample` app with two sub-items:
  - Stateless sections
  - Stateful sections

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/sample/src/main/java/com/google/android/horologist/sectionedlist/SectionedListMenuScreen.kt
+++ b/sample/src/main/java/com/google/android/horologist/sectionedlist/SectionedListMenuScreen.kt
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.sectionedlist
+
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.res.stringResource
+import androidx.wear.compose.material.Chip
+import androidx.wear.compose.material.ChipDefaults
+import androidx.wear.compose.material.ScalingLazyListState
+import androidx.wear.compose.material.Text
+import androidx.wear.compose.material.rememberScalingLazyListState
+import com.google.android.horologist.composables.SectionedList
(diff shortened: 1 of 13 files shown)


sample/src/main/java/com/google/android/horologist/sectionedlist/SectionedListMenuScreen.kt
```diff
@@ -0,0 +1,75 @@
...
+import com.google.android.horologist.sample.R
+import com.google.android.horologist.sample.Screen
+import com.google.android.horologist.sectionedlist.component.Title
+
+@Composable
+fun SectionedListMenuScreen(
+    modifier: Modifier = Modifier,
+    navigateToRoute: (String) -> Unit,
+    scalingLazyListState: ScalingLazyListState = rememberScalingLazyListState(),
+    focusRequester: FocusRequester = remember { FocusRequester() }
```
yschimke, line 40: I don't think the focusRequester adds much here if nothing is managing calling it.  Can also do it as a LaunchedEffect in this composable if you know you want navigate to another destination from here.

yschimke (approved):

sample/src/main/java/com/google/android/horologist/sectionedlist/SectionedListMenuScreen.kt
luizgrp: Not sure I follow the comment, it might be due to my lack of knowledge about the usage of FocusRequester.
It's being passed to SectionedList [here](https://github.com/google/horologist/pull/560/files#diff-8e4163ab4d2612ea3c70be9f3c41a367b5970c04d034f426fda770c8ef5e24a6R43) and SectionedList passes it to ScalingLazyColumn [here](https://github.com/google/horologist/blob/ace1334fc6dc268c5bce5c0dd2da4ad6319b4d0c/composables/src/main/java/com/google/android/horologist/composables/SectionedList.kt#L48).

Similar pattern as in [FadeAwayScreenScalingLazyColumn](https://github.com/google/horologist/blob/ace1334fc6dc268c5bce5c0dd2da4ad6319b4d0c/sample/src/main/java/com/google/android/horologist/sample/FadeAwayScreen.kt#L83) sample, but without the SectionedList as man in the middle. Would that be the problem? Or both cases need improvement?

yschimke: There are two parts to FocusRequester in order to enable RSB scrolling of lists.

1) Setting it via a modifier to something that is focusable.
2) Focusing it via focusRequester.requestFocus()

I couldn't see where 2 is happening.

yschimke: Yep, looks like same issue with FadeAwayScreenScalingLazyColumn, I guess I've never tried to scroll that one.

luizgrp: Thanks for the explanation, I see that FadeAwayScreenScalingLazyColumn has the call to requestFocus [here](https://github.com/google/horologist/blob/ace1334fc6dc268c5bce5c0dd2da4ad6319b4d0c/sample/src/main/java/com/google/android/horologist/sample/FadeAwayScreen.kt#L105). Done the same for the screens added here 👍

PR #570, opened 2022-09-12
Improve DSL for SectionedList

luizgrp:
#### WHAT

Improve DSL for `SectionedList`.

#### WHY

- In order to not allow functions to be called from the incorrect scope without using explicit receiver;
- In order to allow IDE to highlight syntax with different colour;

Screen Shot 2022-09-12 at 15 34 18

#### HOW

- Rename `SectionContentScope` to `SectionScope`;
- Add empty `SectionContentScope` and scope content functions in `SectionScope` with it;
- Add `SectionScopeMarker` annotation, annotated with `DslMarker`;
- Annotate `SectionedListScope`, `SectionScope` and `SectionContentScope` with `SectionScopeMarker`;
- Rename `scope` param in `SectionedList` constructor to `content`;
- Rename `builder` params in `SectionedListScope` functions to `content`;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/composables/src/main/java/com/google/android/horologist/composables/SectionedList.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/SectionedList.kt
@@ -40,7 +40,7 @@ public fun SectionedList(
     scalingLazyListState: ScalingLazyListState,
     modifier: Modifier = Modifier,
     scalingParams: ScalingParams = ScalingLazyColumnDefaults.scalingParams(),
-    scope: SectionedListScope.() -> Unit
+    content: SectionedListScope.() -> Unit
 ) {
     ScalingLazyColumn(
         modifier = modifier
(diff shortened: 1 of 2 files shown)


composables/src/main/java/com/google/android/horologist/composables/SectionedList.kt
```diff
@@ -124,28 +124,36 @@ public data class Section<T> internal constructor(
 }
 
 /**
- * Receiver scope which is used by [SectionedList].
+ * DSL marker used to distinguish between scopes of [SectionedList].
+ */
+@DslMarker
```
yschimke, line 129: TIL: https://kotlinlang.org/docs/type-safe-builders.html#scope-control-dslmarker

yschimke (approved):

PR #576, opened 2022-09-13
Add snapshot tests for SectionedList

luizgrp:
#### WHAT

Add snapshot tests for SectionedList.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/composables/src/test/kotlin/com/google/android/horologist/composables/SectionedListTest.kt
+++ b/composables/src/test/kotlin/com/google/android/horologist/composables/SectionedListTest.kt
@@ -1,0 +1,412 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


composables/src/test/kotlin/com/google/android/horologist/composables/SectionedListTest.kt
```diff
@@ -0,0 +1,412 @@
...
+
+    @Composable
+    private fun SectionedListPreview(
+        scrollState: ScalingLazyListState,
+        content: @Composable () -> Unit
+    ) {
+        RoundPreview(round = true) {
+            Scaffold(
+                modifier = Modifier.fillMaxSize(),
+                timeText = {
```
yschimke, line 194: Showing the time text with content overlapping seems curious.  Is it useful here?

yschimke (approved):

composables/src/test/kotlin/com/google/android/horologist/composables/SectionedListTest.kt
luizgrp: good point, removed 👍

PR #588, opened 2022-09-16
Media Playback Benchmark

yschimke:
#### WHAT

Add a sample playback macrobenchmark.

See https://ui.perfetto.dev/#!/?s=92d3faa2759e6f5eaf1fc94d93c9b4a45f94c4cd0fe4c98ad8d9dd02724599b

#### WHY

Allow visibility into app activity via profiler with custom tracing to investigate issues.

#### HOW

Extracted from https://github.com/google/horologist/pull/585

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-sample-benchmark/src/main/java/com/google/android/horologist/mediasample/benchmark/MediaControllerHelper.kt
+++ b/media-sample-benchmark/src/main/java/com/google/android/horologist/mediasample/benchmark/MediaControllerHelper.kt
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.mediasample.benchmark
+
+import android.content.ComponentName
+import android.os.Looper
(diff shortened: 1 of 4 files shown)


media-sample-benchmark/src/main/java/com/google/android/horologist/mediasample/benchmark/MediaControllerHelper.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.media3.session.MediaController
+import androidx.media3.session.SessionToken
+import androidx.test.platform.app.InstrumentationRegistry
+import com.google.common.util.concurrent.ListenableFuture
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.withContext
+
+object MediaControllerHelper {
+    fun lookupController(component: ComponentName): ListenableFuture<MediaBrowser> {
```
wrkben, line 32: given the style of other parts of the project, why the sudden introduction of listenablefuture instead of suspend functions?

yschimke: Because I'm not inserting it into code that is naturally suspend functions. It's in the setup block of the macrobenchmark test, and the await is an additional stop on the ListenableFuture in all cases, so I'm not introducing it, just not hiding it this time.

I think if I want to put it in a suspend form, I need to add a `runBlocking(Dispatchers.Main) {}` to the setup block as well.

yschimke: I might kick off a discussion with the performance team.

wrkben (approved):

media-sample-benchmark/src/main/java/com/google/android/horologist/mediasample/benchmark/MediaControllerHelper.kt
yschimke: I'll follow up.

PR #592, opened 2022-09-19
Improve BrowseScreen to allow more customisation

luizgrp:
#### WHAT

Improve `BrowseScreen` to allow more customisation.

#### WHY

So `BrowseScreen` can be used without having to also use `media-ui`'s model `PlaylistDownloadUiModel`.

#### HOW

- Add `PlaylistDownloadBrowseScreen` with previous code from `BrowseScreen`;
- Add new implementation for `BrowseScreen`;
- Add previews;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
LGTM

I'll try this out in a sample app, it's hard for me to judge where the balance is currently.

PR #594, opened 2022-09-20
Fix data race in DataClient.dataItemFlow

### LINKED ISSUE #593: Data race in DataClient.dataItemFlow(…)
Hello,

As you can see in the snippet below, the code collecting a flow returned by `DataClient.dataItemFlow(…)` can end up on an old value and never receiving the new one.

This would happen if a new value gets generated while the `this@dataItemFlow.getDataItem(uri).await()` call is running, or after, before the `addListener(…)` call completes.

https://github.com/google/horologist/blob/6d75e70311410aa392b509987d9d0d893b7119ab/datalayer/src/main/java/com/google/android/horologist/data/store/impl/DataItemFlow.kt#L53-L57

I can send a PR, should I?

Have a great day!

LouisCAD:
#### WHAT

Fixes #593

#### WHY

Because data races are bad.

#### HOW

By ensuring we are subscribed to changes before getting the current value.

#### Checklist :clipboard:
- [ ] ~~Add explicit visibility modifier and explicit return types for public declarations~~
- [ ] ~~Run spotless check~~
- [ ] ~~Run tests~~
- [ ] ~~Update metalava's signature text files~~

### DIFF
--- a/datalayer/src/main/java/com/google/android/horologist/data/store/impl/DataItemFlow.kt
+++ b/datalayer/src/main/java/com/google/android/horologist/data/store/impl/DataItemFlow.kt
@@ -50,11 +50,11 @@ public fun <T> DataClient.dataItemFlow(
         .authority(nodeId)
         .build()
 
-    val item: DataItem? = this@dataItemFlow.getDataItem(uri).await()
+    addListener(listener, uri, DataClient.FILTER_LITERAL).await() // Ensure we are subscribed to updates first,
+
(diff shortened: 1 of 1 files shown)


LouisCAD:
A CI job was cancelled, I don't know why 🤔

yschimke:
> A CI job was cancelled, I don't know why 🤔

Timeout, github CI is flaky.

yschimke:
Thanks!

PR #597, opened 2022-09-20
Make KnownDuration a data class

yschimke:
#### WHAT

Make KnownDuration a data class

#### WHY

Want equality for StateFlow updates.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/media/src/main/java/com/google/android/horologist/media/model/MediaPosition.kt
+++ b/media/src/main/java/com/google/android/horologist/media/model/MediaPosition.kt
@@ -27,7 +27,7 @@ import kotlin.time.Duration
 public sealed class MediaPosition(
     public open val current: Duration
 ) {
-    public class KnownDuration internal constructor(
+    public data class KnownDuration internal constructor(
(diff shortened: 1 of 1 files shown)


luizgrp (approved):
there was a reason it wasn't added as data class, maybe it only made sense in the previous media toolkit project, can't remember now

yschimke:
That sounds familiar, but without this the position keeps changing evey second even when paused.

luizgrp:
for future reference: https://github.com/google/horologist/blob/bdb1f95100c1aa6ac747674fc397072e24efa074/media-data/src/test/java/com/google/android/horologist/media/data/repository/PlayerRepositoryImplTest.kt#L948

yschimke:
data classes can be bad for public APIs, but if I was holding a hard line about that, I think I've softened now.

https://jakewharton.com/public-api-challenges-in-kotlin/

yschimke:
Let's revert if we remember a reason we still hold.

PR #600, opened 2022-09-20
Fix TimeText behavior in ScalingLazyList according to scrollState

moncefguettat:
#### WHAT

I tried to customize the `ScalingLazyColumn` offset but the `TimeText` disappeared and was scrolled away. 

the expected behavior is that the time text remain and consider the added offset.

#### WHY

Changing the `initialCenterItemScrollOffset` in `ScalingLazyListState` and also the `itemOffset` in the `AutoCenteringParams` does not reflect in the `fadeAway` behavior of the TimeText. 

Currently, we are only handling the `itemIndex` customization and that's why if we try to move the centered item with the offset will make the TimeText scrolled away.

#### HOW

I added a second property in the [NavScaffoldViewModel.kt](https://github.com/moncefguettat/horologist/blob/main/compose-layout/src/main/java/com/google/android/horologist/compose/navscaffold/NavScaffoldViewModel.kt) to handle to the offset changes, which is used by the `TimeText` `fadeAway` behavior to react to any changes to the centered Item `Index` or `Offset`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/compose-layout/src/androidTest/java/com/google/android/horologist/compose/navscaffold/NavScaffoldTest.kt
+++ b/compose-layout/src/androidTest/java/com/google/android/horologist/compose/navscaffold/NavScaffoldTest.kt
@@ -70,6 +70,7 @@ class NavScaffoldTest {
 
         lateinit var aScrollState: ScalingLazyListState
         lateinit var bScrollState: ScalingLazyListState
+        lateinit var cScrollState: ScalingLazyListState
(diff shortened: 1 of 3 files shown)


yschimke (approved):
Thanks, looks good!

yschimke:
CLA accepted slightly after the PR landed.

PR #605, opened 2022-09-21
Add stateful version of PodcastControlButtons

luizgrp:
#### WHAT

Add stateful version of `PodcastControlButtons`.

#### WHY

For convenience, so `PlayerViewModel` and `PlayerUiState` can be passed as params.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/components/PodcastControlButtons.kt
```diff
@@ -24,6 +24,46 @@ import com.google.android.horologist.media.ui.components.controls.MediaButtonDef
...
+/**
+ * Stateful version of [PodcastControlButtons].
+ *
+ * This version listens to [PlayerUiState]s emitted from [PlayerViewModel] to update the controls.
+ */
+@ExperimentalHorologistMediaUiApi
+@Composable
+public fun PodcastControlButtons(
+    playerViewModel: PlayerViewModel,
+    playerUiState: PlayerUiState,
```
yschimke, line 39: Doesn't playerViewModel provide playerUiState? Is it an optimisation to provide both?

luizgrp: Good point, didn't put much thought on this now - adding this composable to be used in the docs as example so I just followed the same pattern for [DefaultPlayerScreenControlButtons](https://github.com/google/horologist/blob/7b492a41c12a9a65448d5e722b5b0c78b7c30cf8/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/player/PlayerScreen.kt#L123-L125). When adding that one, I think the idea was to have clear separation:
- the component will render its state based on `PlayerUiState`;
- the component will hook the controls with functions from `PlayerViewModel`;

It gives flexibility to pass a `PlayerUiState` that wasn't emitted by `PlayerViewModel`. Not sure if this makes much sense, happy to change both in another PR.

PR #606, opened 2022-09-21
Fallback to title and albumArtist when displayTitle and artist aren't set

fstanis:


luizgrp (approved):
Nice! just wondering if tests will have to be adapted in MediaMapperTest?

PR #608, opened 2022-09-22
Upgrade Android Gradle plugin to 7.3.0

luizgrp:
#### WHAT

Upgrade Android Gradle plugin to 7.3.0.

#### HOW

Applied all suggested changes by Android Studio.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
The failure is a bug in lint https://b.corp.google.com/issues/196420849, so feel free to add suppression.

yschimke (approved):

media-sync/src/androidTest/java/com/google/android/horologist/media/sync/workers/TestSyncModule.kt
```diff
@@ -1,67 +0,0 @@
...
-import dagger.Provides
-import dagger.hilt.InstallIn
-import dagger.hilt.components.SingletonComponent
-import kotlinx.coroutines.CoroutineDispatcher
-import kotlinx.coroutines.Dispatchers
-import javax.inject.Singleton
-
-@Module
-@InstallIn(SingletonComponent::class)
-object TestSyncModule {
```
luizgrp, line 33: This module contains only one test, and ignoring it was still making the test suite to crash in CI (runs fine locally). So it makes me think that there is something with Hilt that is making the tests crash after upgrading AGP. Will raise a separate issue to investigate it.

audio/build.gradle
```diff
@@ -78,6 +73,12 @@ android {
             res.srcDirs += 'src/sharedTest/res'
         }
     }
+    lint {
+        checkReleaseBuilds false
+        textOutput file('stdout')
```
yschimke, line 78: I know this is mainly a move, but I have a lot of stdout files for some reason in my workspace.  Intended?

luizgrp: not sure, I also wanted to delete it as it seems unnecessary, the xml output is already being generated in the build output folder

but decided to do in another PR to check why this was being generated in first place

maybe you should tell us, it was added in the very first [PR](https://github.com/google/horologist/pull/1/files#diff-57dd3f986fa1bf0c6e02f39a90d9ae9499d7f93880f6fe4665baa83ffdda4ca3R57) in this repo and is being copied over and over :D

yschimke: Probably from Accompanist.  But I didn't have these files previously.

luizgrp: yeah I suspected that, searched there and couldn't find it.. will remove them 👍

PR #610, opened 2022-09-22
Add twitter compose rules ktlint check

luizgrp:
#### WHAT

Run [Twitter's Jetpack Compose Rules](https://github.com/twitter/compose-rules) in CI.

#### WHY

In order to improve quality of project's composables.

#### HOW

Add plugin via [kolinter](https://twitter.github.io/compose-rules/ktlint/#using-with-kotlinter) given it's not currently working with [spotless](https://twitter.github.io/compose-rules/ktlint/#using-with-spotless). Given it might be potentially  doing duplicated work and increasing build time, we might decide to only add it to the project once it's supported by spotless.

#### RESULTS

- A lot of findings in composables functions in preview files, test files, and sample apps, which we should filter out for those rules.

- `modifier-missing-check` in many functions.
```
Lint error > [twitter-compose:modifier-missing-check] This @Composable function emits content but doesn't have a modifier parameter.

See https://twitter.github.io/compose-rules/rules/#when-should-i-expose-modifier-parameters for more information.
```

- `param-order-check` in `WearNavScaffold`, `SegmentedProgressIndicator`, `FillerScreen`, `BigScalingLazyColumn`, `BigColumn`, `MediaPlayerScaffold`, `StandardChip` and other internal functions.
```
Lint error > [twitter-compose:param-order-check]         Parameters in a composable function should be ordered following this pattern: params without defaults, modifiers, params with defaults and optionally, a trailing function that might not have a default param.

See https://twitter.github.io/compose-rules/rules/#ordering-composable-parameters-properly for more information.
```

- `modifier-reused-check` in `PlayPauseButton` (seems to be intended?).
```
[twitter-compose:modifier-reused-check] Modifiers should only be used once and by the root level layout of a Composable. This is true even if
appended to or with other modifiers e.g. 'modifier.fillMaxWith()'.

Use Modifier (with a capital 'M') to construct a new Modifier that you can pass to other Composables.

See https://twitter.github.io/compose-rules/rules/#dont-re-use-modifiers for more information.
```

- `twitter-compose:multiple-emitters-check` in `TimePicker`.
```
Lint error > [twitter-compose:multiple-emitters-check] Composable functions should only be emitting content into the composition from one source at their top level.
```

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,7 +56,8 @@ jobs:
         run: |
           ./gradlew --scan --stacktrace \
               spotlessCheck \
-              lintDebug
+              lintDebug \
+              lintKotlin
 
       - name: Unit Tests
         run: |
--- a/build.gradle
+++ b/build.gradle
@@ -36,6 +36,8 @@ buildscript {
         classpath libs.paparazziPlugin
 
         classpath libs.dagger.hiltandroidplugin
+
+        classpath libs.twitter.compose.rules.ktlint
     }
 }
 
@@ -47,6 +49,7 @@ plugins {
     alias(libs.plugins.ksp) apply false
     alias(libs.plugins.kapt) apply false
     alias(libs.plugins.protobuf) apply false
+    alias(libs.plugins.kolinter) apply false
 }
 
 apply plugin: 'org.jetbrains.dokka'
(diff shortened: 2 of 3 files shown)


yschimke (approved):
Nice idea!

luizgrp:
Closing as kotlinter does not support baselines - https://github.com/jeremymailen/kotlinter-gradle/issues/167

Using kotlinter was a temporary option until spotless can enable custom ruleset https://github.com/diffplug/spotless/issues/1220

PR #612, opened 2022-09-22
Decouple PlayerViewModel

fstanis:
#### WHAT
Separates most `PlayerViewModel` logic into a plain class.

#### WHY
This makes it easier to implement a `PlayerViewModel` without having a subclass a `ViewModel` for niche use-cases, e.g. multiple players in a single `ViewModel`. Also makes testing slightly easier.

#### HOW
Moves most logic that doesn't require a lifecycle from `PlayerViewModel` into `PlayerUiStateProducer` and adds a `PlayerUiController` interface (which `PlayerViewModel` implements) so that composables like `DefaultPlayerScreenControlButtons` wouldn't need to receive a `PlayerViewModel` (but still can).

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/components/PodcastControlButtons.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/components/PodcastControlButtons.kt
@@ -24,18 +24,18 @@ import com.google.android.horologist.media.ui.components.controls.MediaButtonDef
 import com.google.android.horologist.media.ui.components.controls.SeekBackButton
 import com.google.android.horologist.media.ui.components.controls.SeekButtonIncrement
 import com.google.android.horologist.media.ui.components.controls.SeekForwardButton
+import com.google.android.horologist.media.ui.state.PlayerUiController
 import com.google.android.horologist.media.ui.state.PlayerUiState
-import com.google.android.horologist.media.ui.state.PlayerViewModel
 
 /**
- * Stateful version of [PodcastControlButtons].
+ * Convenience wrapper of [PodcastControlButtons].
  *
- * This version listens to [PlayerUiState]s emitted from [PlayerViewModel] to update the controls.
+ * This version passes events to the provided [PlayerUiController].
  */
 @ExperimentalHorologistMediaUiApi
 @Composable
 public fun PodcastControlButtons(
-    playerViewModel: PlayerViewModel,
+    playerController: PlayerUiController,
     playerUiState: PlayerUiState,
     modifier: Modifier = Modifier,
     showProgress: Boolean = true,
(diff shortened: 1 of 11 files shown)


media-ui/src/main/java/com/google/android/horologist/media/ui/components/PodcastControlButtons.kt
```diff
@@ -24,18 +24,18 @@ import com.google.android.horologist.media.ui.components.controls.MediaButtonDef
 import com.google.android.horologist.media.ui.components.controls.SeekBackButton
 import com.google.android.horologist.media.ui.components.controls.SeekButtonIncrement
 import com.google.android.horologist.media.ui.components.controls.SeekForwardButton
+import com.google.android.horologist.media.ui.state.PlayerUiController
 import com.google.android.horologist.media.ui.state.PlayerUiState
-import com.google.android.horologist.media.ui.state.PlayerViewModel
 
 /**
- * Stateful version of [PodcastControlButtons].
+ * Convenience wrapper of [PodcastControlButtons].
```
yschimke, line 31: For Composables, Stateful has specific meaning which is useful here.  But given state is passed in, this seems correct.

media-ui/src/main/java/com/google/android/horologist/media/ui/state/PlayerUiController.kt
```diff
@@ -0,0 +1,27 @@
...
+
+package com.google.android.horologist.media.ui.state
+
+/**
+ * Represents controls commonly present in a player's UI.
+ */
+public interface PlayerUiController {
+    public fun play()
+    public fun pause()
+    public fun skipToPreviousMedia()
```
yschimke, line 25: Can you add seekForward, seekBack, since we have them in other controls?

Can be follow up.

media-ui/src/main/java/com/google/android/horologist/media/ui/state/PlayerViewModel.kt
```diff
@@ -23,80 +23,40 @@ import androidx.lifecycle.viewModelScope
...
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.StateFlow
-import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.stateIn
 
 @ExperimentalHorologistMediaUiApi
 public open class PlayerViewModel(
     private val playerRepository: PlayerRepository
-) : ViewModel() {
+) : ViewModel(), PlayerUiController {
```
yschimke, line 35: Should it have a PlayerUiController, rather than be a PlayerUiController?

#tioli

yschimke (approved):
LGTM - but I'd love @luizgrp take.

luizgrp (approved):
nice one! I think this also address what we were discussing [here](https://github.com/google/horologist/pull/605#discussion_r976670794)

media-ui/src/main/java/com/google/android/horologist/media/ui/state/PlayerUiController.kt
fstanis: Done.

media-ui/src/main/java/com/google/android/horologist/media/ui/state/PlayerViewModel.kt
fstanis: Done - it's also now a class rather than interface.

media-ui/src/main/java/com/google/android/horologist/media/ui/state/PlayerUiStateProducer.kt
```diff
@@ -0,0 +1,73 @@
...
+
+import com.google.android.horologist.media.ExperimentalHorologistMediaApi
+import com.google.android.horologist.media.repository.PlayerRepository
+import com.google.android.horologist.media.ui.ExperimentalHorologistMediaUiApi
+import com.google.android.horologist.media.ui.components.controls.SeekButtonIncrement
+import com.google.android.horologist.media.ui.state.mapper.PlayerUiStateMapper
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.combine
+
+@ExperimentalHorologistMediaUiApi
```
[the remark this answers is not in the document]
fstanis: Done on both new classes.

PR #618, opened 2022-09-23
Introduce HighBandwidthRequest.Type

yschimke:
#### WHAT

Introduce HighBandwidthRequest.Type

#### WHY

Future changes benefit from not needing to have 3 fields, and instead have a map.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
@wrkben I've broken up my PR, but trying to land.  So will request post review.  to make forward progress.

PR #619, opened 2022-09-23
Change pinned networks to a Set.

yschimke:
#### WHAT

Change pinned networks to a Set.

#### WHY

Ahead of a change to return Flow<Set<NetworkType>>, make the change to UI classes separately.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
@wrkben I've broken up my PR, but trying to land. So will request post review. to make forward progress.

PR #624, opened 2022-09-26
Use a smarter timeout for requesting High Bandwidth Network

yschimke:
#### WHAT

Remember the lease time in NetworkLease, and change await to be 3 seconds from the lease time regardless of this current request time.

#### WHY

Scrolling through a list of images when set to request high bandwidth network, that is unavailable causes a 3 second delay before each image.  Instead the first request should assume that delay, and subsequent should delay 3 seconds from when that initial lease was created. 

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
@wrkben I've broken up my PR, but trying to land. So will request post review. to make forward progress.

wrkben (approved):

PR #631, opened 2022-09-26
Improve Section.State to require a type parameter

luizgrp:
#### WHAT

Improve `Section.State` to require a type parameter.

#### WHY

In order to prevent at compile time a `State` with an invalid type to be passed to a `Section`, which would cause a cast exception at runtime. 

For example, in the samples below with the current code where the `Section` has type `String` and `State` has type `Int`:

Screen Shot 2022-09-26 at 14 26 26

Screen Shot 2022-09-26 at 15 03 55

Same code after these changes showing issues at compile time:

Screen Shot 2022-09-26 at 14 25 41

Screen Shot 2022-09-26 at 15 02 21

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/composables/src/main/java/com/google/android/horologist/composables/SectionedList.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/SectionedList.kt
@@ -89,24 +89,26 @@ internal fun <T> Section<T>.display(scope: ScalingLazyListScope) {
     }
 
     when (section.state) {
-        Section.State.Loading -> {
+        is Section.State.Loading -> {
             section.loadingContent?.let { content ->
                 scope.item { SectionContentScope.content() }
             }
         }
-        is Section.State.Loaded<*> -> {
+
+        is Section.State.Loaded -> {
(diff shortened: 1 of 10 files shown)


composables/src/main/java/com/google/android/horologist/composables/SectionedList.kt
```diff
@@ -89,24 +89,26 @@ internal fun <T> Section<T>.display(scope: ScalingLazyListScope) {
...
             }
         }
-        is Section.State.Loaded<*> -> {
+
+        is Section.State.Loaded -> {
             val list = section.state.list
             scope.items(list.size) { index ->
-                @Suppress("UNCHECKED_CAST")
-                section.loadedContent(SectionContentScope, list[index] as T)
+                section.loadedContent(SectionContentScope, list[index])
```
yschimke, line 101: Nice! the money line.

yschimke (approved):

PR #635, opened 2022-09-27
Add sample screen for expandable section with SectionedList

luizgrp:
#### WHAT

Add sample screen for expandable section with `SectionedList`.

https://user-images.githubusercontent.com/878134/192574067-b73bd6df-0c41-4219-93d5-58cb19c00676.mp4

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/sample/src/main/java/com/google/android/horologist/sectionedlist/expandable/SectionedListExpandableScreen.kt
+++ b/sample/src/main/java/com/google/android/horologist/sectionedlist/expandable/SectionedListExpandableScreen.kt
@@ -1,0 +1,193 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


sample/src/main/java/com/google/android/horologist/sectionedlist/expandable/SectionedListExpandableScreen.kt
```diff
@@ -0,0 +1,193 @@
...
+            loaded { (text, iconTint) ->
+                SingleLineChip(
+                    text = text,
+                    imageVector = Icons.Outlined.Circle,
+                    iconTint = iconTint
+                )
+            }
+        }
+
+        section(state = laterSectionState) {
```
yschimke, line 127: My 2c - TIOLI

I like seeing these sections split out.  IF I'm implementing these I'd want to see my top level list and just a sequence of sections.  And then separate reusable sections.  But may make the sample less illustrative.

luizgrp: agreed, I've done it in the previews [here](https://github.com/google/horologist/blob/a2d9094a74c930bdc2baf94ca2a42e5933736c42/composables/src/debug/java/com/google/android/horologist/composables/SectionedListPreview.kt#L97-L98) and I think it makes the API look more convoluted than it is. 
I will think on a better way of extracting these sections in separate functions and iterate 👍

luizgrp: https://github.com/google/horologist/pull/637

PR #638, opened 2022-09-28
Add Not Connected PlayerUIState

yschimke:
#### WHAT

Add Not Connected PlayerUIState.

#### WHY

Avoid duplicating this in every app later.

#### HOW

Simple constant will connected = false, and null media.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/state/PlayerUiState.kt
```diff
@@ -64,4 +64,25 @@ public data class PlayerUiState(
     val seekBackButtonIncrement: SeekButtonIncrement = SeekButtonIncrement.Unknown,
     val seekForwardButtonIncrement: SeekButtonIncrement = SeekButtonIncrement.Unknown,
     val connected: Boolean
-)
+) {
+    companion object {
+        /**
+         * Value for UIs before a connected player is available.
+         */
+        val NotConnected = PlayerUiState(
```
fstanis, line 72: Can you also update this to use it?

https://github.com/google/horologist/blob/8193c0f5b713f1f5b8fef7e007ce4c89359c875f/media-ui/src/main/java/com/google/android/horologist/media/ui/state/PlayerViewModel.kt#L56-L70

yschimke: Seemed to work just fine, types were always nullable.  Placeholders, then nothing playing, then media.

PR #642, opened 2022-09-29
Improve docs for media-sample module

luizgrp:
#### WHAT

Improve docs for `media-sample` module.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Much better

Issue #363, opened 2022-07-11
[CI] Add check for files added without using Git LFS


luizgrp:
Improve CI script to check and fail when snapshot images are added without using Git LFS as per https://github.com/cashapp/paparazzi#git-lfs

PR #649, opened 2022-09-30
Add check for files added without using Git LFS

luizgrp:
#### WHAT

Add check for files added without using Git LFS

#### WHY

#363

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
Can we print out the images?

```
+ diff /dev/fd/63 /dev/fd/62
++ git ls-files ':(attr:filter=lfs)'
++ sort
++ git lfs ls-files -n
++ sort
Error: Process completed with exit code 1.
```

Or you expect it to be obvious since you are adding?

yschimke (approved):

luizgrp:
> Can we print out the images?

done:

```
+ diff /dev/fd/63 /dev/fd/62
++ git ls-files ':(attr:filter=lfs)'
++ sort
++ git lfs ls-files -n
++ sort
31a32
> media-sample/src/main/res/drawable-mdpi/kyoto.jpg
Error: Process completed with exit code 1.
```

PR #657, opened 2022-10-04
Change default to show browse screen

yschimke:
#### WHAT

Change default to show browse screen.

#### WHY

It wasn't clear if this was a conscious decision to load items, but that's not finding items anyway, so instead show a browse screen.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):

yschimke:
Going to close, it's a setting already.

If we change to load previous items, then we should fallback to navigating to the library and consider removing this setting.

PR #658, opened 2022-10-05
Add heading semantics to Title

luizgrp:
#### WHAT

Add heading semantics to `Title`.

#### WHY

In order to improve [accessibility](https://developer.android.com/jetpack/compose/accessibility#define-headings).

#### HOW

- Add [heading](https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/package-summary#(androidx.compose.ui.semantics.SemanticsPropertyReceiver).heading()) semantics;
- Improve kdoc;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
If you regenerate the A11y tests, A11ySnapshotHandler already detects the header.  Should visually show where it's being used if our A11y tests cover these titles already.

yschimke (approved):
Nice!

media-ui/src/main/java/com/google/android/horologist/media/ui/components/base/SectionTitle.kt
```diff
@@ -20,29 +20,38 @@ import androidx.annotation.StringRes
...
 @Composable
-internal fun Title(
+internal fun SectionTitle(
     text: String,
     modifier: Modifier = Modifier
 ) {
     Text(
         text = text,
-        modifier = modifier,
+        modifier = modifier.semantics { heading() },
```
yschimke, line 54: Is it possible to override this? Not sure I need to , just wondering what the pattern is to change semantics applied by framework components?

luizgrp: my understanding is that [clearAndSetSemantics](https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/package-summary#(androidx.compose.ui.Modifier).clearAndSetSemantics(kotlin.Function1)) can be used for that, and a custom strategy can be passed to that first param 

@ataulm wrote about it [here](https://medium.com/google-developer-experts/grouping-semantics-in-jetpack-compose-ui-93fa47e615db)

also more info in the docs [here](https://developer.android.com/jetpack/compose/semantics#properties)

yschimke: Yep, so you'll need to apply at a higher level.  This `heading()` would apply on top of any set on modifier passed to SectionTitle?

Not a problem, just checking.

luizgrp:
> If you regenerate the A11y tests, A11ySnapshotHandler already detects the header. Should visually show where it's being used if our A11y tests cover these titles already.

thanks! nice to see them there!

yschimke (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/components/base/SectionTitle.kt
luizgrp: added unit tests:
https://github.com/google/horologist/pull/658/files#diff-1839557f463f0f055f3f29198701eedb4f7c56d449212a9df57d50405b648e08R45

media-ui/src/androidTest/java/com/google/android/horologist/media/ui/components/base/TitleTest.kt
```diff
@@ -0,0 +1,58 @@
...
+            .assertExists()
+    }
+
+    @Test
+    fun givenHeadingMofifierIsOverridden_thenHeadingModifierIsNOTPresent() {
+        // given
+        composeTestRule.setContent {
+            Title(
+                text = "text",
+                modifier = Modifier.clearAndSetSemantics {}
```
luizgrp, line 50: My understanding is that this will clear all the semantics, whereas I wonder if there is a way of clearing just one in specific, in this case, the Heading

yschimke: Nice, my reading of the docs, was that it applied below this composable.

I am surprised because this reads as `Modifier.clearAndSetSemantics {}.semantics { heading() }`.

But the test looks good.

PR #661, opened 2022-10-06
Fixes for resources in screenshot tests.

yschimke:
#### WHAT

Fixes for resources in screenshot tests.

#### WHY

Screenshot tests were relying on placeholders which broken, so fix it with a fake ImageLoader only from resources.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/test/java/com/google/android/horologist/media/ui/components/actions/ShowPlaylistChipTest.kt
```diff
@@ -78,18 +75,17 @@ class ShowPlaylistChipTest {
     }
 
     @Test
-    fun givenNOName_thenDoesNOTDisplayArtwork() {
+    fun givenNOName_thenDoesDisplayArtwork() {
```
yschimke, line 78: This seems like the behaviour was always different.  Is there anywhere that does hide the artwork when name is null?

media-ui/src/test/java/com/google/android/horologist/media/ui/components/MediaChipTest.kt
```diff
@@ -43,95 +44,102 @@ class MediaChipTest {
...
-                    placeholder = rememberVectorPainter(image = Icons.Default.Album)
-                )
+            FakeImageLoader.Resources.override {
+                Box(
+                    modifier = Modifier.background(Color.Black),
+                    contentAlignment = Alignment.Center
+                ) {
+                    MediaChip(
+                        title = "Red Hot Chilli Peppers",
+                        artworkUri = R.drawable.ic_uamp,
```
luizgrp, line 54: nice! 

nit: as tests are part of the "documentation" on how to use our components, this might lead users to think that they can pass a drawable resource to `MediaChip.artworkUri`. I think it's a good trade-off, there is that FakeImageLoader.Resources.override there. Maybe `R.drawable.ic_uamp` should be placed in a constant defined in `FakeImageLoader` in order to help clarify this?

compose-tools/src/main/java/com/google/android/horologist/compose/tools/coil/FakeImageLoader.kt
```diff
@@ -71,6 +75,25 @@ public class FakeImageLoader(private val imageFn: suspend (ImageRequest) -> Imag
...
+        public val Resources: FakeImageLoader =
+            FakeImageLoader { request ->
+                val context = request.context
+                val data = request.data as? Int
+                val drawable = data?.let { context.getDrawable(data) }
+
+                if (drawable != null) {
+                    SuccessResult(drawable, request, DataSource.DISK)
+                } else {
+                    ErrorResult(context.getDrawable(R.drawable.ic_uamp), request, FileNotFoundException(request.data.toString()))
```
luizgrp, line 87: nit: was this drawable added only for tests? then I'd suggest to place in the debug folder instead of main

media-ui/src/test/java/com/google/android/horologist/media/ui/components/MediaChipTest.kt
```diff
@@ -43,95 +44,102 @@ class MediaChipTest {
...
-                    placeholder = rememberVectorPainter(
-                        image = Icons.Default.Album,
-                        tintColor = Color.Blue
+            FakeImageLoader.Resources.override {
+                Box(
+                    modifier = Modifier.background(Color.Black),
+                    contentAlignment = Alignment.Center
+                ) {
+                    MediaChip(
+                        media = MediaUiModel(id = "id", artworkUri = "artworkUri"),
```
luizgrp, line 107: this is a confusing scenario, the test is passing an invalid uri and the FakeImageLoader is defaulting to R.drawable.ic_uamp - I think FakeImageLoader should not load anything in this scenario, so you would have spotted this, and fixed this test to:
1) pass R.drawable.ic_uamp to artworkUri
2) remove placeholder

same for test below

media-ui/src/test/java/com/google/android/horologist/media/ui/components/actions/ShowPlaylistChipTest.kt
luizgrp: maybe it was documenting a behaviour of an old code, can't remember!

luizgrp (approved):
nice! great to no have to rely on placeholders anymore! it makes the tests clearer!

compose-tools/src/main/java/com/google/android/horologist/compose/tools/coil/FakeImageLoader.kt
yschimke: I think this is done one level above

```
debugImplementation projects.composeTools
```

PR #664, opened 2022-10-07
Add previews to measure char limit of string resources

luizgrp:
#### WHAT

Add previews to measure char limit of string resources.

#### WHY

In order to help to get translations for them.

#### HOW

- Add previews checking the maximum number of a large character that fit on the component without truncating, with normal font, on a small round device.
- Add `PreviewForCharLimit` annotation to document these previews that were built specifically for that.
- Add CHAR_LIMIT indication in the string descriptions.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
Is it worth committing to the repo? Your call.

yschimke (approved):

luizgrp:
> Is it worth committing to the repo? 

Good point, no, I don't think so. But I will leave them until the translations are done, just in case we need to check them in another font or device spec.

luizgrp:
Closing and just adding the char limits in https://github.com/google/horologist/pull/665

PR #667, opened 2022-10-07
Add a11y tests

luizgrp:
#### WHAT

Add a11y snapshot tests for `MediaArtwork` and `ShowPlaylistChip`. 
Also changed test for `MediaChip` to display image. 

#### WHY

Apart from the benefit of having the tests, I wanted to check if this [line](https://github.com/google/horologist/blob/e51902e114b14a2466a53c3aaf96af2f58e45945/media-ui/src/main/java/com/google/android/horologist/media/ui/components/MediaArtwork.kt#L37) was not causing duplicated feedback on talkback (manual tests did not indicate it).

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Nice!

PR #668, opened 2022-10-10
Paparazzi 1.1.0

yschimke:
#### WHAT

Paparazzi Snapshot

#### WHY

Testing of snapshots and code cleanup.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
Most changes are on the edges, the ordering of applying the Circle crop has changed.

PR #669, opened 2022-10-10
Scroll Away modifier

yschimke:
#### WHAT

Scroll Away modifier from Wear Compose 1.1

#### WHY

fadeAway is a less optimised implementation, the new scrollAway modifier if following approved designs and more performant.

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/compose/compose-material/src/commonMain/kotlin/androidx/wear/compose/material/ScrollAway.kt?q=scrollAway

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScrollAway.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScrollAway.kt
@@ -1,0 +1,146 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


BowerSteve (approved):

compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScrollAway.kt
```diff
@@ -0,0 +1,146 @@
...
+                    scaleY = motionFraction
+                    transformOrigin = TransformOrigin(pivotFractionX = 0.5f, pivotFractionY = 0.0f)
+                }
+            }
+        }
+    }
+)
+
+// Trivial modifier that neither measures nor places the content.
+private fun Modifier.ignore(): Modifier = this.then(
```
fstanis, line 122: Maybe I'm missing something, but is this needed? Couldn't it be no-op or return an empty `Modifier`?

If not, can we rename this to something like `emptyMeasure()` instead of `ignore()`?

yschimke: We should upstream that fix if so.

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/compose/compose-material/src/commonMain/kotlin/androidx/wear/compose/material/ScrollAway.kt;l=121?q=Modifier.ignore

Will land as is.

PR #671, opened 2022-10-10
Add ranged value for complications

yschimke:
#### WHAT

Add ranged value for complications

#### WHY

A a succinct API for Ranged Values.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/tiles/src/main/java/com/google/android/horologist/tiles/complication/ComplicationTemplate.kt
+++ b/tiles/src/main/java/com/google/android/horologist/tiles/complication/ComplicationTemplate.kt
@@ -19,10 +19,12 @@ package com.google.android.horologist.tiles.complication
 import android.content.Context
 import androidx.wear.watchface.complications.data.ComplicationData
 import androidx.wear.watchface.complications.data.ComplicationType
+import com.google.android.horologist.tiles.ExperimentalHorologistTilesApi
 
 /**
  * A complication renderer from Data to any of the supported types.
  */
+@ExperimentalHorologistTilesApi
 public interface ComplicationTemplate<D> {
     public val context: Context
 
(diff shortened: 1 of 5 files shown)


yschimke:
Not required, just noticed it missing.

yschimke:
@ataulm going to land, happy to fix with any review fixes.

LouisCAD:
There's a link between Tiles, Watch Face complications, and Horologist? I'm not sure I get it 🤔

ataulm:
Horologist provides some helpers for tiles and helpers for complications (there's no link between tiles and complications).

tiles/src/main/java/com/google/android/horologist/tiles/complication/ComplicationTemplate.kt
```diff
@@ -19,10 +19,12 @@ package com.google.android.horologist.tiles.complication
 import android.content.Context
 import androidx.wear.watchface.complications.data.ComplicationData
 import androidx.wear.watchface.complications.data.ComplicationType
+import com.google.android.horologist.tiles.ExperimentalHorologistTilesApi
 
 /**
  * A complication renderer from Data to any of the supported types.
  */
+@ExperimentalHorologistTilesApi
 public interface ComplicationTemplate<D> {
```
ataulm, line 28: @yschimke maybe @LouisCAD is referring to the package name, in which case +1, we should move

LouisCAD: Yes, the fact that it's in a tiles artifact, in a tiles package and, and has a `@ExperimentalHorologistTilesApi` annotation makes it very confusing, and looks like a mistake.

PR #673, opened 2022-10-17
Use Service's lifecycleScope

yschimke:
#### WHAT

Assume LifecycleService or LifecycleOwner for all services. Use lifecycleScope for the CoroutineScope.

#### WHY

No point creating an additional CoroutineScope and managing it's lifecycle.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
nice!

PR #675, opened 2022-10-17
[RFC] Streaming UI mode

yschimke:
#### WHAT

Streaming UI mode - consider what code paths would be needed for testing streaming instead of download only.

#### WHY

Things like prefetch only make sense with streaming, and so the reference app isn't useful for testing.

#### HOW

TODO add a settings for download/streaming and then make the UI reasonably usable in both modes.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
@luizgrp purely for discussion, I don't think the changes to media-ui are right, but want to discuss first.

It's useful for testing to have a streaming mode.

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/entity/PlaylistDownloadScreen.kt
```diff
@@ -142,7 +142,7 @@ public fun PlaylistDownloadScreen(
 }
 
 @Composable
-private fun MediaContent(
+internal fun MediaContent(
```
luizgrp, line 145: can't to see the usage in the code review, just checking if this change is needed/intended?

luizgrp (approved):
Sounds good 👍 

Left comments for discussion.

PR #677, opened 2022-10-17
Changes ShuffleButton to be a ToggleButton

fstanis:
#### WHAT
Makes `ShuffleButton` togglable.

#### WHY
For accessbility and localization purposes, being a plain `Button` isn't enough to convey state and how it changes.

#### HOW
Changes `Button` into `ToggleButton` in the composable.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/components/controls/ShuffleButton.kt
```diff
@@ -21,35 +21,50 @@ import androidx.compose.material.icons.filled.Shuffle
...
         enabled = enabled,
-        colors = colors
+        colors = colors,
+        checked = shuffleOn
     ) {
         val icon = if (shuffleOn) Icons.Default.ShuffleOn else Icons.Default.Shuffle
         Icon(
             imageVector = icon,
-            contentDescription = stringResource(id = R.string.horologist_shuffle_button_content_description),
+            contentDescription = "",
```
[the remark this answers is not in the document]
fstanis: Done

media-ui/src/main/java/com/google/android/horologist/media/ui/components/controls/ShuffleButton.kt
```diff
@@ -21,35 +21,50 @@ import androidx.compose.material.icons.filled.Shuffle
...
         enabled = enabled,
-        colors = colors
+        colors = colors,
+        checked = shuffleOn
     ) {
         val icon = if (shuffleOn) Icons.Default.ShuffleOn else Icons.Default.Shuffle
         Icon(
             imageVector = icon,
-            contentDescription = stringResource(id = R.string.horologist_shuffle_button_content_description),
+            contentDescription = null,
```
luizgrp, line 67: nit: we have a const `DECORATIVE_ELEMENT_CONTENT_DESCRIPTION`

yschimke (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/components/controls/ShuffleButton.kt
```diff
@@ -21,35 +21,50 @@ import androidx.compose.material.icons.filled.Shuffle
...
+import androidx.wear.compose.material.ToggleButtonDefaults.toggleButtonColors
 import com.google.android.horologist.media.ui.ExperimentalHorologistMediaUiApi
 import com.google.android.horologist.media.ui.R
 import com.google.android.horologist.media.ui.semantics.CustomSemanticsProperties.iconImageVector
 
 @ExperimentalHorologistMediaUiApi
 @Composable
 public fun ShuffleButton(
-    onClick: () -> Unit,
+    onToggle: (Boolean) -> Unit,
```
luizgrp, line 40: should we align the signature of this function with the ToggleButton API? or we'd rather it to abstract the implementation details?

I'm easy, if we do the former, that's my suggestion of changes:
- ShuffleButton -> ShuffleToggleButton
- onToggle -> onCheckedChange
- shuffleOn -> checked
- change order of params: 1st - checked, 2nd - onCheckedChange

luizgrp (approved):
nice catch!

media-ui/src/main/java/com/google/android/horologist/media/ui/components/controls/ShuffleButton.kt
fstanis: Good call - I renamed it to `ShuffleToggleButton` and changed the order. However, I kept `shuffleOn` and `onToggle` - I feel the concept of "checked" can be confusing here since it's not actually a checkbox. But let me know if you think we should follow it fully.

luizgrp: sounds good 👍

Issue #678, opened 2022-10-17
Implement cancel download in progress in Browse screen


luizgrp:
### WHAT

In the Browse screen, if the user taps on a playlist that is currently being downloaded, it should prompt the user if they wish to cancel download, and cancel the playlist download if they decide to proceed.

### WHY

In order to align with Figma specs.

### HOW

Implementation should be similar to what's implemented in [UampEntityScreen](https://github.com/google/horologist/blob/2a9759acc2c5665d67dcafae791715628125eb56/media-sample/src/main/java/com/google/android/horologist/mediasample/ui/entity/UampEntityScreen.kt) and [UampEntityScreenViewModel](https://github.com/google/horologist/blob/2a9759acc2c5665d67dcafae791715628125eb56/media-sample/src/main/java/com/google/android/horologist/mediasample/ui/entity/UampEntityScreenViewModel.kt#L92)

luizgrp:
Implementation should be similar to what's implemented in [UampEntityScreen](https://github.com/google/horologist/blob/2a9759acc2c5665d67dcafae791715628125eb56/media-sample/src/main/java/com/google/android/horologist/mediasample/ui/entity/UampEntityScreen.kt) and [UampEntityScreenViewModel](https://github.com/google/horologist/blob/2a9759acc2c5665d67dcafae791715628125eb56/media-sample/src/main/java/com/google/android/horologist/mediasample/ui/entity/UampEntityScreenViewModel.kt#L92)

luizgrp:
Blocked by https://github.com/google/horologist/issues/681

yschimke:
Low priority

PR #679, opened 2022-10-17
Add separate click handler for item with download in progress and custom action label

luizgrp:
#### WHAT

Add separate click handler for item with download in progress and custom action label.

#### WHY

In order to align with Figma specs.

#### HOW

- Add `onDownloadItemInProgressClick` param to `PlaylistDownloadBrowseScreen` - pending implementation / tracked in https://github.com/google/horologist/issues/678
- Add `onDownloadItemInProgressClickActionLabel` param to `PlaylistDownloadBrowseScreen`

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
But torn on adding it in but not activated, but makes the UI more complete. And deliberate decision.

PR #686, opened 2022-10-19
Allow volume decreases

yschimke:
#### WHAT

Allow volume decreases.

Also removes a redundant volume scroll listener.

#### WHY

Bug with volume decreases not working.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
cc @clickxu

clickxu (approved):
Thanks for fixing this.

PR #692, opened 2022-10-21
Update translations

fstanis:


media-ui/src/main/res/values-pt-rBR/strings.xml
```diff
@@ -0,0 +1,45 @@
...
+  <string name="horologist_browse_downloads_empty">Faça o download de músicas para começar a ouvir</string>
+  <string name="horologist_browse_library_title">Biblioteca</string>
+  <string name="horologist_browse_library_playlists">Playlists</string>
+  <string name="horologist_browse_library_playlists_button">Playlists</string>
+  <string name="horologist_browse_library_settings_button">Configurações</string>
+  <string name="horologist_browse_playlist_title">Playlists</string>
+  <string name="horologist_playlist_download_button_download">Download no relógio</string>
+  <string name="horologist_playlist_download_button_cancel">Cancelar</string>
+  <string name="horologist_playlist_download_button_download_content_description">Download disponível</string>
+  <string name="horologist_playlist_download_button_cancel_content_description">Cancelar downloads</string>
```
luizgrp, line 39: nit: this string was removed in https://github.com/google/horologist/pull/683/files#diff-3776025d54383445560711eddca73fcf77ac40863cbd7ac8c68225f0315ec67e

luizgrp (approved):
nice one!

it looks like it needs another sync, as new strings were added and some removed

media-ui/src/main/res/values-pt-rBR/strings.xml
fstanis: Unfortunately, this may happen sometimes because translations lag behind git HEAD and this is largely automated. It'll be fixed in the next update. #693 disables the linter check for this.

audio-ui/src/main/res/values-b+es+419/strings.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="utf-8"?><resources>
```
yschimke, line 1: Does this map to this? https://github.com/androidx/media/tree/release/libraries/ui/src/main/res/values-b%2Bsr%2BLatn

Are we sure the path is right?

PR #693, opened 2022-10-21
Disable linter for translations

fstanis:


audio-ui/build.gradle
```diff
@@ -82,6 +82,10 @@ android {
         checkReleaseBuilds false
         textReport true
     }
+    lintOptions{
+        disable 'MissingTranslation'
+        disable 'ExtraTranslation'
```
luizgrp, line 87: would it not be better to keep this one so the extra translations are identified and can be removed manually?
unless the plans are to have this done automatically via sync?

luizgrp (approved):

audio-ui/build.gradle
luizgrp: got answer in https://github.com/google/horologist/pull/692#discussion_r1001649365

fstanis: They'll be removed as part of the automated process - the issue is that these linter checks right now may prevent that process because they're targetting autogenerated files (#692 right now), so I'd rather we don't have them.

PR #695, opened 2022-10-21
Implement SquareSegmentedProgressIndicator

oas004:
#### WHAT
Here I have tried to implement the square version of the `SegmentedProgressIndicator` . 

#### WHY
Because of #431 

#### HOW
I have made the API similar to `SegmentedProgressIndicator`. The difficulties on this comes from painting the rounded square. There is no method in the Canvas to draw a progress of a rounded square shape. I have tried to do this by combining `drawArch` and `drawLine` based on these sections.

Now, the implementation is too tied to having a sum of 8 on the section weight. I need to decouple that somehow. I also kinda want the stroke cap to be round. If I can't have one end round and one square, I am unsure if I can make that possible with this approach. The reason is that then you would be able to see the original 8 sections even though they are supposed to be combined into one section.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

image


yschimke:
Sorry, I was on vacation last week. I'll take a look in next couple of days.

yschimke:
It looks really good.  I'm still working through it and might put up some changes for your PR.

Main feedback

- The anim preview is great, but should use updateTransition so preview tooling works

```
    val transition = updateTransition(
        targetState = PreviewAnimationState.Start,
        label = "Square Progress Indicator"
    )

    val progress by transition.animateFloat(
        label = "Progress",
        targetValueByState = {
            it.target
        },
        transitionSpec = {
            tween(durationMillis = 1000, easing = LinearEasing)
        }
    )
``` 

- We should follow the conventions from SegmentedProgressIndicator, so maybe use ProgressIndicatorSegment as the segment types, add trackColor as a param.

- Define progressSemantics on Canvas

- The padding 38 hides drawing outside the canvas I think.

yschimke:
I'm going to try a different strategy.  Basically artificially inserting the gaps as artificial segments, drawing as the 8 adjacent lines/arcs. And for each line/arc drawing a subset based on the progress through my computed segments.

oas004:
> It looks really good. I'm still working through it and might put up some changes for your PR.
> 
> Main feedback
> 
> * The anim preview is great, but should use updateTransition so preview tooling works
> 
> ```
>     val transition = updateTransition(
>         targetState = PreviewAnimationState.Start,
>         label = "Square Progress Indicator"
>     )
> 
>     val progress by transition.animateFloat(
>         label = "Progress",
>         targetValueByState = {
>             it.target
>         },
>         transitionSpec = {
>             tween(durationMillis = 1000, easing = LinearEasing)
>         }
>     )
> ```
> 
>  * We should follow the conventions from SegmentedProgressIndicator, so maybe use ProgressIndicatorSegment as the segment types, add trackColor as a param.
> * Define progressSemantics on Canvas
> * The padding 38 hides drawing outside the canvas I think.
> * for comparing against SegmentedProgressIndicator, we should reuse same colours.

Sorry for late reply, my girlfriend had a birthday yesterday I had to attend. Thanks for awesome suggestions!  I have in 
0cdb56d moved the previews out to the debug source set, added the updateTransition and made sure that they work in the preview tooling as well :) I have also changed the colors and the api to match the `SegmentedProgressIndicator`. I also added the progressSemantics modifier to the Canvas composable (is that correct?) :) For the padding, I think you are correct. Is that something that should be addressed in the drawing algorithm? Or should the user use the modifier to have whatever the padding they want? :)

The strategy sounded kinda advanced, but I think it is a great idea if it works! Feel free to hijack the PR if that is easier :)

yschimke:
No plan to hijack, wanted to present some options.

oas004:
> No plan to hijack, wanted to present some options.

Alright :) I had some sleep on the idea and I was wondering; Is it possible to have like an inverse stroke cap? Or should the cap just be a rectangle like it is now? :)

yschimke:
Not sure. It ends up overlapping depending on how you configure the Stroke and padding.  But something to play with.

yschimke:
@oas004 let me know when it's ready to take a look.

oas004:
> @oas004 let me know when it's ready to take a look.

Thank you man! I think I introduced one defect when there is a lot of segments. Will try to fix that and let you know😊

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
```diff
@@ -0,0 +1,457 @@
...
+@ExperimentalHorologistComposablesApi
+@Composable
+public fun SquareSegmentedProgressIndicator(
+    modifier: Modifier = Modifier,
+    progress: Float,
+    strokeWidth: Dp = ProgressIndicatorDefaults.StrokeWidth,
+    trackColor: Color = MaterialTheme.colors.onBackground.copy(alpha = 0.1f),
+    cornerRadiusDp: Dp = 10.dp,
+    trackSegments: List<ProgressIndicatorSegment>,
+    paddingDp: Dp = ProgressIndicatorDefaults.StrokeWidth
```
oas004, line 64: Here the user can give a padding that is going to be in-between the segments (so the gap size), would it make sense to maybe make this an enum and have fixed sizes? Or say that it can't be bigger than 10 or something? If it is bigger than 10 it starts to almost not make sense. Just thinking out loud here, what do you guys think? :)  Maybe following the same as api as `SegmentedProgressIndicator` makes sense 🤔

oas004:
@yschimke I think this is ready to be looked at. I took a lot of things from your branch https://github.com/yschimke/horologist/tree/sspb that was way better! That helped a lot! Thank you so much for that! 🙏

composables/src/debug/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicatorPreview.kt
```diff
@@ -0,0 +1,196 @@
...
+import kotlinx.coroutines.delay
+
+enum class PreviewAnimationState(val target: Float) {
+    Start(0f), End(1f)
+}
+
+@OptIn(ExperimentalHorologistComposablesApi::class)
+@WearSquareDevicePreview
+@Composable
+fun PreviewProgressAnimation() {
```
oas004, line 56: I think this got very nice now! 😄

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
```diff
@@ -0,0 +1,458 @@
...
+) {
+    check(progress in 0.0..1.0) {
+        "Only progress between 0.0 and 1.0 is allowed."
+    }
+
+    check(cornerRadiusDp in 0.dp..90.dp) {
+        "If your corner radius is above 90, you should use SegmentedProgressIndicator."
```
oas004, line 72: Is this enough? Or should I expand on it? :)

composables/src/test/kotlin/com/google/android/horologist/composables/SquareSegmentedProgressIndicatorTest.kt
```diff
@@ -0,0 +1,185 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.dp
+import app.cash.paparazzi.DeviceConfig
+import com.google.android.horologist.paparazzi.ExperimentalHorologistPaparazziApi
+import com.google.android.horologist.paparazzi.WearPaparazzi
+import org.junit.Rule
+import org.junit.Test
+
+class SquareSegmentedProgressIndicatorTest {
```
oas004, line 33: Do I need to add more Paparazzi tests? Or is this enough?

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
yschimke: +1, can't answer on this specific. But copying api and features from SPI makes sense.

oas004: I think the API from SPI is that the user can choose the padding angle which in that case is the equivalent of our spacing here. If they choose 360f then the entire circle is just a padding I would guess, and I think that might be a bit strange 🤔

composables/src/test/kotlin/com/google/android/horologist/composables/SquareSegmentedProgressIndicatorTest.kt
yschimke: I think enough, expand more as we find a reason.

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
```diff
@@ -0,0 +1,458 @@
...
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import androidx.wear.compose.material.MaterialTheme
+import androidx.wear.compose.material.ProgressIndicatorDefaults
+
+/**
+ * SquareSegmentedProgressIndicator represents a segmented progress indicator with
+ * square shape.
+ *
+ * @param modifier
```
yschimke, line 45: nit: 
```suggestion
 * @param modifier The modifier to be applied to the layout.
```

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
```diff
@@ -0,0 +1,458 @@
...
+    trackColor: Color = MaterialTheme.colors.onBackground.copy(alpha = 0.1f),
+    cornerRadiusDp: Dp = 10.dp,
+    trackSegments: List<ProgressIndicatorSegment>,
+    paddingDp: Dp = ProgressIndicatorDefaults.StrokeWidth
+) {
+    check(progress in 0.0..1.0) {
+        "Only progress between 0.0 and 1.0 is allowed."
+    }
+
+    check(cornerRadiusDp in 0.dp..90.dp) {
```
yschimke, line 71: I'm not sure this is true. Perhaps we should constrain it based on strokeWidth? or make it a boolean for none/strokeWidth.  It's not in degrees, so not sure 90 make sense here.

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
yschimke: I'm not sure this is true. Perhaps we should constrain it based on strokeWidth? or make it a boolean for none/strokeWidth. It's not in degrees, so not sure 90 make sense here.

composables/src/debug/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicatorPreview.kt
yschimke: Yep - it's awesome

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
oas004: Yeah I think you are correct when I think about it. I guess this will depend on the size as well, so I will remove this check. I'm not sure I get 100% what you mean about constraining the corner radius based on StrokeWidth. Do you mean the API should be like this?

```
@ExperimentalHorologistComposablesApi
@Composable
public fun SquareSegmentedProgressIndicator(
    progress: Float,
    strokeWidth: Dp = ProgressIndicatorDefaults.StrokeWidth,
    trackColor: Color = MaterialTheme.colors.onBackground.copy(alpha = 0.1f),
    shouldHaveStrokeWidthCornerRadius: Boolean = false,
    trackSegments: List<ProgressIndicatorSegment>,
    paddingDp: Dp = ProgressIndicatorDefaults.StrokeWidth
)
``` 
and then setting it to none if it is false making it a square? :)

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
oas004: Nice suggestion!Thanks for pointing it out! I added it to the [5985e44](https://github.com/google/horologist/pull/695/commits/5985e44b15ab7ce73f320a35006006496d35dabe) commit and rebased to update the branch. Hope that was okay :)

composables/src/main/java/com/google/android/horologist/composables/SquareSegmentedProgressIndicator.kt
yschimke: Sorry was mixing this up with the gap.  But the rest of my comment stands.

yschimke (approved):

PR #705, opened 2022-10-26
Changes SeekButtonIncrement to only have two subclasses

fstanis:
#### WHAT
Changes `SeekButtonIncrement` to only have two subclasses

#### WHY
Cleaner API and fixes a few bugs that can happen with object identity (e.g. `Other(5)` used to be different from `Five`).

#### HOW
`SeekButtonIncrement` now only has `Unknown` and `Known` subclasses, `Known` is a data class and the companion object has `val`s to replace old subclasses.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/components/controls/SeekButtonIncrement.kt
```diff
@@ -17,30 +17,21 @@
...
-
-    public object Ten : SeekButtonIncrement(10)
-
-    public object Thirty : SeekButtonIncrement(30)
+public sealed class SeekButtonIncrement {
+    public object Unknown : SeekButtonIncrement() {
+        override fun toString(): String = "SeekButtonIncrement(UNKNOWN)"
+    }
 
-    public data class Other(override val seconds: Int) : SeekButtonIncrement(seconds)
```
luizgrp, line 33: I'm happy with the changes, but will just bring the following for consideration:

The scenario described in the PR description:

> Other(5) used to be different from Five

Is valid to me, e.g.: the user would like to display the [Replay icon](https://fonts.corp.google.com/icons?icon.query=Replay&icon.platform=android) with 5 seconds instead of the [Replay 5 icon](https://fonts.corp.google.com/icons?icon.query=Replay+5&icon.platform=android) with 5 seconds.

With these changes, it might not be clear enough to the user that a "Known(5 or 10 or 30)" will display a different icon than "Known(15 or any other number not in 5,10,30)".

---

The initial idea was to have a model that represent each available icon for the UI, plus the extra "Other" that defaults to the "Unknown" icon. Those definitions seemed more close to the UI representation, whereas "Known"/"Unknown" seem closer to a domain representation.

Maybe what is causing confusion is the fact that "Other" and "Unknown" display the same icon?

Maybe the kdocs of current code should be improved to clarify that those sealed classes will display a different icon?

luizgrp (approved):
LGTM

media-ui/src/main/java/com/google/android/horologist/media/ui/components/controls/SeekButtonIncrement.kt
fstanis: I see - I wasn't aware of this context. However, that means the original sealed class is tightly coupled with icons that are available and that's not ideal. I'm inclined to remove `Five`, `Ten` and `Thirty` from this class all together and add a `mapOf<SeekButtonIncrement, Icon>` instead somewhere, possibly as a parameter to `SeekForwardButton` and `SeekBackButton`. WDYT?

luizgrp: Why do you think that is not ideal having the sealed classes tightly coupled with the icons available?
I see that as a good way of modelling what it is possible to be displayed on the screen, but I might have not seen your perspective yet. Maybe elaborate the following sentence of the PR description: "fixes a few bugs that can happen with object identity", what have you faced?

On this new proposal, would Seek(Back/Forward)Button also receive `Icon` as param? Or would they use the map to retrieve the icon? If the latter, then we would have the same issue of not be able to have 5 increment with "Replay icon", if I understood correctly.

fstanis: My issue with the current implementation is that `SeekButtonIncrement` is part of the UI state (and the `PlayerUiState` object, consequently) but is tightly coupled with an external depedency (`androidx.compose.material.icons`). The simplest side-effect of this is that any tests relying on `PlayerUiState` may fail if the icons change, for example if androidx adds a 20s icon and we add that to `SeekButtonIncrement`, any assertion for `Other(20)` will now fail. This failure is valid for strictly UI layer components, but `PlayerUiState` may be used in a UI that doesn't expose icons at all, making that test failure not useful.

Because icons are IMO part of the presentation and not state, I'd argue that `Other(5)` and `Five` equivalent on state level, so the decision of whether 5s has a special icon should be decided on composable level.

TLDR: I'd make the icon decided in the composable signature:

```diff
@ExperimentalHorologistMediaUiApi
@Composable
public fun SeekBackButton(
    onClick: () -> Unit,
    seekButtonIncrement: SeekButtonIncrement,
+   incrementToIconMapping: Map<SeekButtonIncrement, Icon> = mapOf(
+       SeekButtonIncrement.Known(5) to Icons.Default.Replay5
+       SeekButtonIncrement.Known(10) to Icons.Default.Replay10
+       SeekButtonIncrement.Known(30) to Icons.Default.Replay30
+   ),
+   defaultIcon: Icon = Icons.Default.Replay,
    modifier: Modifier = Modifier,
    enabled: Boolean = true,
    colors: ButtonColors = MediaButtonDefaults.mediaButtonDefaultColors,
```

luizgrp: Thanks for elaborating on that. I think the SeekButtonIncrement model itself is not tightly coupled with material icons, the composables SeekBackButton and SeekForwardButton that are, they use the model SeekButtonIncrement to choose the icons. You are right that SeekButtonIncrement was modelled based on the available options in material icons, but there is nothing holding us in providing our own custom icons for other values. In fact, material icons doesn't provide a [generic Forward icon](https://fonts.corp.google.com/icons?icon.query=forward) so the Replay was used and [flipped](https://fonts.corp.google.com/icons?icon.query=forward) horizontally :D

---

I think this suggested API is a bit convoluted. SeekButtonIncrement was created to have a single param to pass the increment value + icon. If we want that choice to be decoupled, I see little benefit in keeping SeekButtonIncrement. Maybe it can be deleted and SeekBackButton just receive "seconds" and "icon" as params?

fstanis: > benefit in keeping SeekButtonIncrement

The only one I see is to avoid having a nullable type for unknown duration, i.e. `if (increment == SeekButtonIncrement.Unknown)` is better than checking for null. But maybe we can find a better solution for that, e.g. alias zero duration or infinity as unknown.

> just receive "seconds" and "icon" as params

I like this. Maybe something like this then:

```diff
public object MediaButtonDefaults {
    ...

+   public fun seekBackIcon(seekButtonIncrement: SeekButtonIncrement) =
+       when(seekButtonIncrement) {
+           SeekButtonIncrement.Known(5) -> Icons.Default.Replay5
+           ...
+       }
}

@ExperimentalHorologistMediaUiApi
@Composable
public fun SeekBackButton(
    onClick: () -> Unit,
    seekButtonIncrement: SeekButtonIncrement,
+   icon: Icon = MediaButtonDefaults.seekBackIcon(seekButtonIncrement),
    modifier: Modifier = Modifier,
    enabled: Boolean = true,
    colors: ButtonColors = MediaButtonDefaults.mediaButtonDefaultColors,
```

luizgrp: sounds good 👍

yschimke (approved):
LGTM

PR #707, opened 2022-10-28
Add base-ui module

luizgrp:
#### WHAT

Add `base-ui` module.

#### WHY

In order to host the base UI components as per redlines of our Figma project.

Having them in a separate module will allow these components to also be used in the sample apps, which are now duplicating code.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
Are these now considered reusable components wider than media?

luizgrp:
> Are these now considered reusable components wider than media?

yes, the components in this module are following the redlines of the "compose for wear os component spec" figma project, which is not specific to media.

yschimke:
@luizgrp awesome idea!  Hopefully if we can publish the Figma at some point, these can back them up.

PR #714, opened 2022-11-01
Fix lint error ExpiredTargetSdkVersion

luizgrp:
#### WHAT

Fix lint error `ExpiredTargetSdkVersion`.

#### WHY

Our builds are now failing with:

```
Error: Google Play requires that apps target API level 31 or higher.
 [ExpiredTargetSdkVersion]
        targetSdk 30
```

#### HOW

Add uses-feature hardware type watch to AndroidManifest file of `media-sync` and `datalayer` modules.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/src/main/AndroidManifest.xml
+++ b/datalayer/src/main/AndroidManifest.xml
@@ -18,4 +18,6 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.google.android.horologist.datalayer">
 
+    <uses-feature android:name="android.hardware.type.watch" />
+
 </manifest>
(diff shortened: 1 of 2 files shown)


fstanis (approved):

datalayer/src/main/AndroidManifest.xml
```diff
@@ -18,4 +18,6 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.google.android.horologist.datalayer">
 
+    <uses-feature android:name="android.hardware.type.watch" />
```
yschimke, line 21: Not sure this works since datalayer is a module for Wear and mobile.

yschimke (changes requested):

datalayer/src/main/AndroidManifest.xml
luizgrp: can revert the changes of this PR in https://github.com/google/horologist/pull/713 if you think it's a better alternative

yschimke: Partial revert?

PR #716, opened 2022-11-01
Disable ExpiredTargetSdkVersion lint check

luizgrp:
#### WHAT

Disable `ExpiredTargetSdkVersion` lint check.

Related issue: https://issuetracker.google.com/issues/196420849

#### WHY

Our builds are now failing with:

```
Error: Google Play requires that apps target API level 31 or higher.
 [ExpiredTargetSdkVersion]
        targetSdk 30
```

Due to this [check](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:lint/libs/lint-checks/src/main/java/com/android/tools/lint/checks/GradleDetector.kt;l=252;drc=2457a6388492397c8bb0d6f1af7f6e00cdfb2d49).

However, Wear OS apps required target API level are still 30, according to: https://support.google.com/googleplay/android-developer/answer/11926878

#### HOW

Add baseline lint files for each module, with only `ExpiredTargetSdkVersion` listed there.

Other alternatives that were tried but did NOT work:

- Current disable lint option is not being taken into consideration: https://github.com/google/horologist/blob/8d1b285f5819a9cdc17185217fef4f1bdd1d041b/build.gradle#L209
- Disable lint option in each module's build.gradle
- Comment suppress
```groovy
        //noinspection ExpiredTargetSdkVersion
        targetSdk 30
```

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/build.gradle
+++ b/datalayer/build.gradle
@@ -62,6 +62,8 @@ android {
     lint {
         checkReleaseBuilds false
         textReport true
+
+        baseline(file("quality/lint/lint-baseline.xml"))
     }
     namespace 'com.google.android.horologist.datalayer'
 }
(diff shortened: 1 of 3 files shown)


luizgrp:
Could not reopen https://github.com/google/horologist/pull/713 - I guess because the original branch was deleted 🙄

yschimke (approved):
LGTM - but also ok with reducing the scope to just datalayer.

Issue #715, opened 2022-11-01
Refactor EntityDialog


laiyichin:
Refactor [EntityDialog](https://github.com/google/horologist/blob/main/media-sample/src/main/java/com/google/android/horologist/mediasample/ui/entity/UampEntityScreen.kt#L166) to move it from UampEntityScreen to [base-ui](https://github.com/google/horologist/tree/main/base-ui/src/debug/java/com/google/android/horologist/base/ui/components) and add [preview](https://github.com/google/horologist/tree/main/base-ui/src/debug/java/com/google/android/horologist/base/ui/components) and [tests](https://github.com/google/horologist/tree/main/base-ui/src/test/java/com/google/android/horologist/base/ui/components).

luizgrp:
Redlines:
![Screen Shot 2022-11-01 at 20 06 28](https://user-images.githubusercontent.com/878134/199330825-751e9060-701f-428e-9a7b-9a52de3bd0b5.png)

Could also rename it, it was called EntityDialog because it was only used in the EntityScreen.

luizgrp:
@laiyichin was this issue fully done in https://github.com/google/horologist/pull/718?

Do you remember if we kept this ticket open because is missing snapshot tests for it?

PR #718, opened 2022-11-02
Entity dialog refactoring

laiyichin:
#### WHAT
Refactor EntityDialog in UampEntityViewScreen to AlertDialog in base-ui.

#### WHY
This is a commonly used composable that could be abstracted out and re-used by other components.

https://github.com/google/horologist/issues/715

#### HOW
- Move the composable to the base-ui
- Add preview in debug
- Split the composable into two composable -  dialog + alert because dialog with custom content cannot be previewed yet

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

UampEntityScreen Cancel Playlist Download
cancel_download
UampEntityScreen Remove Downloaded Playlist 
remove_download


base-ui/src/main/java/com/google/android/horologist/base/ui/components/AlertDialog.kt
```diff
@@ -0,0 +1,93 @@
...
+ */
+@Composable
+public fun AlertDialog(
+    text: String,
+    proceedText: String,
+    cancelText: String,
+    onCancelButtonClick: () -> Unit,
+    onProceedButtonClick: () -> Unit,
+    showDialog: Boolean,
+    scalingLazyListState: ScalingLazyListState
```
luizgrp, line 42: It is good practice in compose to have a Modifier param in each composable function, and that param should be passed only to the top level compose of the hierarchy, in this case, to Dialog.

Good read related: https://chris.banes.dev/posts/always-provide-a-modifier/

I know you only moved this function from one class to another, but `EntityDialog` was private, not exposed to general use, that's why it wasn't built with a Modifier param (but it should have had anyway tbh).

Ah, and there is also a convention about the order of the [Modifier](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier) param:

>  It should appear as the first optional parameter in the parameter list; after all required parameters (except for trailing lambda parameters) but before any other parameters with default values.

base-ui/src/main/java/com/google/android/horologist/base/ui/components/AlertDialog.kt
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * This composable fulfils the redlines of the following components:
+ * - AlertDialog
+ */
+@Composable
+public fun AlertDialog(
+    text: String,
+    proceedText: String,
+    cancelText: String,
```
luizgrp, line 38: given the texts are not visible on the screen, could you please rename these params to:
- okButtonContentDescription
- cancelButtonContentDescription

Could you also please add default values to them:
- "OK"
- "Cancel"

These values should be in the `res/values/strings.xml` in `base-ui` module.

base-ui/src/main/java/com/google/android/horologist/base/ui/components/AlertDialog.kt
```diff
@@ -0,0 +1,93 @@
...
+import androidx.compose.ui.text.style.TextAlign
+import androidx.wear.compose.material.MaterialTheme
+import androidx.wear.compose.material.ScalingLazyListState
+import androidx.wear.compose.material.Text
+import androidx.wear.compose.material.dialog.Alert
+import androidx.wear.compose.material.dialog.Dialog
+
+/**
+ * This composable fulfils the redlines of the following components:
+ * - AlertDialog
```
luizgrp, line 32: nit: could you please change it to:

```
* - AlertDialog - Title + body + buttons
```

luizgrp (commented):
Looking good, left few comments related to the specs in Figma, which I did not add fully to the ticket, my apologies!

Could you also please add to the ticket description, screenshots of the dialogs (cancel downloads and remove downloads) of the entity screen that were changed in this PR?

Also, if you feel like doing it, [this](https://github.com/google/horologist/blob/4e7b4a6184a9a31526d852c426dd1475960b682e/media-sample/src/main/java/com/google/android/horologist/mediasample/ui/playlists/UampPlaylistsScreen.kt#L80) dialog from `UampPlaylistsScreen` could also be changed to use the `AlertDialog` introduced by this PR. I can raise a ticket if you prefer to not mix with this one for now 👍

base-ui/src/main/java/com/google/android/horologist/base/ui/components/AlertDialog.kt
luizgrp: please ignore this comment, Title was not implemented yet in this Dialog

base-ui/src/main/java/com/google/android/horologist/base/ui/components/AlertDialog.kt
```diff
@@ -0,0 +1,93 @@
...
+public fun AlertDialogAlert(
+    text: String,
+    proceedText: String,
+    cancelText: String,
+    onCancelButtonClick: () -> Unit,
+    onProceedButtonClick: () -> Unit
+) {
+    Alert(
+        title = {
+            Text(
```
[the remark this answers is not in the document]
luizgrp: Please ignore the comment above, I just realised that this Dialog does not have a title.

I updated the redlines in https://github.com/google/horologist/issues/715, this Text here should follow the specs of the "Body" component. So no max lines, style Body2, etc

luizgrp (commented):
just realised that this dialog does not have a title yet, see my new comments!

base-ui/src/main/java/com/google/android/horologist/base/ui/components/AlertDialog.kt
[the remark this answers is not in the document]
laiyichin: Added title and body both.

base-ui/src/main/java/com/google/android/horologist/base/ui/components/AlertDialog.kt
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * This composable fulfils the redlines of the following components:
+ * - AlertDialog - Title + body + buttons
+ */
+@Composable
+public fun AlertDialog(
+    body: String,
+    onCancelButtonClick: () -> Unit,
+    onProceedButtonClick: () -> Unit,
```
luizgrp, line 41: nit: in order to avoid confusion, maybe this param should be called `onOKButtonClick` in order to match the name of param `okButtonContentDescription` - it could be the other way around as well `onProceedButtonClick` and `proceedButtonContentDescription`, but "OK" is what's written in Figma

base-ui/src/debug/java/com/google/android/horologist/base/ui/components/AlertDialogAlertPreview.kt
```diff
@@ -0,0 +1,46 @@
...
+        okButtonContentDescription = "Ok",
+        cancelButtonContentDescription = "Cancel"
+    )
+}
+
+@WearPreviewDevices
+@Composable
+fun AlertDialogAlertWithLongBodyPreview() {
+    AlertDialogAlert(
+        title = "Title",
```
luizgrp, line 39: nit: given title has a constraint of max 3 lines, it would be nice to have a preview with a long title text to show how it's displayed / truncated

base-ui/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,20 @@
...
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<resources>
+    <string name="default_ok_button_content_description">OK</string>
+    <string name="default_cancel_button_content_description">Cancel</string>
```
luizgrp, line 19: nit: name could be changed to imply that these strings are specific to "AlertDialog" rather than being for "general purposes" (we should avoid having that), e.g.: `alert_dialog_ok_button_content_description`

luizgrp (approved):
LGTM, thanks for doing this!

PR #724, opened 2022-11-04
Update comments for Modifier.edgeSwipeToDismiss

BowerSteve:
Modifier.edgeSwipeToDismiss must exist within a SwipeToDismissBox that uses the same SwipeToDismissBoxState. Otherwise, edgeSwipeToDismiss waits indefinitely for the anchors to be provided that would tell it the swipeable limits.


sample/src/main/java/com/google/android/horologist/navsample/NavWearApp.kt
```diff
@@ -145,6 +145,10 @@ fun NavWearApp(
 
             val pagerState = rememberPagerState()
             PagerScreen(
+                // When using Modifier.edgeSwipeToDismiss, it is required that the element on
+                // which the modifier applies exists within a SwipeToDismissBox which shares
```
yschimke, line 149: SwipeToDismissBox or SwipeDismissableNavHost?

yschimke (approved):
Pending one comment for discussion.

sample/src/main/java/com/google/android/horologist/navsample/NavWearApp.kt
BowerSteve: Either, I guess, because SwipeDismissableNavHost passes it on to a SwipeToDismissBox.

PR #728, opened 2022-11-09
Treat non-positive duration as unknown

fstanis:


media/src/main/java/com/google/android/horologist/media/model/MediaPosition.kt
```diff
@@ -48,7 +48,7 @@ public sealed class MediaPosition(
                 "Duration has to be greater than zero [current: $current] [duration: $duration]."
             }
             check(current <= duration) {
-                "Current position has to be greater than duration [current: $current] [duration: $duration]."
+                "Duration can't be less than current position [current: $current] [duration: $duration]."
```
luizgrp, line 51: oops! thanks!

luizgrp (approved):

PR #730, opened 2022-11-09
Moves icon out of DeviceChip

fstanis:
#### WHAT
Makes `DeviceChip` accept parameters like icon and description directly.

#### WHY
To make it easier to have custom icons for any subclass of `AudioOutput`.

#### HOW
By moving most of `DeviceChip` logic into `VolumeScreen` and turning it into parameters with defaults that match current behavior.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests

### DIFF
--- a/audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/DeviceChip.kt
+++ b/audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/DeviceChip.kt
@@ -16,12 +16,12 @@
 
 package com.google.android.horologist.audio.ui.components
 
+import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.IntrinsicSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.width
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.onClick
 import androidx.compose.ui.semantics.semantics
@@ -29,8 +29,6 @@ import androidx.compose.ui.semantics.stateDescription
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.wear.compose.material.Chip
 import androidx.wear.compose.material.ChipDefaults
-import androidx.wear.compose.material.Icon
-import androidx.wear.compose.material.MaterialTheme
 import androidx.wear.compose.material.Text
 import com.google.android.horologist.audio.ui.R
 
@@ -38,7 +36,7 @@ import com.google.android.horologist.audio.ui.R
 public fun DeviceChip(
     volumeDescription: String,
     deviceName: String,
-    icon: ImageVector,
+    icon: @Composable BoxScope.() -> Unit,
     onAudioOutputClick: () -> Unit,
     modifier: Modifier = Modifier
 ) {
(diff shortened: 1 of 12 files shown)


audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/DeviceChip.kt
```diff
@@ -37,32 +32,23 @@ import androidx.wear.compose.material.ChipDefaults
...
-import com.google.android.horologist.audio.VolumeState
 import com.google.android.horologist.audio.ui.R
 
 @Composable
 public fun DeviceChip(
-    volumeState: VolumeState,
-    audioOutput: AudioOutput,
+    volumeDescription: String,
+    deviceName: String,
+    icon: ImageVector,
```
luizgrp, line 41: I think this is a good change, so now this composable is independent of any "domain" model.

A strategy that we've done in the `media-ui` module is to always provide a [composable independent of domain](https://github.com/google/horologist/blob/9258c1d847e2a9e952c633fc9046bd02361684d9/media-ui/src/main/java/com/google/android/horologist/media/ui/components/MediaChip.kt#L71-L77), and then for some components we provide a [second version dependent on domain](https://github.com/google/horologist/blob/9258c1d847e2a9e952c633fc9046bd02361684d9/media-ui/src/main/java/com/google/android/horologist/media/ui/components/MediaChip.kt#L47-L48), that [calls](https://github.com/google/horologist/blob/9258c1d847e2a9e952c633fc9046bd02361684d9/media-ui/src/main/java/com/google/android/horologist/media/ui/components/MediaChip.kt#L57) that first version. 

Do you think this component could benefit from that strategy? That would keep these changes backwards compatible, given this component is not marked as experimental anymore.

luizgrp (approved):

audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/DeviceChip.kt
yschimke, line 41: Should we make this slightly more flexible?

MediaRouter has an optional icon Uri (or maybe drawable????) that could be used here instead of generic icon.  How about a slot here?

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:mediarouter/mediarouter/src/main/java/androidx/mediarouter/media/MediaRouter.java;l=1303;bpv=1

```
        /**
         * Gets the URI of the icon representing this route.
         * <p>
         * This icon will be used in picker UIs if available.
         * </p>
         *
         * @return The URI of the icon representing this route, or null if none.
         */
        @Nullable
        public Uri getIconUri() {
            return mIconUri;
        }
```

yschimke (approved):

audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/DeviceChip.kt
fstanis: Changed into slot.

fstanis: I feel this composable is a bit too simple and specific to have two versions, so I'd rather avoid that, but happy to discuss further if you feel strongly.

audio/src/main/java/com/google/android/horologist/audio/AudioOutput.kt
```diff
@@ -32,12 +32,19 @@ public interface AudioOutput {
      */
     public val name: String
 
+    /**
+     * Optional type of output which may be associated with an icon or displayed name.
+     */
+    public val type: String
+        get() = ""
```
yschimke, line 39: Should optional be nullable?

yschimke (approved):

audio/src/main/java/com/google/android/horologist/audio/AudioOutput.kt
fstanis: I prefer avoiding nullable, in this case there wouldn't be a difference between "" and null (code paths would likely do `?: ""` anyway), so I'd keep it non-nullable.

PR #731, opened 2022-11-10
Fix macrobenchmark and test running in CI

yschimke:
#### WHAT

Fix macrobenchmark and test running in CI

#### WHY

Had a recent issue with macrobenchmark not building after unrelated changes.

#### HOW

Run in a separate CI build.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/scripts/run-instrumented-tests.sh
+++ b/scripts/run-instrumented-tests.sh
@@ -23,6 +23,7 @@ SHARD_INDEX=0
 # By default we don't log
 LOG_FILE=""
 TEST_OPTIONS=""
+MACROBENCHMARK_RULES=""
 
 # Parse parameters
 for i in "$@"; do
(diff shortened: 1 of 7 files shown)


scripts/run-instrumented-tests.sh
```diff
@@ -93,7 +93,7 @@ fi
 
 # If we don't have a task yet, use the defaults
 if [[ -z "$TASK" ]]; then
-  TASK="connectedCheck"
+  TASK="connectedCheck -x :media-sample-benchmark:check"
```
luizgrp, line 96: I think this line never run because `TASK` will always have a value:

run-instrumented-tests is always [called](https://github.com/google/horologist/blob/9258c1d847e2a9e952c633fc9046bd02361684d9/.github/workflows/build.yml#L152) with `--run-affected` param, which then [sets](https://github.com/google/horologist/blob/9258c1d847e2a9e952c633fc9046bd02361684d9/scripts/run-instrumented-tests.sh#L84-L85) a value to TASK

yschimke: Good catch, I'll let this run complete and see what happens then fix.

PR #732, opened 2022-11-10
Bump targetSdk to fix ExpiredTargetSdkVersion lint error

luizgrp:
#### WHAT

Bump `targetSdk` to 33.

#### WHY

In order to fix `ExpiredTargetSdkVersion` lint error.

#### HOW

- Remove lint baseline added in https://github.com/google/horologist/pull/713.
- Bump version.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


datalayer/build.gradle
```diff
@@ -25,7 +25,7 @@ android {
 
     defaultConfig {
         minSdk 26
-        targetSdk 30
+        targetSdk 33
```
yschimke, line 28: Won't this cause issues for apps that use this library but that target 30?

https://stackoverflow.com/a/49034381/1542667

luizgrp: those apps can use [merge rule markers](https://developer.android.com/studio/build/manage-manifests#merge_rule_markers) to solve it, e.g. `tools:node="replace"` will make their apps to use their target

yschimke: That sounds like pain, are we really at the point where this Android Studio lint is completely broken for Wear libraries?

luizgrp: I think unfortunately it's the state of Wear that don't match the same required API as mobile https://support.google.com/googleplay/android-developer/answer/11926878.

If you are building a library that is not specific to wear, like this datalayer, then you should aim to target the latest API (33), but then wear apps that are on 30 will have to use that merge marker.

The pain would be the same for mobile apps that haven't updated their target sdk to the latest yet.

PR #733, opened 2022-11-10
Bump released versions

yschimke:
#### WHAT

Bump released versions

#### WHY

Test against latest versions, if we have any issues they are only getting fixed in new versions.

#### HOW

./gradlew versionCatalogUpdate

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):

yschimke:
Screenshot changes, presumably due to 

> We have updated the placeholder background drawing to merge colors rather than layer them where possible to reduce the risk of alpha blending of the different cropped layers from allowing underlying colors to bleed through at the placeholder background edges 

https://developer.android.com/jetpack/androidx/releases/wear-compose#1.1.0-rc01

PR #734, opened 2022-11-10
Use collectAsStateWithLifecycle

yschimke:
#### WHAT

Use collectAsStateWithLifecycle

#### WHY

We have a standard implement now, and the current name scares devrel.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -11,7 +11,7 @@ androidx-wear-watchface = "1.2.0-alpha03"
 androidxActivity = "1.5.1"
 androidxCore = "1.8.0"
 androidxFragment = "1.5.3"
-androidxLifecycle = "2.5.1"
+androidxLifecycle = "2.6.0-alpha03"
 androidxNavigation = "2.5.2"
 androidxStartup = "1.1.1"
 androidxTracing = "1.1.0"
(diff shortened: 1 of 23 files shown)


gradle/libs.versions.toml
```diff
@@ -11,7 +11,7 @@ androidx-wear-watchface = "1.2.0-alpha04"
 androidxActivity = "1.6.1"
 androidxCore = "1.9.0"
 androidxFragment = "1.5.4"
-androidxLifecycle = "2.5.1"
+androidxLifecycle = "2.6.0-alpha03"
```
ataulm, line 14: What's the impact of depending on an alpha version here for consumers of Horologist? Will such projects automatically use this version even if they directly depended on an older version?

yschimke: Yep - I'm torn.  I was staying on releases, but we do use some alphas,betas,rc, so it's not clear it's wrong.  Alphas are high quality.

astamato (approved):
LGTM if you're ok with the alpha dependency

ataulm (approved):
:+1: same as Ale - I think it's fine but I wonder if it's unexpected: can we be explicit about what our strategy is going forward so that we don't have to make such decisions on a case by case basis?

yschimke:
https://github.com/google/horologist/issues/736

PR #737, opened 2022-11-11
Rotary and haptics

Kpeved:
#### WHAT
Add 2 new modifiers for rotary input - `rotaryWithFling` and `rotaryWithScroll`. Those modifiers should be used  instead of `scrollableColumn` modifier. 

Both modifiers support haptics out of the box.

#### WHY
Rotary support improvements

#### HOW
`rotaryWithFling` modifier: 
For a proper fling support a `velocityTracker` is used for speed tracking. Once the speed goes up to some threshold and no other scroll events come - fling starts with the tracked speed. `FlingBehavior` is used for proper fling and can be changed in the parameters of the modifier. 

`rotaryWithScroll` modifier: 
An animation between scroll values was added for properly supporting Bezel rotary input. 

Haptics: 
A separate haptics implementation was added for producing haptics when scrolling happens. Haptics can be changed or switched off by reimplementing proper haptic classes

#### Checklist :clipboard:
- [x] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/navscaffold/ScrollableColumn.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/navscaffold/ScrollableColumn.kt
@@ -41,6 +41,7 @@ import kotlinx.coroutines.launch
  * }
  * ```
  */
+@Deprecated("Use .rotaryWithFling or .rotaryWithScroll instead")
 public fun Modifier.scrollableColumn(
     focusRequester: FocusRequester,
     scrollableState: ScrollableState
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
@@ -14,22 +14,6 @@
  * limitations under the License.
  */
 
-/*
- * Copyright 2022 The Android Open Source Project
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      https://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
 package com.google.android.horologist.compose.rotaryinput
 
 import android.view.HapticFeedbackConstants
@@ -79,19 +63,19 @@ public fun <T> Flow<T>.throttleLatest(timeframe: Long): Flow<T> =
 /**
  * Interface for Rotary haptic feedback
  */
-interface RotaryHapticFeedback {
-    fun performHapticFeedback(type: RotaryHapticsType)
+public interface RotaryHapticFeedback {
+    public fun performHapticFeedback(type: RotaryHapticsType)
 }
 
 /**
  * Rotary haptic types
  */
 @JvmInline
-value class RotaryHapticsType(val type: Int) {
-    companion object {
-        val ScrollTick = RotaryHapticsType(1)
-        val ScrollItemFocus = RotaryHapticsType(2)
-        val ScrollLimit = RotaryHapticsType(3)
+public value class RotaryHapticsType(private val type: Int) {
+    public companion object {
+        public val ScrollTick: RotaryHapticsType = RotaryHapticsType(1)
+        public val ScrollItemFocus: RotaryHapticsType = RotaryHapticsType(2)
+        public val ScrollLimit: RotaryHapticsType = RotaryHapticsType(3)
     }
 }
 
@@ -99,10 +83,10 @@ value class RotaryHapticsType(val type: Int) {
  * Remember rotary haptic feedback.
  */
 @Composable
-fun rememberRotaryHapticFeedback(
+public fun rememberRotaryHapticFeedback(
     throttleThresholdMs: Long = 40,
     hapticsChannel: Channel<RotaryHapticsType> = rememberHapticChannel(),
-    rotaryHaptics: RotaryHapticFeedback = rememberDefaultRotaryHapticFeedback(),
+    rotaryHaptics: RotaryHapticFeedback = rememberDefaultRotaryHapticFeedback()
 ): RotaryHapticFeedback {
     return remember(throttleThresholdMs, hapticsChannel, rotaryHaptics) {
         object : RotaryHapticFeedback {
@@ -141,35 +125,36 @@ private fun rememberHapticChannel() =
     }
 
 @Composable
-fun rememberDefaultRotaryHapticFeedback() =
+public fun rememberDefaultRotaryHapticFeedback(): RotaryHapticFeedback =
     LocalView.current.let { view -> remember { DefaultRotaryHapticFeedback(view) } }
 
 /**
  * Default Rotary implementation for [RotaryHapticFeedback]
  */
-class DefaultRotaryHapticFeedback(private val view: View) : RotaryHapticFeedback {
+public class DefaultRotaryHapticFeedback(private val view: View) : RotaryHapticFeedback {
 
     override fun performHapticFeedback(
-        hapticType: RotaryHapticsType,
+        type: RotaryHapticsType
     ) {
-
-        when (hapticType) {
+        when (type) {
             RotaryHapticsType.ScrollItemFocus -> {
                 view.performHapticFeedback(HapticFeedbackConstants.LONG_PRESS)
             }
+
             RotaryHapticsType.ScrollTick -> {
                 view.performHapticFeedback(HapticFeedbackConstants.CLOCK_TICK)
             }
+
             RotaryHapticsType.ScrollLimit -> {
                 view.performHapticFeedback(HapticFeedbackConstants.LONG_PRESS)
             }
         }
     }
 
-    companion object {
+    public companion object {
         // Hidden constants from HapticFeedbackConstants.java
-        const val ROTARY_SCROLL_TICK = 18
-        const val ROTARY_SCROLL_ITEM_FOCUS = 19
-        const val ROTARY_SCROLL_LIMIT = 20
+        public const val ROTARY_SCROLL_TICK: Int = 18
+        public const val ROTARY_SCROLL_ITEM_FOCUS: Int = 19
+        public const val ROTARY_SCROLL_LIMIT: Int = 20
     }
 }
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
@@ -160,7 +160,7 @@ public class DefaultRotaryHapticFeedback(private val view: View) : RotaryHapticF
             }
 
             RotaryHapticsType.ScrollTick -> {
-                view.performHapticFeedback(HapticFeedbackConstants.CLOCK_TICK)
+                view.performHapticFeedback(HapticFeedbackConstants.KEYBOARD_TAP)
             }
 
             RotaryHapticsType.ScrollLimit -> {
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
@@ -1,0 +1,175 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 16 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
```diff
@@ -0,0 +1,601 @@
...
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.flow.receiveAsFlow
+import kotlinx.coroutines.flow.transformLatest
+import kotlin.math.abs
+import kotlin.math.sign
+
+private const val DEBUG = false
```
yschimke, line 52: TODO to remove?

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
```diff
@@ -0,0 +1,601 @@
...
+private inline fun debugLog(generateMsg: () -> String) {
+    if (DEBUG) {
+        println("RotaryScroll: ${generateMsg()}")
+    }
+}
+
+/**
+ * A modifier which connects rotary events with scrollable.
+ * This modifier supports fling.
```
yschimke, line 61: Do we have some link to explain fling?

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
```diff
@@ -0,0 +1,601 @@
...
+    rotaryHaptics: RotaryHapticFeedback = rememberRotaryHapticFeedback()
+): Modifier = rotaryHandler(
+    rotaryScrollHandler = RotaryDefaults.rememberFlingHandler(scrollableState, flingBehavior),
+    rotaryHaptics = rotaryHaptics
+)
+    .focusRequester(focusRequester)
+    .focusable()
+
+/**
+ * A modifier which connects rotary events with scrollable.
```
yschimke, line 91: Can you expand on this? I'm not clear the difference between this and fling.

compose-layout/src/main/java/com/google/android/horologist/compose/navscaffold/ScrollableColumn.kt
```diff
@@ -41,6 +41,7 @@ import kotlinx.coroutines.launch
  * }
  * ```
  */
+@Deprecated("Use .rotaryWithFling or .rotaryWithScroll instead")
```
yschimke, line 44: If there is one that is almost equivalent to scrollableColumn, could we define a ReplaceWith?

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
```diff
@@ -0,0 +1,160 @@
...
+ * val flow = flow {
+ *     for (i in 1..30) {
+ *         delay(100)
+ *         emit(i)
+ *     }
+ * }
+ * ```
+ * With timeframe=1000 only those integers will be received: 1, 10, 20, 30 .
+ */
+public fun <T> Flow<T>.throttleLatest(timeframe: Long): Flow<T> =
```
yschimke, line 55: internal?

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
```diff
@@ -0,0 +1,160 @@
...
+ */
+public interface RotaryHapticFeedback {
+    public fun performHapticFeedback(type: RotaryHapticsType)
+}
+
+/**
+ * Rotary haptic types
+ */
+@JvmInline
+public value class RotaryHapticsType(private val type: Int) {
```
yschimke, line 74: Why not enum? No issue, just curious.

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * Remember rotary haptic feedback.
+ */
+@Composable
+public fun rememberRotaryHapticFeedback(
+    throttleThresholdMs: Long = 40,
+    hapticsChannel: Channel<RotaryHapticsType> = rememberHapticChannel(),
+    rotaryHaptics: RotaryHapticFeedback = rememberDefaultRotaryHapticFeedback()
+): RotaryHapticFeedback {
+    return remember(throttleThresholdMs, hapticsChannel, rotaryHaptics) {
```
yschimke, line 91: why remembering on throttleThresholdMs?

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
```diff
@@ -0,0 +1,601 @@
...
+                HighResRotaryScrollHandler(
+                    rotaryFlingBehaviorFactory = { rotaryFlingBehavior() },
+                    scrollBehaviourFactory = { scrollBehavior() }
+                )
+            }
+        }
+    }
+
+    @Composable
+    private fun isLowResInput(): Boolean = LocalContext.current.packageManager
```
yschimke, line 189: Do we need to call this more than once? Once per scrollable container, seems ok, just wondering.

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
```diff
@@ -0,0 +1,601 @@
...
+     * @param flingBehavior Logic describing Fling behaviour. If null - fling will not happen
+     * @param isLowRes Whether the input is Low-res (a bezel) or high-res(a crown/rsb)
+     */
+    @Composable
+    public fun rememberFlingHandler(
+        scrollableState: ScrollableState,
+        flingBehavior: FlingBehavior? = null,
+        isLowRes: Boolean = isLowResInput()
+    ): RotaryScrollHandler {
+        val viewConfiguration = ViewConfiguration.get(LocalContext.current)
```
yschimke, line 161: Does this benefit from being in a Composable? Seems similar to pattern of resources.

```
@Composable
@ReadOnlyComposable
internal fun nativeViewConfiguration(): ViewConfiguration {
  ViewConfiguration.get(LocalContext.current)
}
```

I'm out of my depth, but just wondering.

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
```diff
@@ -0,0 +1,601 @@
...
+                    viewConfiguration
+                )
+            }
+
+            fun scrollBehavior() = AnimationScrollBehavior(scrollableState)
+
+            if (isLowRes) {
+                LowResRotaryScrollHandler(
+                    rotaryFlingBehaviorFactory = { rotaryFlingBehavior() },
+                    scrollBehaviourFactory = { scrollBehavior() }
```
yschimke, line 177: inconsistent behaviour spelling

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
yschimke: Maybe there isn't anywhere good to cache it.

yschimke:
Looks really good, but I'm out of my depth on the fling behaviour.

yschimke:
You'll have to migrate existing usages to land.

```
w: /home/runner/work/horologist/horologist/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/entity/EntityScreen.kt: (56, 14): 'scrollableColumn(FocusRequester, ScrollableState): Modifier' is deprecated. Use .rotaryWithFling or .rotaryWithScroll instead
```

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
Kpeved: Added a short description how algorithm works. 
Do you think we should add a description what a fling is?

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
Kpeved: Added explanation that this modifier doesn't support fling

compose-layout/src/main/java/com/google/android/horologist/compose/navscaffold/ScrollableColumn.kt
Kpeved: Depends on what behaviour we want to persuade - with fling or without fling. 
There will be also another modifier `rotaryWithSnap` in the next PR.

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
Kpeved: I was thinking maybe this can be a useful public method for other usecases.. But it's not related to haptics, so let it be internal

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
Kpeved: Value class is more universal as I remember. It's widely used in AndroidX and in wear compose as well

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
Kpeved: LocalContext can only be called in a composable scope, that's why it's in composable and not in `remember` section directly

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
yschimke: Maybe, maybe not.  It might just be me who only read it for wear compose.

Just found https://developer.android.com/develop/ui/views/animations/fling-animation

So maybe it's general enough we don't need to.

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
yschimke: I think what I find surprising is direct use of Android APIs outside of specific conversion composables. It seems most android APIs get wrapped with a specific composable, like resources. But your call.

Kpeved: Right, it should be in a separate android specific part if we decide to go with multiplatform library. 
Similar to Resources.android.kt. 
When rotary code will be moved to AndroidX this should be done as well.

Though I'm not sure that we should do it on a current stage. What do you think?

P.S. In your example - ViewConfiguration  also Android class, so if we want to isolate compose code from android, we should do it for ViewConfiguration as well

yschimke: Leave it then, solve when you do it in Androidx.

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
```diff
@@ -0,0 +1,178 @@
...
+         * (start or end) and can't scroll further
+         */
+        public val ScrollLimit: RotaryHapticsType = RotaryHapticsType(3)
+    }
+}
+
+/**
+ * Remember rotary haptic feedback.
+ */
+@Composable
```
yschimke, line 103: Can you add the experimental annotation.

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
```diff
@@ -0,0 +1,631 @@
...
+ *   focusRequester.requestFocus()
+ * }
+ * ```
+ * @param focusRequester Requests the focus for rotary input
+ * @param scrollableState Scrollable state which will be scrolled while receiving rotary events
+ * @param flingBehavior Logic describing fling behavior.
+ * @param rotaryHaptics Class which will handle haptic feedback
+ */
+@Suppress("ComposableModifierFactory")
+@Composable
```
yschimke, line 94: Can you add the experimental annotation to public APIs.

yschimke (approved):

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
Kpeved: You mean this annotation? 
ExperimentalHorologistComposeLayoutApi
Or something else?

yschimke: Yep, that one.

Kpeved: Done

PR #740, opened 2022-11-14
Uses built-in Android strings for base-ui

fstanis:
#### WHAT
Makes `base-ui` use [Android built-in strings](https://developer.android.com/reference/android/R.string).

#### WHY
Make translation easier.

#### HOW
By removing `res` all together from base-ui.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


luizgrp (approved):

media-ui/src/main/res/values/strings.xml
```diff
@@ -39,9 +39,9 @@
...
     <string description="Content description of the Download in Progress button, on the Playlist Download screen. It lets the user know that the playlist is currently being downloaded, and lets the user cancel the download of the playlist. [CHAR_LIMIT=NONE]" name="horologist_playlist_download_progress_button_cancel_content_description">Download in progress</string>
-    <string description="Description of the action of the Download in Progress button on the screen of a playlist that is currently being downloaded. Lets the user know that taking this action, it will remove the download. [CHAR_LIMIT=NONE]" name="horologist_playlist_download_progress_button_cancel_action_label">cancel</string>
+    <string description="Description of the action of the Download in Progress button on the screen of a playlist that is currently being downloaded. Lets the user know that taking this action, it will remove the download. [CHAR_LIMIT=NONE]" name="horologist_playlist_download_progress_button_cancel_action_label">Cancel</string>
```
laiyichin, line 42: We are soon adding canceling functionality for individual media. Shall we rephrase this to something like "playlist/media"?

laiyichin: Actually, I can do that along with adding the functionality.

luizgrp: Given they are different buttons, I think it is better to have separate content descriptions instead of reusing them just because the text is the same

base-ui/src/main/java/com/google/android/horologist/base/ui/components/AlertDialog.kt
```diff
@@ -43,8 +43,8 @@ public fun AlertDialog(
     scalingLazyListState: ScalingLazyListState,
     modifier: Modifier = Modifier,
     title: String = "",
-    okButtonContentDescription: String = stringResource(R.string.alert_dialog_ok_button_content_description),
-    cancelButtonContentDescription: String = stringResource(R.string.alert_dialog_cancel_button_content_description)
+    okButtonContentDescription: String = stringResource(R.string.ok),
```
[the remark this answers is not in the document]
fstanis: Yeah, I think using Android strings for defaults works, but we should do it sparingly.

PR #746, opened 2022-11-15
Volume Screen with Label

yschimke:
#### WHAT

Volume Screen with Label.

#### WHY

Some usages may not have a changeable output.

#### HOW

Split out an internal function, add a new API without params.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


audio-ui/src/main/res/values/strings.xml
```diff
@@ -24,4 +24,5 @@
     <string description="Indicates there's nothing to output audio to, e.g. there's no speaker and no headphones connected at the moment. Appears on the device output button chip. [CHAR LIMIT=20]" name="horologist_output_none">None</string>
     <string description="Content description of the SetVolume button. It lets the user navigate to the screen to control the system volume. [CHAR_LIMIT=NONE]" name="horologist_set_volume_content_description">Set Volume</string>
     <string description="Content description of the AudioOutput button. It lets the user navigate to the system bluetooth settings screen to connect to a headset. [CHAR_LIMIT=NONE]" name="horologist_audio_output_content_description">Audio Output</string>
+    <string description="Center label on the Volume Screen. [CHAR_LIMIT=NONE]" name="horologist_volume_screen_volume_label">Volume</string>
```
fstanis, line 27: Nit: the description should explain the context, e.g. "Label that tells the user they are in the part of the app for changing output volume" (also `CHAR_LIMIT` can't be `NONE` if it's visible in the UI)

fstanis (commented):
I'm thinking we should maybe rename the original (with `audioOutputUi`) `VolumeScreen`to something like `VolumeAudioOutputScreen` to differentiate them?

yschimke:
@fstanis I'm not keen on renaming it, I've added a brief comment to disambiguate the versions. My main objection to renaming it is more that I expect the label version is less common.  VolumeScreenWithLabel?

fstanis (approved):

PR #753, opened 2022-11-16
Add readme screenshot for square segmented progress indicator

oas004:
#### WHAT
Add square segmented progress indicator to the README collection so the user knows that it is there.

#### WHY
Since we merged #695 I think it makes sense to add it to the collection in the README.

#### HOW
Adding the image as a png in the docs and add the reference to the markdown table.


README.md
```diff
@@ -24,9 +24,9 @@ High quality prebuilt composables, such as Time and Date pickers.
...
+DatePicker             |  TimePickerWith12HourClock |  TimePicker |  SegmentedProgressIndicator |  SquareSegmentedProgressIndicator
+:-------------------------:|:-------------------------:|:-------------------------:|:-------------------------:|:-------------------------:
+<img src="https://raw.githubusercontent.com/google/horologist/main/docs/composables/date_picker.png" height="120" width="120" >  |  <img src="https://raw.githubusercontent.com/google/horologist/main/docs/composables/time_12h_picker.png" height="120" width="120"> | <img src="https://raw.githubusercontent.com/google/horologist/main/docs/composables/time_24h_picker.png" height="120" width="120"> | <img src="https://raw.githubusercontent.com/google/horologist/main/docs/composables/segmented_progress_indicator.png" height="120" width="120"> | <img src="https://raw.githubusercontent.com/google/horologist/main/docs/composables/square_segmented_progress_indicator.png" height="120" width="120">
```
oas004, line 29: Will this actually work? Or is it just not showing because the png is not in main yet for it to be on raw.githubusercontent ? 🤔

oas004: Is it also a bit strange that it is on the side and you then have to scroll to see it? Maybe it is better to have a new row or something? Wdyt? :)

yschimke: Exactly, proof is https://raw.githubusercontent.com/google/horologist/52e2912f8d9fc46fe823e90a7d04451fd0a799dd/docs/composables/square_segmented_progress_indicator.png

yschimke: +1, split the progress indicators into a separate line

oas004: Sweet! Then it should work :)

yschimke (approved):

PR #754, opened 2022-11-17
Add FromDatePicker to horologist

rajat4064g:
#### WHAT
Added `FromDatePicker` to `composables`.

#### WHY
[DatePicker](https://github.com/google/horologist/blob/main/composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt) doesn't support picking a date only future to a particular date. `FromDatePicker` restricts the user to pick a date before `fromDate`.

#### HOW
Added the parameter `fromDate` along with parameters of `DatePicker` in `FromDatePicker`. Then added appropriate offsets wherever required.


composables/src/main/java/com/google/android/horologist/composables/FromDatePicker.kt
```diff
@@ -0,0 +1,295 @@
...
+ * ensure that it will draw properly on smaller screens it does not take account of user font size
+ * overrides for MaterialTheme.typography.display2 which is used to display the main picker value.
+ *
+ * @param onDateConfirm the button event handler.
+ * @param modifier the modifiers for the `Box` containing the UI elements.
+ * @param date the initial value to seed the picker with.
+ * @param fromDate the minimum date to be selected in picker
+ */
+@Composable
+public fun FromDatePicker(
```
yschimke, line 69: I'd suggest we consider designs that allow some restriction of dates, rather than just forward.

@BowerSteve any thoughts?

yschimke:
Could you give some context of how you will use this? Calendar/Todo app?  Understanding your requirements will help for the review.

rajat4064g:
> Could you give some context of how you will use this? Calendar/Todo app? Understanding your requirements will help for the review.

This could be used in `Countdown to date` complication which is part of wear clock

composables/src/main/java/com/google/android/horologist/composables/FromDatePicker.kt
BowerSteve: Yes, I think adding optional from and to dates would make sense here. And also, I think those could be added to DatePicker itself, rather than spinning off a separate component, as the parameters are optional anyway.

rajat4064g: Created a new PR for optional `fromDate` and `toDate` parameters in DatePicker - https://github.com/google/horologist/pull/791

rajat4064g: Closing this PR

PR #757, opened 2022-11-17
Rotary improvements

Kpeved:
#### WHAT
A rotary was improved by fixing some race conditions on high-res devices

#### WHY
When RSB was scrolled too fast some rotary events were skipped

#### HOW
Snapping: 
Instead of relying on separate flag which was set manually, now we rely on `snapJob.isActive` to clearly distinguish whether the snapping is in progress or not. 
If a new snapping happens when `snapJob.isActive == true` old job is not killed - but rather updated. This helps us to speed up things.

Scrolling :
Method `handleEvent` now accepts `targetValue` which is the absolute number of pixels by which the scroll should happen. Before it was a cumulative sum - every new call would sum up all values. 
Example : 
 - Calling `handleEvent` with  50 , 60, 100, 20 
Before :  would result in  230 px scrolled
Now: would result in 20px scrolled 

Tests will be added in a separate CL 

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke (approved):
Not sure I'm the right person to review in detail.  

Ideally if we have such complicated code, it might be nice to have a test. But I'll leave it to your judgement when it's worth doing that?

PR #758, opened 2022-11-18
Add auth modules

luizgrp:
#### WHAT

Add auth modules.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Nice, selection of modules makes sense.

PR #760, opened 2022-11-18
Add OAuth PCKE implementation

luizgrp:
#### WHAT

Add [OAuth PCKE](https://developer.android.com/training/wearables/apps/auth-wear#pkce) implementation adapted from [wear-os-samples](https://github.com/android/wear-os-samples/tree/cf28cb3b042ee00ecf1313f067f39c2add243ee7/WearOAuth).

#### HOW

Implementation is coordinated by `AuthPKCEViewModel`. 
The VM needs 3 repositories to perform the auth steps:
- `AuthPKCEConfigRepository` - to get the configurations needed to fetch the necessary info;
- `AuthPKCEOAuthCodeRepository` - to fetch oauth code;
- `AuthPKCETokenRepository` - to fetch the tokens (access/refresh);

The VM is also accepts a `AuthPKCETokenPayloadListener` in case clients are using the default implementations supplied and would like to persist the token themselves.

The implementation of the components above relies on heavy usage of generics in order to allow customisation of each step.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth-data/src/main/java/com/google/android/horologist/auth/data/pkce/impl/google/AuthPKCETokenRepositoryGoogleImpl.kt
+++ b/auth-data/src/main/java/com/google/android/horologist/auth/data/pkce/impl/google/AuthPKCETokenRepositoryGoogleImpl.kt
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.auth.data.pkce.impl.google
+
+import android.util.Log
+import com.google.android.horologist.auth.data.ExperimentalHorologistAuthDataApi
+import com.google.android.horologist.auth.data.pkce.AuthPKCETokenRepository
+import com.google.android.horologist.auth.data.pkce.impl.AuthPKCEDefaultConfig
+import com.google.android.horologist.auth.data.pkce.impl.AuthPKCEOAuthCodeDefaultPayload
+import com.google.android.horologist.auth.data.pkce.impl.google.api.GoogleOAuthService
+import com.google.android.horologist.auth.data.pkce.impl.google.api.GoogleOAuthService.Companion.REQUEST_GRANT_TYPE_PARAM_VALUE
+import com.google.android.horologist.auth.data.pkce.impl.google.api.TokenResponse
+import kotlinx.coroutines.CancellationException
+
+@ExperimentalHorologistAuthDataApi
+public class AuthPKCETokenRepositoryGoogleImpl(
+    private val googleOAuthService: GoogleOAuthService
+) : AuthPKCETokenRepository<AuthPKCEDefaultConfig, AuthPKCEOAuthCodeDefaultPayload, TokenResponse> {
+
+    override suspend fun fetch(
+        config: AuthPKCEDefaultConfig,
+        codeVerifier: String,
+        oAuthCodePayload: AuthPKCEOAuthCodeDefaultPayload
+    ): Result<TokenResponse> {
+        Log.d(TAG, "Requesting token...")
+
+        return try {
+            val response = googleOAuthService.token(
+                clientId = config.clientId,
+                clientSecret = config.clientSecret,
+                code = oAuthCodePayload.code,
+                codeVerifier = codeVerifier,
+                grantType = REQUEST_GRANT_TYPE_PARAM_VALUE,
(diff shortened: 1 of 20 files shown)


auth-ui/src/main/java/com/google/android/horologist/auth/ui/pkce/AuthPKCEViewModel.kt
```diff
@@ -0,0 +1,90 @@
...
+import com.google.android.horologist.auth.data.pkce.AuthPKCETokenRepository
+import com.google.android.horologist.auth.ui.ExperimentalHorologistAuthUiApi
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.SharingStarted
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+
+@ExperimentalHorologistAuthUiApi
+public open class AuthPKCEViewModel<AuthPKCEConfig, OAuthCodePayload, TokenPayload>(
```
yschimke, line 36: Do any elements of these need to be saved to savedStateHandle?

yschimke: Can app stop during the CheckPhone stage?

yschimke, line 36: Is idea that app extends and hooks into Hilt or creates a factory?

auth-data/src/main/java/com/google/android/horologist/auth/data/pkce/impl/google/api/TokenResponse.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.squareup.moshi.Json
+import com.squareup.moshi.JsonClass
+
+@JsonClass(generateAdapter = true)
+public data class TokenResponse(
+    @Json(name = "access_token") val accessToken: String,
+    @Json(name = "expires_in") val expiresIn: Int,
+    @Json(name = "id_token") val idToken: String?,
+    @Json(name = "refresh_token") val refreshToken: String?,
+    @Json(name = "scope") val scope: String?,
```
yschimke, line 28: Are these actually nullable?

auth-data/src/main/java/com/google/android/horologist/auth/data/pkce/impl/google/AuthPKCETokenRepositoryGoogleImpl.kt
```diff
@@ -0,0 +1,63 @@
...
+        } catch (e: CancellationException) {
+            throw e
+        } catch (e: Exception) {
+            e.printStackTrace()
+            Result.failure(e)
+        }
+    }
+
+    private companion object {
+        private val TAG = AuthPKCETokenRepositoryGoogleImpl::class.java.simpleName
```
yschimke, line 61: Seems like a long tag.

yschimke (approved):
LGTM - for inprogress development

auth-ui/src/main/java/com/google/android/horologist/auth/ui/pkce/AuthPKCEViewModel.kt
luizgrp: > Do any elements of these need to be saved to savedStateHandle?

it would be good to save the CheckPhone state + OAuthCodePayload (would have to change this generic to parcelable or serializable), Success and Failed states

> Can app stop during the CheckPhone stage?

Not sure I understand it, could the app be killed? The phone or the wear app? yes, for both cases

luizgrp: I'd expect the former, but if the latter works for the user, that's fine too. Do you see any issue?
I have plans to explore the possibility of adding builders for the UI components and I might make use of a factory for this VM, I will see.

auth-data/src/main/java/com/google/android/horologist/auth/data/pkce/impl/google/api/TokenResponse.kt
luizgrp: If I'm not mistaken, `refresh_token` came with null value when I was testing it last week, hence added nullability to those. It's not clear in the [documentation](https://developers.google.com/identity/protocols/oauth2/native-app#exchange-authorization-code). 

`id_token` seems to be:

> This property is only returned if your request included an identity scope

auth-data/src/main/java/com/google/android/horologist/auth/data/pkce/impl/google/AuthPKCETokenRepositoryGoogleImpl.kt
luizgrp: suggestions? single small tag "auth" to be used across all classes, even across different modules?

luizgrp:
merging this and will address improvements in follow up prs

auth-data/src/main/java/com/google/android/horologist/auth/data/pkce/impl/google/AuthPKCETokenRepositoryGoogleImpl.kt
luizgrp: https://github.com/google/horologist/pull/860

PR #765, opened 2022-11-22
Add OAuth Device Grant implementation

luizgrp:
#### WHAT

Add [OAuth Device Grant](https://developer.android.com/training/wearables/apps/auth-wear#DAG) implementation adapted from [wear-os-samples](https://github.com/android/wear-os-samples/tree/cf28cb3b042ee00ecf1313f067f39c2add243ee7/WearOAuth).

#### HOW

Implementation is coordinated by `AuthDeviceGrantViewModel`. 
The VM needs 3 repositories to perform the auth steps:
- `AuthDeviceGrantConfigRepository` - to get the configurations needed to fetch the necessary info;
- `AuthDeviceGrantVerificationInfoRepository` - to fetch the uri and user code;
- `AuthDeviceGrantTokenRepository` - to fetch the tokens (access/refresh);

The VM is also accepts a `AuthDeviceGrantTokenPayloadListener` in case clients are using the default implementations supplied and would like to persist the token themselves.

The implementation of the components above relies on heavy usage of generics in order to allow customisation of each step.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


auth-data/src/main/java/com/google/android/horologist/auth/data/oauth/common/impl/google/api/GoogleOAuthServiceFactory.kt
```diff
@@ -0,0 +1,47 @@
...
+import retrofit2.Retrofit
+import retrofit2.converter.moshi.MoshiConverterFactory
+
+@ExperimentalHorologistAuthDataApi
+public class GoogleOAuthServiceFactory(
+    private val moshi: Moshi
+) {
+
+    public fun get(): GoogleOAuthService {
+        val httpClient = OkHttpClient.Builder()
```
yschimke, line 32: Consider passing this as a dependency, apps should ideally have a small number(1, maybe 2 or 3) of OkHttpClient instances, and the logging interceptor is not specific to this factory.

auth-data/src/main/java/com/google/android/horologist/auth/data/oauth/devicegrant/impl/google/AuthDeviceGrantTokenRepositoryGoogleImpl.kt
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Poll the Auth server for the token. This will only return when the user has finished their
+     * authorization flow on the paired device.
+     *
+     * For this sample the various exceptions aren't handled.
+     */
+    private tailrec suspend fun retrieveToken(
+        config: AuthDeviceGrantDefaultConfig,
+        deviceCode: String,
+        interval: Int
```
yschimke, line 69: Is it simpler to pass duration here? callers can pass 1.seconds.

auth-ui/src/main/java/com/google/android/horologist/auth/ui/oauth/devicegrant/AuthDeviceGrantViewModel.kt
```diff
@@ -0,0 +1,86 @@
...
+    private val authDeviceGrantTokenRepository: AuthDeviceGrantTokenRepository<AuthDeviceGrantConfig, VerificationInfoPayload, TokenPayload>,
+    private val checkPhonePayloadMapper: (AuthDeviceGrantConfig, VerificationInfoPayload) -> String,
+    private val authDeviceGrantTokenPayloadListener: AuthDeviceGrantTokenPayloadListener<TokenPayload> = AuthDeviceGrantTokenPayloadListenerNoOpImpl()
+) : ViewModel() {
+
+    private val _uiState =
+        MutableStateFlow<AuthDeviceGrantScreenState>(AuthDeviceGrantScreenState.Idle)
+    public val uiState: StateFlow<AuthDeviceGrantScreenState> = _uiState.stateIn(
+        scope = viewModelScope,
+        started = SharingStarted.WhileSubscribed(stopTimeoutMillis = 5_000),
```
yschimke, line 46: Not sure what the right behaviour is, but WhileSubscribed and MutableStateFlow seem like a weird combo.  I'm sure I've done this before.  But I consider the WhileSubscribe(5) a way to keep subscribed while you temporarily navigate somewhere else to avoid tearing down subscriptions.

Not sure of a better suggestion, so mainly for discussion.

yschimke (approved):

auth-data/src/main/java/com/google/android/horologist/auth/data/oauth/devicegrant/impl/google/AuthDeviceGrantTokenRepositoryGoogleImpl.kt
luizgrp: Not sure in this case, the value for this param comes from the server [response](https://github.com/google/horologist/pull/765/files#diff-544e96c8ebec3537bb4b6e25fa21b1717016073e7ce3b57b8c8619815748d956R27) serialized as Int.

This code is exact copy from wear-os-samples, will spend some time later iterating on it and will see if this can be addressed.

auth-ui/src/main/java/com/google/android/horologist/auth/ui/oauth/devicegrant/AuthDeviceGrantViewModel.kt
luizgrp: cool, I see it's like this across the library, so I'd suggest to raise an issue and we could change everywhere in one go with a better approach

luizgrp: https://github.com/google/horologist/pull/910

PR #766, opened 2022-11-22
Replace BluetoothFragment explicit intent with androidx.mediarouter call

wrkben:
Androidx.mediarouter now supports the invocation of the Wear Bluetooth fragment. This PR changes the call to the Wear Bluetooth fragment to use [SystemOutputSwitcherDialogController](https://developer.android.com/reference/androidx/mediarouter/app/SystemOutputSwitcherDialogController) instead.

The new API always use `closeOnConnect` as true so this change also deprecates this boolean parameter for the API to be honest about what it does.

### DIFF
--- a/audio/src/main/java/com/google/android/horologist/audio/AudioOutputRepository.kt
+++ b/audio/src/main/java/com/google/android/horologist/audio/AudioOutputRepository.kt
@@ -35,5 +35,14 @@ public interface AudioOutputRepository : AutoCloseable {
     /**
      * Action to launch output selection by the user.
      */
+    @Deprecated(
+        "closeOnConnect is always true, use #launchOutputSelection() instead",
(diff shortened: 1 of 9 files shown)


audio/src/main/java/com/google/android/horologist/audio/AudioOutputRepository.kt
```diff
@@ -35,5 +35,11 @@ public interface AudioOutputRepository : AutoCloseable {
     /**
      * Action to launch output selection by the user.
      */
+    @Deprecated(message = "closeOnConnect is always true, use #launchOutputSelection() instead")
```
yschimke, line 38: nit: Consider using ReplaceWith to fix the call sites.

yschimke (approved):

PR #767, opened 2022-11-22
Add Auth Device Grant to sample app

luizgrp:
#### WHAT

Add Auth Device Grant to sample app

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/sample/src/main/java/com/google/android/horologist/auth/oauth/devicegrant/AuthDeviceGrantScreen.kt
+++ b/sample/src/main/java/com/google/android/horologist/auth/oauth/devicegrant/AuthDeviceGrantScreen.kt
@@ -16,23 +16,50 @@
 
 package com.google.android.horologist.auth.oauth.devicegrant
 
-import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.runtime.Composable
-import androidx.compose.ui.Alignment
+import androidx.compose.runtime.getValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.text.style.TextAlign
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.lifecycle.viewmodel.compose.viewModel
 import androidx.wear.compose.material.Text
+import com.google.android.horologist.auth.ui.oauth.devicegrant.AuthDeviceGrantScreenState
 
 @Composable
-fun AuthDeviceGrantScreen() {
-    Box(modifier = Modifier.fillMaxSize()) {
+fun AuthDeviceGrantScreen(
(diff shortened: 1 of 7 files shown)


sample/src/main/java/com/google/android/horologist/auth/oauth/devicegrant/AuthDeviceGrantScreen.kt
```diff
@@ -16,23 +16,50 @@
...
 @Composable
-fun AuthDeviceGrantScreen() {
-    Box(modifier = Modifier.fillMaxSize()) {
+fun AuthDeviceGrantScreen(
+    modifier: Modifier = Modifier,
+    viewModel: AuthDeviceGrantScreenViewModel = viewModel(factory = AuthDeviceGrantScreenViewModel.Factory)
+) {
+    val state by viewModel.uiState.collectAsStateWithLifecycle()
+
+    if (state == AuthDeviceGrantScreenState.Idle) {
```
yschimke, line 39: Should this be in a SideEffect? or LaunchedEffect?

yschimke (approved):

sample/src/main/java/com/google/android/horologist/auth/oauth/devicegrant/AuthDeviceGrantScreen.kt
luizgrp: I would need to learn more about it, happy to apply your suggestion if you'd know what's more appropriate.

Based on the docs, [SideEffect](https://developer.android.com/jetpack/compose/side-effects#sideeffect-publish) doesn't seem appropriated as it shouldn't be invoked in every recomposition. 

I'm not sure about [LaunchedEffect](https://developer.android.com/jetpack/compose/side-effects#launchedeffect) either because the call `viewModel.startAuthFlow` is not a suspend function. The ViewModel launches a coroutine with its scope there, not sure this should be with this compose scope. Now maybe if this composable recomposes while `state` is `Idle` we might have an unwanted side effect, which is more than one call to `viewModel.startAuthFlow`. So maybe we could use a `LaunchedEffect` with a remembered boolean flag, to make sure this is only called once? Let me know your thoughts.

yschimke: If you don't have SideEffect, then I believe it's running on every recomposition, but crucially within a time sensitive part of the UI work. I think SideEffect would be a good idea to just delay it until after the composition succeeds.

https://developer.android.com/jetpack/compose/side-effects#state-effect-use-cases

> As covered in the [Thinking in Compose](https://developer.android.com/jetpack/compose/mental-model) documentation, composables should be side-effect free. When you need to make changes to the state of the app (as described in the [Managing state documentation](https://developer.android.com/jetpack/compose/state) doc), you should use the Effect APIs so that those side effects are executed in a predictable manner.

luizgrp: https://github.com/google/horologist/pull/771

PR #768, opened 2022-11-22
Add locale previews

yschimke:
#### WHAT

Add locale previews.

#### WHY

No urgency, but it felt like a gap of our translations.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
nice!

PR #769, opened 2022-11-22
Add Google Sign-In implementation and sample

luizgrp:
#### WHAT

Add [Google Sign-In](https://developer.android.com/training/wearables/apps/auth-wear#Google-Sign-in) implementation adapted from [wear-os-samples](https://github.com/android/wear-os-samples/tree/cf28cb3b042ee00ecf1313f067f39c2add243ee7/WearStandaloneGoogleSignIn).

#### HOW

Implementation is done in `GoogleSignInScreen` with `GoogleSignInViewModel` coordinating the screen states and allowing a listener to the account (`AuthGoogleSignInAccountListener`) to be hooked into it.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth-data/build.gradle
+++ b/auth-data/build.gradle
@@ -92,6 +92,7 @@ dependencies {
     implementation libs.androidx.wear
     implementation libs.androidx.wear.phone.interactions
     implementation libs.androidx.wear.remote.interactions
+    implementation libs.playservices.auth
     implementation libs.retrofit2.retrofit
     implementation libs.retrofit2.convertermoshi
     implementation libs.moshi.adapters
--- a/auth-ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/GoogleSignInScreen.kt
+++ b/auth-ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/GoogleSignInScreen.kt
@@ -1,0 +1,161 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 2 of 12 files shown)


auth-data/build.gradle
```diff
@@ -92,6 +92,7 @@ dependencies {
     implementation libs.androidx.wear
     implementation libs.androidx.wear.phone.interactions
     implementation libs.androidx.wear.remote.interactions
+    implementation libs.playservices.auth
```
yschimke, line 95: flagging that this will become a borg dependency, dragging in all possible libraries :)

We should keep in mind how to modularise in future.

luizgrp: yeah I might move this flow to its own module, didn't want to overcomplicate it for now

auth-ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/GoogleSignInScreen.kt
```diff
@@ -0,0 +1,161 @@
...
+
+        return if (task.isSuccessful) {
+            task.result
+        } else {
+            val exception = task.exception
+            check(exception is ApiException)
+            val message = GoogleSignInStatusCodes.getStatusCodeString(exception.statusCode)
+            Log.w(TAG, "Sign in failed: code=${exception.statusCode}, message=$message")
+
+            onTaskFailed()
```
yschimke, line 152: Is it worth providing some indication to the caller as to the reason?

auth-ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/GoogleSignInScreen.kt
```diff
@@ -0,0 +1,161 @@
...
+                    viewModel.startAuthFlow()
+                }
+            }
+        }
+
+        AuthGoogleSignInScreenState.SelectAccount -> {
+            val context = LocalContext.current
+
+            var googleSignInAccount by remember {
+                mutableStateOf(GoogleSignIn.getLastSignedInAccount(context))
```
yschimke, line 72: IIRC it's possible to already have a signed in account? Should this be a feature of the auth flow, that we might check for a mechanism that support automatic/existing signin and default to that?

yschimke (approved):

auth-ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/GoogleSignInScreen.kt
luizgrp: yes and yes! it's planned for the sign in prompt screen

PR #774, opened 2022-11-23
fix #729

### LINKED ISSUE #729: Add support for MediaPosition with unset position
media3 has a sentinel for unset position: [`C.POSITION_UNSET`](https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/C.html#POSITION_UNSET). However, our code doesn't take this into account, so since it's negative it would likely crash here:

https://github.com/google/horologist/blob/2528ad0c4f8a048015acbbcac6673c6f410bbac8/media/src/main/java/com/google/android/horologist/media/model/MediaPosition.kt#L44-L46

We could either treat this as if position is 0, but a more correct solution may be to add `MediaPosition.UnknownPosition` and `MediaPosition.Unknown` to handle this case.

kul3r4:
#### WHAT
Avoid launching exception when position or durations are unknowns

#### WHY
While unknown duration is rare, unknown position can happen for live streaming

#### HOW
add a new object Unknown for Media position which represent all the cases where a percentage cannot be calculated

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


media/src/main/java/com/google/android/horologist/media/model/MediaPosition.kt
```diff
@@ -24,37 +24,33 @@ import kotlin.time.Duration
...
-                "Current position can't be a negative value [current: $current] [duration: $duration]."
-            }
-            check(duration.isPositive()) {
-                "Duration has to be greater than zero [current: $current] [duration: $duration]."
-            }
-            check(current <= duration) {
-                "Duration can't be less than current position [current: $current] [duration: $duration]."
+        ): MediaPosition {
+            if (!current.isPositive() || !duration.isPositive() || current >= duration) {
+                return Unknown
```
fstanis, line 43: We should perhaps keep the exception and keep the return value `KnownDuration` - the idea behind, if it reaches this point, then something probably went wrong in `MediaPositionMapper`.

media-data/src/main/java/com/google/android/horologist/media/data/mapper/MediaPositionMapper.kt
```diff
@@ -26,12 +26,16 @@ import kotlin.time.Duration.Companion.milliseconds
...
+
 public object MediaPositionMapper {
     public fun map(player: Player?): MediaPosition? {
         return if (player == null || player.currentMediaItem == null) {
             null
-        } else if (player.duration == C.TIME_UNSET || player.duration <= 0L) {
-            MediaPosition.UnknownDuration(player.currentPosition.milliseconds)
+        } else if (player.duration == C.TIME_UNSET ||
+            player.duration <= 0L ||
+            player.currentPosition <= 0L
```
fstanis, line 36: Also add `player.currentPosition == C.POSITION_UNSET` to match the `player.duration == C.TIME_UNSET` condition (even though both are technically redundant because those constants are `<= 0`, but it's better not to depend on that).

media/src/main/java/com/google/android/horologist/media/model/MediaPosition.kt
```diff
@@ -24,37 +24,33 @@ import kotlin.time.Duration
...
             }
 
             val percent =
                 current.inWholeMilliseconds.toFloat() / duration.inWholeMilliseconds.toFloat()
 
             return KnownDuration(current, duration, percent)
         }
+
+        public fun createUnknown(): Unknown {
```
fstanis, line 52: I'd maybe get rid of this and just use `MediaPosition.Unknown` directly above, the other `create` is sort of needed for validation.

media/src/test/java/com/google/android/horologist/media/model/MediaPositionTest.kt
```diff
@@ -37,45 +36,48 @@ class MediaPositionTest {
         val result = MediaPosition.create(current = current, duration = duration)
 
         // then
-        assertThat(result.current).isEqualTo(current)
-        assertThat(result.duration).isEqualTo(duration)
-        assertThat(result.percent).isEqualTo(0.5f)
+        assertThat(result is MediaPosition.KnownDuration).isTrue()
```
fstanis, line 39: Nit: `assertThat(result).isInstanceOf(MediaPosition.KnownDuration::class.java)` should give better errors on test failure (same in code below).

fstanis (approved):
LGTM. We should also follow-up with a CL that hides the progress bar all together when position is `MediaPosition.Unknown`. This could be something like `showProgress: Boolean` added into `TrackPositionUiModel` and then surfaced in the UI.

media-data/src/main/java/com/google/android/horologist/media/data/mapper/MediaPositionMapper.kt
kul3r4: exactly, I think they seemed redundant to me. When I added `C.POSITION_UNSET` I got:

> Error: This declaration is opt-in and its usage should be marked with @androidx.media3.common.util.UnstableApi or @OptIn(markerClass = androidx.media3.common.util.UnstableApi.class) [UnsafeOptInUsageError from androidx.annotation.experimental]
>             player.currentPosition == C.POSITION_UNSET.toLong() ||

even if I was adding `@OptIn(UnstableApi::class)`, any advise?

fstanis: Try adding at the top of the file:

```
@file:OptIn(UnstableApi::class)

package com.google.android.horologist.media.data.mapper

import androidx.annotation.OptIn
import androidx.media3.common.util.UnstableApi
```

yschimke:
Can you update the title to a standalone description? "Add support for MediaPosition with unset position" ?

media-data/src/main/java/com/google/android/horologist/media/data/mapper/MediaPositionMapper.kt
fstanis: Actually, looking at it, seems `C.POSITION_UNSET` can't be used here - it's an int and not a long. It's not super clear what an unknown position would be.

Let's just keep it as `player.duration == C.TIME_UNSET || player.duration <= 0L || player.currentPosition < 0` like you originally put it (but < instead of <= for position since position 0 is valid).

kul3r4: done

fstanis (approved):

luizgrp:
@kul3r4 @fstanis before merging this, could we please make sure that the gradle task `:media:test` is passing? this task unfortunately is not running in CI, I will raise an issue now

PR #776, opened 2022-11-24
Add Google Sign-Out functionality to sample app

luizgrp:
#### WHAT

Add Google Sign-Out functionality to sample app

#### WHY

In order to allow to sign out so the sign-in flow can be tested multiples times.

#### HOW

Add the logic into AuthGoogleSignOutScreen / AuthGoogleSignOutViewModel, display a success confirmation message when successfully or fail silently for now.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/sample/src/main/java/com/google/android/horologist/auth/googlesignin/GoogleSignOutScreen.kt
+++ b/sample/src/main/java/com/google/android/horologist/auth/googlesignin/GoogleSignOutScreen.kt
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.auth.googlesignin
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.SideEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.lifecycle.viewmodel.compose.viewModel
+import androidx.navigation.NavHostController
+import androidx.wear.compose.material.Text
+import androidx.wear.compose.material.dialog.Confirmation
+import com.google.android.horologist.sample.R
+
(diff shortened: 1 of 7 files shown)


sample/src/main/java/com/google/android/horologist/auth/googlesignin/GoogleSignOutViewModel.kt
```diff
@@ -0,0 +1,82 @@
...
+) : ViewModel() {
+
+    private val _uiState = MutableStateFlow(GoogleSignOutScreenState.Idle)
+    public val uiState: StateFlow<GoogleSignOutScreenState> = _uiState.stateIn(
+        scope = viewModelScope,
+        started = SharingStarted.WhileSubscribed(stopTimeoutMillis = 5_000),
+        initialValue = GoogleSignOutScreenState.Idle
+    )
+
+    init {
```
yschimke, line 47: I know this is sample code, but do we have clear guidance for when to use `init{}` vs a SideEffect?

```
    when (state) {
        AuthGoogleSignInScreenState.Idle -> {
            SideEffect {
                if (!executedOnce) {
                    executedOnce = true
                    viewModel.startAuthFlow()
                }
            }
        }
```

sample/src/main/java/com/google/android/horologist/auth/googlesignin/GoogleSignOutScreen.kt
```diff
@@ -0,0 +1,63 @@
...
+                    modifier = Modifier.align(Alignment.CenterHorizontally),
+                    textAlign = TextAlign.Center,
+                    text = stringResource(id = R.string.google_sign_out_success_message)
+                )
+            }
+        }
+
+        GoogleSignOutScreenState.Failed -> {
+            SideEffect {
+                navController.popBackStack()
```
yschimke, line 59: Should we show the error somehow? Snackbar?

yschimke (approved):

sample/src/main/java/com/google/android/horologist/auth/googlesignin/GoogleSignOutScreen.kt
luizgrp: could do, chosen to leave it for now as per pr description - it's logging the exception in logcat in the VM

sample/src/main/java/com/google/android/horologist/auth/googlesignin/GoogleSignOutViewModel.kt
luizgrp: Not sure about the guidance, I'd personally don't like to have a VM with side effects in its constructor. 

In this case the call to sign out could have been done in the caller screen, it wouldn't need a screen and vm just for this. The only reason I added like this was to have all the code contained in a separate class inside a "googlesignin" package, thus avoiding the caller class `AuthMenuScreen` to become a god class.

Any strong opinions about this? I will iterate back on this, just needed this functionality for now and didn't spend much time in designing it.

yschimke: No, mainly observing a mix in these recent PRs.

luizgrp: cool, will probably remove this init once improving the sign out and address the failure case from the comment below

PR #778, opened 2022-11-24
Add file optin for AlertDialogAlertPreview

oas004:
#### WHAT
Add optIn for entire file in AlertDialogAlertPreview instead of opting in on each function in the file.

#### WHY
Following up on https://github.com/google/horologist/pull/775#discussion_r1031685145

#### HOW
Adding the optin for the file and removing the annotations on the functions

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
thanks!

PR #779, opened 2022-11-24
Add Auth composables

luizgrp:
#### WHAT

Add prototype versions of the following Auth composables:
- CheckYourPhoneScreen
- SignedInConfirmationDialog
- AuthErrorScreen

Update samples to use them.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth-composables/src/main/java/com/google/android/horologist/auth/composables/dialog/SignedInConfirmationDialog.kt
+++ b/auth-composables/src/main/java/com/google/android/horologist/auth/composables/dialog/SignedInConfirmationDialog.kt
@@ -1,0 +1,156 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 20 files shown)


auth-composables/src/main/java/com/google/android/horologist/auth/composables/dialog/SignedInConfirmationDialog.kt
```diff
@@ -0,0 +1,156 @@
...
+        onDismissRequest = currentOnDismissOrTimeout,
+        modifier = modifier
+    ) {
+        Column(
+            modifier = modifier.fillMaxSize(),
+            verticalArrangement = Arrangement.Center
+
+        ) {
+            Text(
+                text = "Success!",
```
yschimke, line 149: literal content string.

yschimke (approved):

auth-composables/src/main/java/com/google/android/horologist/auth/composables/dialog/SignedInConfirmationDialog.kt
luizgrp: thanks for flagging it! this particular one I left intentionally as it is throw away code and will it remove in the next pr, I added to the resource file the strings of the other composables that are more aligned with the designs

luizgrp: https://github.com/google/horologist/pull/796

PR #780, opened 2022-11-25
Workaround placeholder inconsistency with PagerItemCard

yschimke:
#### WHAT

Workaround issue with placeholders.
Add a preview of PagingItem with placeholder.
Also removes a CircularProgressIndicator that does nothing.

#### WHY

Repro of a placeholder bug

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

image

### DIFF
--- a/sample/src/main/java/com/google/android/horologist/paging/PagingScreen.kt
+++ b/sample/src/main/java/com/google/android/horologist/paging/PagingScreen.kt
@@ -16,7 +16,7 @@
 
 package com.google.android.horologist.paging
 
-import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.runtime.Composable
@@ -97,7 +97,9 @@ private fun PagingItemCard(
     modifier: Modifier = Modifier,
     onClick: () -> Unit = {}
 ) {
-    val chipPlaceholderState = rememberPlaceholderState { item != null }
+    // Workaround for https://issuetracker.google.com/issues/260343754
+    val chipPlaceholderState =
+        if (item != null) rememberPlaceholderState { true } else rememberPlaceholderState { false }
 
     TitleCard(
         modifier = modifier
@@ -165,7 +167,6 @@ private class MyBackend {
         }
 
         val itemsAfter = backendDataList.size - to - 1
-        println(itemsAfter)
         return DesiredLoadResultPageResponse(data = currentSublist, itemsAfter = itemsAfter)
     }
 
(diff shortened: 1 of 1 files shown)


sample/src/main/java/com/google/android/horologist/paging/PagingScreen.kt
```diff
@@ -73,38 +77,34 @@ fun PagingScreen(navController: NavController) {
...
-            if (lazyPagingItems.loadState.append == LoadState.Loading) {
-                item {
-                    CircularProgressIndicator(
-                        modifier = Modifier
-                            .fillMaxWidth()
-                            .wrapContentWidth(Alignment.CenterHorizontally)
-                    )
+                PagingItemCard(item) {
+                    if (item != null) {
+                        navController.navigate("pagingItem?id=${item.item}")
```
luizgrp, line 86: nit: it would be nice to have this organised like in [media-sample](https://github.com/google/horologist/blob/a46362113077bbd1ccb27a082def16971c1fe760/media-ui/src/main/java/com/google/android/horologist/media/ui/navigation/NavigationScreens.kt#L114)

sample/src/main/java/com/google/android/horologist/paging/PagingScreen.kt
```diff
@@ -218,3 +218,25 @@ data class PagingItem(
...
+            delay(1000)
+            item = PagingItem(10)
+        }
+        PagingItemCard(modifier = Modifier.fillMaxWidth(), item = item)
+    }
+}
+
+@WearSquareDevicePreview
+@Composable
+fun PagingItemCardPreview2() {
```
luizgrp, line 237: nit: better name, e.g. PagingItemCardPreviewWithoutDelay?

sample/src/main/java/com/google/android/horologist/paging/PagingScreen.kt
```diff
@@ -73,38 +77,34 @@ fun PagingScreen(navController: NavController) {
...
     item: PagingItem?,
-    navController: NavController
+    modifier: Modifier = Modifier,
+    onClick: () -> Unit = {}
 ) {
     val chipPlaceholderState = rememberPlaceholderState { item != null }
 
     TitleCard(
-        modifier = Modifier
+        modifier = modifier
```
luizgrp, line 103: 👍

luizgrp (approved):
LGTM

sample/src/main/java/com/google/android/horologist/paging/PagingScreen.kt
yschimke: I agree, but it's a pretty big change. The awkward thing with these samples is as we spend more time on them, they tend towards real apps with modularity, DI. I'm going to keep as is for now, but wouldn't object if you feel strongly enough to refactor.

It also points to whether we should look for some compose-navigation friendly DSL library. https://github.com/raamcosta/compose-destinations

luizgrp: no strong feelings to change this, hence marked as nit comment

PR #781, opened 2022-11-25
[DONOTMERGE] Check if tests from non Android modules are running in CI

luizgrp:
#### WHAT

Check if tests from non Android modules are running in CI.

#### HOW

Add a unit test that should fail.

### DIFF
--- a/media/src/test/java/com/google/android/horologist/media/FailingTest.kt
+++ b/media/src/test/java/com/google/android/horologist/media/FailingTest.kt
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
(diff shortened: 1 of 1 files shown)


luizgrp:
Test failed in CI:

```
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':media:test'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:142)
```

Checked in artifacts:
```
java.lang.AssertionError: This test should fail.
	at org.junit.Assert.fail(Assert.java:89)
```

PR #791, opened 2022-11-29
Added fromDate and toDate parameter to DatePicker

rajat4064g:
#### WHAT
Added `fromDate` and `toDate` parameter to `DatePicker` in composables.

#### WHY
DatePicker didn't support adding restrictions on dates both before and after. This change will allow restricting the user from picking a date before `fromDate` and after `toDate`.

#### HOW
Adjusted number of items in pickers and offsets using `fromDate` and `toDate` values to restrict the display of date picker. 
Both of these parameters are kept optional making it backward compatible.

### DIFF
--- a/composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
+++ b/composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
@@ -28,13 +28,12 @@ import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithContentDescription
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.test.performClick
-import androidx.wear.compose.material.PickerState
 import androidx.wear.compose.material.Text
 import com.google.android.horologist.composables.DatePicker
-import com.google.android.horologist.composables.DatePickerMonthState
-import com.google.android.horologist.composables.DatePickerYearState
+import com.google.android.horologist.composables.DatePickerState
 import com.google.android.horologist.composables.ExperimentalHorologistComposablesApi
 import com.google.common.truth.Truth.assertThat
+import org.junit.Assert
 import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
@@ -46,7 +45,7 @@ class DatePickerTest {
     val composeTestRule = createComposeRule()
 
     @Test
-    fun testDate() {
+    fun onDateConfirm_called_when_confirm_clicked() {
         var date by mutableStateOf<LocalDate?>(null)
         composeTestRule.setContent {
             if (date == null) {
@@ -67,9 +66,8 @@ class DatePickerTest {
     }
 
     @Test
-    fun fromDate_after_toDate() {
-        var exception: Exception? = null
-        try {
+    fun should_throw_if_fromDate_after_toDate() {
+        Assert.assertThrows(IllegalArgumentException::class.java) {
             composeTestRule.setContent {
                 DatePicker(
                     onDateConfirm = {},
@@ -78,19 +76,12 @@ class DatePickerTest {
                     toDate = LocalDate.of(2022, 2, 25)
                 )
             }
-        } catch (e: Exception) {
-            exception = e
-        }
-
-        composeTestRule.runOnIdle {
-            assertThat(exception).isInstanceOf(IllegalArgumentException::class.java)
         }
     }
 
     @Test
-    fun date_outsideRange() {
-        var exception: Exception? = null
-        try {
+    fun should_throw_if_date_outside_from_to_range() {
+        Assert.assertThrows(IllegalArgumentException::class.java) {
             composeTestRule.setContent {
                 DatePicker(
                     onDateConfirm = {},
@@ -99,26 +90,61 @@ class DatePickerTest {
                     toDate = LocalDate.of(2022, 3, 25)
                 )
             }
-        } catch (e: Exception) {
-            exception = e
-        }
-
-        composeTestRule.runOnIdle {
-            assertThat(exception).isInstanceOf(IllegalArgumentException::class.java)
         }
     }
 
     @Test
-    fun testDatePickerState() {
-        val yearState: PickerState = PickerState(initialNumberOfOptions = 100, initiallySelectedOption = 0)
-        val monthState: PickerState = PickerState(initialNumberOfOptions = 12, initiallySelectedOption = 0)
+    fun current_values_checked_in_date_picker_state() {
+        val date = LocalDate.of(2022, 4, 25)
 
-        val datePickerYearState = DatePickerYearState(yearState)
-        val datePickerMonthState = DatePickerMonthState(datePickerYearState, monthState)
+        val datePickerState = DatePickerState(date, date, date)
 
-        assertThat(datePickerYearState.selectedYearEqualsFromYear).isTrue()
-        assertThat(datePickerYearState.selectedYearEqualsToYear).isFalse()
-        assertThat(datePickerMonthState.selectedMonthEqualsFromMonth).isTrue()
-        assertThat(datePickerMonthState.selectedMonthEqualsToMonth).isFalse()
+        assertThat(datePickerState.currentYear()).isEqualTo(2022)
+        assertThat(datePickerState.currentMonth()).isEqualTo(4)
+        assertThat(datePickerState.currentDay()).isEqualTo(25)
+    }
+
+    @Test
+    fun year_state_initialised_to_fromYear() {
+        val date = LocalDate.of(2022, 4, 25)
+
+        val datePickerState = DatePickerState(date = date, fromDate = date, toDate = date.plusYears(1))
+
+        assertThat(datePickerState.selectedYearEqualsFromYear).isTrue()
+        assertThat(datePickerState.selectedYearEqualsToYear).isFalse()
+        assertThat(datePickerState.numOfMonths).isEqualTo(9)
+    }
+
+    @Test
+    fun month_state_initialised_to_fromMonth() {
+        val date = LocalDate.of(2022, 4, 25)
+
+        val datePickerState = DatePickerState(date = date, fromDate = date, toDate = date.plusYears(1))
+
+        assertThat(datePickerState.selectedMonthEqualsFromMonth).isTrue()
+        assertThat(datePickerState.selectedMonthEqualsToMonth).isFalse()
+        assertThat(datePickerState.numOfDays).isEqualTo(6)
+    }
+
+    @Test
+    fun year_state_initialised_to_toYear() {
+        val date = LocalDate.of(2022, 4, 25)
+
+        val datePickerState = DatePickerState(date = date, fromDate = date.minusYears(1), toDate = date)
+
+        assertThat(datePickerState.selectedYearEqualsFromYear).isFalse()
+        assertThat(datePickerState.selectedYearEqualsToYear).isTrue()
+        assertThat(datePickerState.numOfMonths).isEqualTo(4)
+    }
+
+    @Test
+    fun month_state_initialised_to_toMonth() {
+        val date = LocalDate.of(2022, 4, 25)
+
+        val datePickerState = DatePickerState(date = date, fromDate = date.minusYears(1), toDate = date)
+
+        assertThat(datePickerState.selectedMonthEqualsFromMonth).isFalse()
+        assertThat(datePickerState.selectedMonthEqualsToMonth).isTrue()
+        assertThat(datePickerState.numOfDays).isEqualTo(25)
     }
 }
--- a/composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
+++ b/composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
@@ -28,9 +28,13 @@ import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithContentDescription
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.test.performClick
+import androidx.wear.compose.material.PickerState
 import androidx.wear.compose.material.Text
 import com.google.android.horologist.composables.DatePicker
+import com.google.android.horologist.composables.DatePickerMonthState
+import com.google.android.horologist.composables.DatePickerYearState
 import com.google.android.horologist.composables.ExperimentalHorologistComposablesApi
+import com.google.common.truth.Truth.assertThat
 import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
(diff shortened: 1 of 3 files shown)


composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -79,36 +83,108 @@ public fun DatePicker(
...
-            initiallySelectedOption = date.year - 1
-        )
-        val monthState = rememberPickerState(
-            initialNumberOfOptions = 12,
-            initiallySelectedOption = date.monthValue - 1
-        )
-        val maxDayInMonth by remember {
+        val yearState =
+            rememberPickerState(
+                initialNumberOfOptions = (toDate?.year ?: 3000) - ((fromDate?.year?.minus(1)) ?: 0),
```
yschimke, line 88: supernit: Seems to make assumptions about the range of year. Maybe worth adding a check if we have a from year > 3000.

There are many ways we can go with this

a) make the null defaults smarter (to defaults to from + 3000)
b) enforce sensible ranges
c) Use reasonable defaults, like +-100 years.

yschimke: Also what to do if date is outside from and to?

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -79,36 +83,108 @@ public fun DatePicker(
...
+            )
+        val selectedYearEqualsFromYear by
+        remember(yearState) { derivedStateOf { fromDate != null && yearState.selectedOption == 0 } }
+        val selectedYearEqualsToYear by
+        remember(yearState) {
+            derivedStateOf {
+                toDate != null && yearState.selectedOption == yearState.numberOfOptions - 1
+            }
+        }
+        val numOfMonths by
```
yschimke, line 99: formatting indentation looks a bit off.

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -79,36 +83,108 @@ public fun DatePicker(
...
+        remember(yearState) {
+            derivedStateOf {
+                (
+                    if (selectedYearEqualsToYear) {
```
yschimke, line 103: A suggestion, a lot of these calculations seem testable, would it make sense to pull them out to a DatePickerState, which would presumable allow testing this logic also?

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -162,32 +238,53 @@ public fun DatePicker(
...
                     DatePickerImpl(
                         state = monthState,
                         readOnly = selectedColumn != 1,
                         onSelected = { selectedColumn = 1 },
-                        text = { month: Int -> monthNames[month] },
+                        text = { month: Int ->
+                            monthNames[
+                                (
+                                    month +
+                                        (if (selectedYearEqualsFromYear) fromDate?.monthValue?.minus(1) ?: 0 else 0)
```
yschimke, line 264: I think my comment about DatePickerState applies to whereever you have a check on selectedYearEqualsFromYear.

yschimke:
Can you add some unit tests? And also adjust some of the samples, or add a new one.

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
rajat4064g: Added default of +- 100 years

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
rajat4064g: indentation done by spotlessApply command

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
rajat4064g: Created DatePickerYearState and DatePickerMonthState and added test for them

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
rajat4064g: Same as above

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
```diff
@@ -61,4 +65,60 @@ class DatePickerTest {
...
+                    date = LocalDate.of(2022, 3, 25),
+                    fromDate = LocalDate.of(2022, 3, 25),
+                    toDate = LocalDate.of(2022, 2, 25)
+                )
+            }
+        } catch (e: Exception) {
+            exception = e
+        }
+
+        composeTestRule.runOnIdle {
```
BowerSteve, line 85: Please see if we can use simplify this structure using Assert.assertThrows, like this example: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/src/test/kotlin/androidx/compose/ui/node/LayoutNodeTest.kt;l=539

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
```diff
@@ -61,4 +65,60 @@ class DatePickerTest {
 
         composeTestRule.onNodeWithTag("date").assertTextEquals("2022-04-25")
     }
+
+    @Test
+    fun fromDate_after_toDate() {
```
BowerSteve, line 70: Please make the test purpose clearer in the name, such as 'should_throw_if_fromdate_after_todate'

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
```diff
@@ -61,4 +65,60 @@ class DatePickerTest {
 
         composeTestRule.onNodeWithTag("date").assertTextEquals("2022-04-25")
     }
+
```
BowerSteve, line 68: Please rename testDate to 'onDateConfirm_called_when_confirm_clicked' or similar.

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
```diff
@@ -61,4 +65,60 @@ class DatePickerTest {
...
+            exception = e
+        }
+
+        composeTestRule.runOnIdle {
+            assertThat(exception).isInstanceOf(IllegalArgumentException::class.java)
+        }
+    }
+
+    @Test
+    fun date_outsideRange() {
```
BowerSteve, line 91: Please update the name - e.g. 'should_throw_if_date_outside_from_to_range'

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
```diff
@@ -61,4 +65,60 @@ class DatePickerTest {
...
+                    fromDate = LocalDate.of(2022, 2, 25),
+                    toDate = LocalDate.of(2022, 3, 25)
+                )
+            }
+        } catch (e: Exception) {
+            exception = e
+        }
+
+        composeTestRule.runOnIdle {
+            assertThat(exception).isInstanceOf(IllegalArgumentException::class.java)
```
BowerSteve, line 107: As above, see if Assert.assertThrows can be used here to eliminate the need for the try-catch.

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
```diff
@@ -61,4 +65,60 @@ class DatePickerTest {
...
+        }
+    }
+
+    @Test
+    fun testDatePickerState() {
```
BowerSteve, line 112: Please split this into 2 tests, one for the year and one for the month. Then the test name can be something like 'year_state_initialised_to_from_year' etc.

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -65,50 +65,117 @@ import java.time.temporal.TemporalAdjusters
...
+ * @param toDate the maximum date to be selected in picker
  */
 @Composable
 public fun DatePicker(
     onDateConfirm: (LocalDate) -> Unit,
     modifier: Modifier = Modifier,
-    date: LocalDate = LocalDate.now()
+    date: LocalDate = LocalDate.now(),
+    fromDate: LocalDate = date.minusYears(100),
+    toDate: LocalDate = date.plusYears(100)
```
BowerSteve, line 77: Why not make these fully optional e.g. LocalDate? = null, so that we don't have to provide meaningful default values. And this way, if the value is null, no validation needs to be performed.

BowerSteve (commented):
I agree with Yuri that we should try to move the logic for the day/month/year out of the UI into a separate class (could be a single DateState class or separate ones for day/month/year if that works better). As it stands, the logic for e.g. the month appears in several places, which will be hard to maintain.

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
rajat4064g: Default value of +- 100 years was provided to make the ranges reasonable. In case of null, 3000 was the default last year. Check - https://github.com/google/horologist/pull/791#discussion_r1037151972

yschimke: I'd go with Steve's advice over mine. My original concern was that with the defaults, it was easy to pass in a combination of parameters that is invalid. 

One of the other options was to make the `null` defaults smarter (`to` defaults to `from + 3000`)

BowerSteve: Ok, how about trying null defaults, but if either one of the values is provided, the other default is that value +/- 3000. So if a toDate is provided, the fromDate defaults to toDate - 3000. And if a fromDate is provided, toDate defaults to fromDate + 3000.

rajat4064g: done

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
rajat4064g: done

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
rajat4064g: done

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
rajat4064g: done

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
rajat4064g: done

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
rajat4064g: done

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
rajat4064g: done

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
```diff
@@ -61,4 +64,87 @@ class DatePickerTest {
...
+
+        val datePickerState = DatePickerState(date, date, date)
+
+        assertThat(datePickerState.currentYear()).isEqualTo(2022)
+        assertThat(datePickerState.currentMonth()).isEqualTo(4)
+        assertThat(datePickerState.currentDay()).isEqualTo(25)
+    }
+
+    @Test
+    fun year_state_initialised_to_fromYear() {
```
BowerSteve, line 108: Hmm - not sure I understand how year_state_initialised_to_fromYear and year_state_initialised_to_toYear can both pass, they seem to be testing opposite conditions.

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
```diff
@@ -61,4 +64,87 @@ class DatePickerTest {
...
+
+        val datePickerState = DatePickerState(date = date, fromDate = date.minusYears(1), toDate = date)
+
+        assertThat(datePickerState.selectedYearEqualsFromYear).isFalse()
+        assertThat(datePickerState.selectedYearEqualsToYear).isTrue()
+        assertThat(datePickerState.numOfMonths).isEqualTo(4)
+    }
+
+    @Test
+    fun month_state_initialised_to_toMonth() {
```
BowerSteve, line 141: Same here - how can month_state_initialised_to_toMonth and month_state_initialised_to_fromMonth both pass?

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
```diff
@@ -61,4 +64,87 @@ class DatePickerTest {
...
+
+    @Test
+    fun month_state_initialised_to_fromMonth() {
+        val date = LocalDate.of(2022, 4, 25)
+
+        val datePickerState = DatePickerState(date = date, fromDate = date, toDate = date.plusYears(1))
+
+        assertThat(datePickerState.selectedMonthEqualsFromMonth).isTrue()
+        assertThat(datePickerState.selectedMonthEqualsToMonth).isFalse()
+        assertThat(datePickerState.numOfDays).isEqualTo(6)
```
BowerSteve, line 126: Is this check relevant to this test?

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
```diff
@@ -61,4 +64,87 @@ class DatePickerTest {
...
+
+    @Test
+    fun month_state_initialised_to_toMonth() {
+        val date = LocalDate.of(2022, 4, 25)
+
+        val datePickerState = DatePickerState(date = date, fromDate = date.minusYears(1), toDate = date)
+
+        assertThat(datePickerState.selectedMonthEqualsFromMonth).isFalse()
+        assertThat(datePickerState.selectedMonthEqualsToMonth).isTrue()
+        assertThat(datePickerState.numOfDays).isEqualTo(25)
```
BowerSteve, line 148: Can this check be removed? Doesn't seem relevant to this test.

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -250,3 +245,80 @@ private fun DatePickerImpl(
...
+}
+
+internal class DatePickerState constructor(
+    private val date: LocalDate,
+    private val fromDate: LocalDate?,
+    private val toDate: LocalDate?
+) {
+    private val yearOffset = fromDate?.year ?: 1
+    val yearState =
+        PickerState(
```
BowerSteve, line 265: Shouldn't these be using rememberPickerState? Otherwise they would be recomposed when the inputs change. Or is that covered by using remember DatePickerState?

rajat4064g: Yes, it is covered by using remember DatePickerState

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -250,3 +245,80 @@ private fun DatePickerImpl(
...
+
+    private val dayOffset: Int
+        get() = (if (selectedMonthEqualsFromMonth) (fromDate?.dayOfMonth ?: 1) else 1)
+    private val firstDayOfMonth: LocalDate
+        get() = LocalDate.of(
+            currentYear(),
+            currentMonth(),
+            1
+        )
+    private val maxDaysInMonth: Int
```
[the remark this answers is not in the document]
rajat4064g: Day flicker doesn't happen in this version. We're initialising PickerState only once here.

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
rajat4064g: datePickerState.selectedYearEqualsToYear is false here. In `year_state_initialised_to_toYear `, datePickerState is initialised to `DatePickerState(date = date, fromDate = date.minusYears(1), toDate = date)`.

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
rajat4064g: I think, this check will ensure that the logic behind number of days shown on DayPicker is correct.

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
rajat4064g: In `month_state_initialised_to_fromMonth `, 
`datePickerState = DatePickerState(date = date, fromDate = date, toDate = date.plusYears(1))`.

Where as in `month_state_initialised_to_toMonth `,
`datePickerState = DatePickerState(date = date, fromDate = date.minusYears(1), toDate = date)`

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
rajat4064g: Same as above, this check will ensure that the logic behind number of days shown on DayPicker is correct.

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
BowerSteve: Oh I see, it's always taken from the supplied date parameter and you're checking equality with the other parameters.

composables/src/androidTest/java/com/google/android/horologist/DatePickerTest.kt
BowerSteve: Got it, thanks.

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -250,3 +245,80 @@ private fun DatePickerImpl(
...
+        get() = (yearState.selectedOption == 0)
+    val selectedYearEqualsToYear: Boolean
+        get() = (yearState.selectedOption == yearState.numberOfOptions - 1)
+
+    private val monthOffset: Int
+        get() = (if (selectedYearEqualsFromYear) (fromDate?.monthValue ?: 1) else 1)
+    private val maxMonths: Int
+        get() = (if (selectedYearEqualsToYear) (toDate?.monthValue ?: 12) else 12)
+    val numOfMonths: Int
+        get() = (maxMonths.minus(monthOffset - 1))
```
BowerSteve, line 279: Please can we have some more tests to check that this logic in DatePickerState is working as expected. For example, if the from date is 2022/Nov/23 then the day picker should only have 8 options and the month picker should only have 2 options, until a different year is selected. And the year picker should not show any years before 2022.

rajat4064g: done

BowerSteve (approved):
Thanks Rajat, looks good.

yschimke (approved):
Thanks for extracting the logic, looks much better.

PR #794, opened 2022-11-30
Makes media-data take MediaMetadata into account

fstanis:
#### WHAT
Improves how we determine title and artist of playing media. Also modifies our UI model names from "arist" -> "subtitle" which makes it clearer how the string is used.

#### WHY
The current method relies solely on data provided inside `MediaItem`. In practice, this means adding a `MediaItem` with only a URL will result in empty title and artist.

#### HOW
ExoPlayer extracts metadata after playback starts and emits a `EVENT_MEDIA_METADATA_CHANGED`. We can leverage this event and update the metadata when it happens, otherwise fallback to `MediaItem`.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests

### DIFF
--- a/media-data/api/current.api
+++ b/media-data/api/current.api
@@ -240,11 +240,11 @@ package com.google.android.horologist.media.data.mapper {
   }
 
   @com.google.android.horologist.media.data.ExperimentalHorologistMediaDataApi public interface MediaExtrasMapper {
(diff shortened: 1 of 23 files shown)


yschimke:
A bit awkward to have breaking API change, but on the balance probably better to correct once.

And I was tempted to ask for artist also, but the point is the UI composables shouldn't make those decisions.

yschimke (approved):

PR #795, opened 2022-12-02
Removed scrolling delay from Rotary algorithm.

Kpeved:
#### WHAT
Removed scrolling delay from Rotary algorithm.

#### WHY
Scrolling delay is used for filtering negative values at the end of RSB scroll. It was done for a proper fling tracking. This delay was also causing scrolling to have a pretty long delay when direction of the scroll was reversed.

#### HOW
This delay was removed and was left only for Fling tracking and not for scrolling

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
@@ -828,6 +828,7 @@ internal class HighResRotaryScrollHandler(
         rotaryHaptics: RotaryHapticFeedback
     ) {
         val time = event.timestamp
+        val isOppositeScrollValue = isOppositeValueAfterScroll(event.delta)
 
         if (isNewScrollEvent(time)) {
             debugLog { "New scroll event" }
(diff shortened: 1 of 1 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
```diff
@@ -836,16 +837,15 @@ internal class HighResRotaryScrollHandler(
         } else {
             // Filter out opposite axis values from end of scroll, also some values
             // at the start of motion which sometimes appear with a different sign
```
yschimke, line 839: Does the comment need to change?

yschimke (approved):
A bit tough to review as is.  Discussed separately need for tests later, but manual testing seem OK at this early stage.

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
Kpeved: Yes, changed it

PR #798, opened 2022-12-02
Animate media progress bar

fstanis:
#### WHAT
Makes `MediaControlButtons` and others have animated progress.

#### WHY
Generally provides better UX, especially for shorter media and when moving from one media item to another.

#### HOW
By using `animateFloatAsState` for setting progress.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/screens/player/PlayerScreen.kt
```diff
@@ -134,7 +134,8 @@ public fun DefaultPlayerScreenControlButtons(
         onSeekToNextButtonClick = { playerController.skipToNextMedia() },
         seekToNextButtonEnabled = playerUiState.seekToNextEnabled,
         showProgress = playerUiState.trackPosition?.showProgress ?: true,
-        percent = playerUiState.trackPosition?.percent ?: 0f
+        percent = playerUiState.trackPosition?.percent ?: 0f,
+        animateProgress = true
```
yschimke, line 138: We should probably discuss the difference between AnimatedXButtons and XButtons.

Do we want some 

a) simple buttons with no animation
b) buttons with full animation including lottie
c) simple buttons with some (non-lottie) animation

I've been aiming for a+b.

media-ui/src/main/java/com/google/android/horologist/media/ui/components/PlayPauseButton.kt
```diff
@@ -108,10 +113,14 @@ public fun PlayPauseProgressButton(
                 .clip(CircleShape)
                 .background(backgroundColor)
         ) {
+            val progress by animateFloatAsState(
+                targetValue = percent.ifNan(0f),
+                animationSpec = if (animateProgress) ProgressIndicatorDefaults.ProgressAnimationSpec else snap()
```
yschimke, line 118: Does this default to ?

```
const val DefaultDurationMillis: Int = 300
```

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/player/PlayerScreen.kt
yschimke: Should we remove the animateProgress flag and just go for b+c? Turn this on by default?

fstanis: Agreed, removed flag.

yschimke (approved):

fstanis:
After some discussion, this will probably be superseded completely by #806, so treat it as a stopgap solution towards smoother progress.

PR #803, opened 2022-12-05
Enable manual run of build & test github action

luizgrp:
#### WHAT

Enable [manual run](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) of [build & test](https://github.com/google/horologist/blob/fd18cf42b6541a64df4708776a8d770b973c0418/.github/workflows/build.yml) github action.

#### WHY

For some reason one of my pull requests did not trigger the github action, so I'd like to trigger it manually.

Alternatively to this solution, I could push an empty commit.

#### HOW

Add [workflow_dispatch](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) event to build.yml.

#### Checklist :clipboard:
- [N/A] Run spotless check
- [N/A] Run tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,7 @@ on:
     paths-ignore:
       - '**.md'
   pull_request:
+  workflow_dispatch:
 
 jobs:
   build:


fstanis (approved):

luizgrp:
just noticed that this might not work for the scenario described as github does not seem to list branches from forked repos

PR #810, opened 2022-12-07
Set media at a specific position

kul3r4:
#### WHAT
Make it possible to set media at a specific position

#### WHY
Useful for podcasts where content can be often paused and should be resumed where left

#### HOW
Add a new method to PlayerRepository setMedia (media, position)

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/media-data/api/current.api
+++ b/media-data/api/current.api
@@ -329,6 +329,7 @@ package com.google.android.horologist.media.data.repository {
     method public void seekToDefaultPosition(int mediaIndex);
     method public void setMedia(com.google.android.horologist.media.model.Media media);
     method public void setMediaList(java.util.List<com.google.android.horologist.media.model.Media> mediaList);
+    method public void setMediaList(java.util.List<com.google.android.horologist.media.model.Media> mediaList, int index, kotlin.time.Duration? position);
     method public void setMediaListAndPlay(java.util.List<com.google.android.horologist.media.model.Media> mediaList, int index);
     method public void setPlaybackSpeed(float speed);
     method public void setShuffleModeEnabled(boolean shuffleModeEnabled);
--- a/media-data/src/main/java/com/google/android/horologist/media/data/repository/PlayerRepositoryImpl.kt
+++ b/media-data/src/main/java/com/google/android/horologist/media/data/repository/PlayerRepositoryImpl.kt
@@ -17,6 +17,7 @@
 package com.google.android.horologist.media.data.repository
 
 import android.util.Log
+import androidx.media3.common.C
 import androidx.media3.common.MediaItem
 import androidx.media3.common.Player
 import androidx.media3.common.Timeline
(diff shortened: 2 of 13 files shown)


media-data/src/main/java/com/google/android/horologist/media/data/repository/PlayerRepositoryImpl.kt
```diff
@@ -319,6 +319,19 @@ public class PlayerRepositoryImpl(
         }
     }
 
+    /**
+     * This operation will stop the current MediaItem that is playing, if there is one, as per
+     * [Player.setMediaItem] and seek to the specified position.
+     */
+    override fun setMedia(media: Media, position: Long) {
```
yschimke, line 326: While this is probably for a single item, do we want to make this on the setMediaList method. I could see us going either way.

But API wise, maybe we want

- a few common specific methods
- one form with all options that allows full flexibility

I don't know the answer, so consider this raising to discuss.

kul3r4: Added ` setMediaList(mediaList: List<Media>, index: Int, position: Long)`

media-data/api/current.api
```diff
@@ -328,7 +328,9 @@ package com.google.android.horologist.media.data.repository {
     method public void seekForward();
     method public void seekToDefaultPosition(int mediaIndex);
     method public void setMedia(com.google.android.horologist.media.model.Media media);
+    method public void setMedia(com.google.android.horologist.media.model.Media media, long position);
```
yschimke, line 331: I'd suggest removing this one, just use the advanced one.

kul3r4: done

yschimke (approved):

media/src/main/java/com/google/android/horologist/media/repository/PlayerRepository.kt
```diff
@@ -136,6 +136,17 @@ public interface PlayerRepository {
...
+    /**
+     * Clears the playlist, adds the specified [Media] list and resets the position to
+     * the provided position.
+     *
+     * @param mediaList The new [Media] list.
+     * @param index The [Media] index to start playback from
+     * @param position The media position.
+     */
+
+    public fun setMediaList(mediaList: List<Media>, index: Int, position: Long)
```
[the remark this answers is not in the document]
luizgrp: should we make `position` nullable and provide a default null value so we don't force users to pass one?

the reason I'm suggesting to make it nullable instead of zero is due to the fact that the "default position" of a media [does not seem to be zero in media3](https://github.com/androidx/media/blob/c2cbb6370a265efc93661b659e70b2679506e995/libraries/common/src/main/java/androidx/media3/common/Player.java#L1590-L1593)

luizgrp (approved):
LGTM - few comments to consider

laiyichin:
LGTM!

Issue #800, opened 2022-12-04
Suggestion: Make it possible to add a brush track colour to ProgressIndicatorSegment


oas004:
This is just a suggestion. Please feel free to close this if you don't agree :) 

In the current API for the SegmentedProgressIndicator and the SquareSegmentedProgressIndicator composables, the track segments has to be defined as follows

```
public data class ProgressIndicatorSegment(
    val weight: Float,
    val indicatorColor: Color,
    val trackColor: Color? = null,
    val inProgressTrackColor: Color? = null
)
```

I was thinking it could be cool to provide an alternative to this where the indicator colour can be a brush like this:

```
public data class ProgressIndicatorSegment(
    val weight: Float,
    val indicatorColor: Brush,
    val trackColor: Color? = null,
    val inProgressTrackColor: Color? = null
)
```
in addition to the API that already is. Maybe it could be an overriding function or something?

This would make it easy to add a gradient on each of the segments making the "increasing progress" effect independent of the segments themself.
With a change like this, it would be easy to make something like this

![Screenshot 2022-12-04 at 03 46 51](https://user-images.githubusercontent.com/54936943/205471650-f1346d04-092f-4d57-8f0b-6443fde73c4b.png)

Even though you only have 3 segments defined.

What do you think? :)

yschimke:
@garanj thoughts?

This makes sense to me. Are you using this in an app that would use this, or I'd it a general improvement?

oas004:
> @garanj thoughts?
> 
> This makes sense to me. Are you using this in an app that would use this, or I'd it a general improvement?

I was thinking as a general improvement. Just thinking it could look a bit better if the user was using few segments, so could be nice to have this opportunity 😊

garanj:
Sounds great!👍

garanj:
How about for colors applied to individual track segments, could the developer specify different brushes for each?

oas004:
I was thinking maybe if there was an override that took a segment with this api
```
public data class ProgressIndicatorSegment(
    val weight: Float,
    val indicatorColor: Brush,
    val trackColor: Color? = null,
    val inProgressTrackColor: Color? = null
)
```

When calling it you could then in theory have different brushes like this 

```
val previewProgressSections = listOf(
    ProgressIndicatorSegment(
        weight = 3f,
        indicatorColor = Brush.horizontalGradient(listOf(Color.Yellow, Color.Blue))
    ),
    ProgressIndicatorSegment(
        weight = 3f,
[...]

PR #819, opened 2022-12-11
Update ProgressSegments to be able to take a brush color as argument.

oas004:
#### WHAT
Update the ProgressIndicatorSegments used in the segmented progress indicator composables, to have a variant that uses brush color instead of just color.

#### WHY
Because of #800 

#### HOW
Updated the ProgressIndicatorSegments API to be a sealed interface that can either be one variant that takes the color or one variant that takes the brush color. 

This is a breaking change in the API. The composables were annotated with the experimental annotations. Is it okay to introduce a breaking change to the APIs?

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


composables/src/main/java/com/google/android/horologist/composables/SegmentedProgressIndicator.kt
```diff
@@ -34,24 +35,50 @@ import androidx.wear.compose.material.MaterialTheme
...
+     * @param weight The proportion of the overall progress indicator that this segment should take up,
+     * as a proportion of the sum of the weights of all the other segments.
+     * @param indicatorColor The color of the indicator bar.
+     * @param trackColor The color of the background progress track. This is optional and if not
+     * specified will default to the [trackColor] of the overall [SegmentedProgressIndicator].
+     * @param inProgressTrackColor The color of the background progress track when this segment is in
+     * progress. This is optional, and if specified will take precedence to the [trackColor] of the
+     * segment or the overall [SegmentedProgressIndicator], when the segment is in progress.
+     */
+    public data class ProgressIndicatorSegmentDefault(
```
oas004, line 56: I don't quite know if this is the best name for this class 🤔 This is the one that can take a color and not a brush.

composables/src/main/java/com/google/android/horologist/composables/SegmentedProgressIndicator.kt
```diff
@@ -34,24 +35,50 @@ import androidx.wear.compose.material.MaterialTheme
...
- * progress. This is optional, and if specified will take precedence to the [trackColor] of the
- * segment or the overall [SegmentedProgressIndicator], when the segment is in progress.
- */
-public data class ProgressIndicatorSegment(
-    val weight: Float,
-    val indicatorColor: Color,
-    val trackColor: Color? = null,
-    val inProgressTrackColor: Color? = null
-)
+public sealed interface ProgressIndicatorSegment {
```
oas004, line 38: Is this an okay approach to this problem? I could not quite get it to work with a secondary constructor.  Is there some cool Kotlin magic I can apply here instead? :)

composables/src/main/java/com/google/android/horologist/composables/SegmentedProgressIndicator.kt
```diff
@@ -185,14 +212,30 @@ private fun DrawScope.drawCircularProgressIndicator(
...
-        useCenter = false,
-        topLeft = offset,
-        size = Size(arcDimen, arcDimen),
-        style = stroke
-    )
+    when (segment) {
+        is ProgressIndicatorSegment.ProgressIndicatorSegmentBrush -> {
+            // Draw Progress
+            drawArc(
+                brush = segment.indicatorBrush,
```
yschimke, line 219: Can we use Brush for both code paths here?

composables/src/main/java/com/google/android/horologist/composables/SegmentedProgressIndicator.kt
oas004: Aha, yeah that makes more sense! I will do that instead! Thanks!

yschimke (approved):

PR #820, opened 2022-12-11
Add snapshot tests for ShuffleToggleButton

### LINKED ISSUE #802: Replace ShuffleToggleButtonTest by snapshot tests
#### WHAT

Replace [ShuffleToggleButtonTest](https://github.com/google/horologist/blob/fd18cf42b6541a64df4708776a8d770b973c0418/media-ui/src/androidTest/java/com/google/android/horologist/media/ui/components/controls/ShuffleToggleButtonTest.kt) by snapshot tests.

#### WHY

Snapshot tests are more appropriate for what those tests are asserting.

#### HOW

- Add snapshot tests and golden images;
- Remove `ShuffleToggleButtonTest` and [hasIconImageVector](https://github.com/google/horologist/blob/fd18cf42b6541a64df4708776a8d770b973c0418/media-ui/src/androidTest/java/com/google/android/horologist/test/toolbox/matchers/SemanticsMatchers.kt#L26) function;

oas004:
#### WHAT
Add snapshot tests for ShuffleToggleButton and removing UI tests that had the same purpose as this snapshot test. 
Also removed the `hasIconImageVector`semantics matcher. 

#### WHY
Because of #802 

#### HOW
Added snapshot tests in the test source set and remove the test from AndroidTest source set. 

resolves #802 

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/androidTest/java/com/google/android/horologist/test/toolbox/matchers/HasProgressBar.kt
```diff
@@ -16,15 +16,8 @@
 
 package com.google.android.horologist.test.toolbox.matchers
 
-import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.semantics.SemanticsProperties
 import androidx.compose.ui.test.SemanticsMatcher
-import com.google.android.horologist.media.ui.ExperimentalHorologistMediaUiApi
-import com.google.android.horologist.media.ui.semantics.CustomSemanticsProperties.IconImageVectorKey
-
-@ExperimentalHorologistMediaUiApi
```
oas004, line 25: Because I removed this function, spotless was complaining saying that I have to rename this class to match the function name of the function that is left in the file. Can I suppress this and name the file `SemanticsMatchers.kt`? Or should I keep the name `HasProgressBar.kt` ?

yschimke: Maybe suppress, so we don't keep flip flopping

oas004: Is there a way to suppress spotless for only this file? I tried 
```
@file:Suppress("filename")
```
But I don't think that suppressed spotless for this issue 🤔

yschimke: Rename if you have to, or leave the existing method.

luizgrp (approved):
LGTM - thank you!

media-ui/src/androidTest/java/com/google/android/horologist/test/toolbox/matchers/HasProgressBar.kt
luizgrp: > Is there a way to suppress spotless for only this file?

I think as last resource (if all the other alternatives didn't work), a [lint baseline](https://developer.android.com/studio/write/lint#snapshot) could be generated to ignore the file.

Not asking for this to be done, just joining the conversation!

oas004: Yeah a lint baseline could work! But it should be an easy way to suppress Spotless IMO, but I just could not find it 🤔

PR #824, opened 2022-12-12
Add nightly workflow for release builds

luizgrp:
#### WHAT

Add nightly workflow for release builds.

#### WHY

So that release builds are checked daily but without increasing our build times for each PR submitted.

#### HOW

Add `build_nightly` with [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) webhook for 2am, daily. 

This workflow has loads of room from improvement:
- Run it only if there were new commits;
- Run unit and/or instrumentation tests for release builds;
- Run it also for `release-0.1-x` branch;

#### Checklist :clipboard:
- [N/A] Run spotless check
- [N/A] Run tests


.github/workflows/build_nightly.yml
```diff
@@ -0,0 +1,57 @@
...
+          path: |
+            ~/.gradle/caches/modules-*
+            ~/.gradle/caches/jars-*
+            ~/.gradle/caches/build-cache-*
+          key: gradle-${{ hashFiles('checksum.txt') }}
+
+      - name: Build (Release)
+        run: |
+          ./gradlew --scan --stacktrace \
+              assembleRelease
```
yschimke, line 42: Can you upload these artifacts also? Would be good to have built APKs for occasional testing.

yschimke (approved):

.github/workflows/build_nightly.yml
luizgrp: good idea, done 👍

PR #830, opened 2022-12-13
Fix order of columnState param

luizgrp:
#### WHAT

Fix order of `columnState` param in composables.

#### WHY

> [Modifier](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier) should appear as the first optional parameter in the parameter list; after all required parameters (except for trailing lambda parameters) but before any other parameters with default values.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
@@ -100,8 +100,8 @@ public class ScalingLazyColumnState(
 
 @Composable
 public fun ScalingLazyColumn(
-    modifier: Modifier = Modifier,
     columnState: ScalingLazyColumnState,
+    modifier: Modifier = Modifier,
     content: ScalingLazyListScope.() -> Unit
 ) {
     val focusRequester = rememberActiveFocusRequester()
--- a/auth-ui/api/current.api
+++ b/auth-ui/api/current.api
@@ -12,7 +12,7 @@ package com.google.android.horologist.auth.ui.common.screens {
   }
 
   public final class SignInPromptScreenKt {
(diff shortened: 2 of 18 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
```diff
@@ -100,8 +100,8 @@ public class ScalingLazyColumnState(
 
 @Composable
 public fun ScalingLazyColumn(
-    modifier: Modifier = Modifier,
     columnState: ScalingLazyColumnState,
+    modifier: Modifier = Modifier,
```
luizgrp, line 104: @yschimke are you happy with this change? highlighting this as it's not marked as experimental

yschimke: It probably should be.  Can you add?

yschimke (approved):

compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
luizgrp: will add in another PR as I have another branch chained to this one

luizgrp: There are loads of classes in compose-layout without the experimental annotation. Should I apply it to all or just this one?

luizgrp: https://github.com/google/horologist/pull/837

PR #834, opened 2022-12-13
Removes setMediaListAndPlay, fixes #496

### LINKED ISSUE #496: Revert workaround to fix playing songs at specific index
After https://github.com/androidx/media/issues/85 gets fixed and media3 dependency is updated in the project, revert changes made in https://github.com/google/horologist/pull/497 that are a workaround for this issue:
- `mediaIndexToSeekTo` and `onTimelineChanged` should be removed;
- calls to `seekTo`, `prepare` and `play` should be made by `setMediaListAndPlay`;

As alternative, function `setMediaListAndPlay` should be removed and a `setMediaList(mediaList, index)` added instead. `setMediaListAndPlay` is taking care of playing at the moment to avoid having clients to code a "delay" call as per in https://github.com/google/horologist/pull/494

kul3r4:
#### WHAT
Removes setMediaListAndPlay

#### WHY
to Address #496 

#### HOW
removes the setMediaListAndPOlay and uses setMediaList

#### Checklist :clipboard:
- [x] Run spotless check
- [ ] Run tests


luizgrp (approved):
LGTM - thanks!

PR #838, opened 2022-12-14
Remove double experimental annotation from SectionedList

luizgrp:
#### WHAT

Remove double experimental annotation from `SectionedList`.

#### WHY

Using this component shouldn't require to opt in to two different experimental annotations.

#### HOW

Remove annotation and opt in via build.gradle for the whole module.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Makes sense

Issue #833, opened 2022-12-13
Convert Groovy gradle files to Kotlin DSL


oas004:
Hey! 

This is just a suggestion and I wanted to know what you guys thought about it. Feel free to close it, if it does not make sense.

I noticed that all the gradle files are in Groovy, what do you guys think about converting it to Kotlin DSL?  I saw that the NowInAndroid app has this, but for instance Accompanist does not. Is there a reason for not converting? :)

yschimke:
I'm for it.

I use kotlin build files exclusively on own projects. Our build came from accompanist, so that's why.

But let's wait for others to weigh in.

oas004:
That makes sense! Will gladly help and contribute if this is something we want :) I agree, let's wait for others to weigh in.

luizgrp:
I'm in! :)

oas004:
Are there more people that needs to weigh in on this? :)

I guess this is something that would be nice to do in multiple PRs?  Is there anything that should be considered when doing this? :)

luizgrp:
I'd recommend reading this [doc](https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html) first, there is a suggestion for [first migration step](https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html#prepare_your_groovy_scripts). 

Multiple small PRs sounds good, if possible, maybe module by module after taking the first migration steps?

oas004:
That sounds great! I think the first step docs you sent me was gold! 🙏😊

yschimke:
Cc @alexvanyo in case you want to share approach for accompanist.

PR #846, opened 2022-12-19
Start preparation phase to convert audio domains gradle from Groovy  DSL to Kotlin DSL

oas004:
#### WHAT
Starting the preparation phase for converting the gradle files in the audio domain from Groovy DSL to Kotlin DSL.

#### WHY

See #833 

#### HOW

My plan here is to start with each domain and go through the preparation phases from: https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html#prepare_your_groovy_scripts

On this step, I will separate the gradle files in each domain by commits, so that if you want to review it gradle file by gradle file, you can do so by reviewing commit by commit.

(By domain here I mean audio is one domain, auth is one domain etc.)

Then step 2 is to actually convert each gradle module to Kotlin DSL and rename the files to start using it. For this step I will make a PR for each module.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/audio-ui/build.gradle
+++ b/audio-ui/build.gradle
@@ -15,30 +15,30 @@
  */
 
 plugins {
-    id 'com.android.library'
-    id 'kotlin-android'
-    id 'org.jetbrains.dokka'
-    id "org.jetbrains.kotlin.kapt"
-    id 'dev.chrisbanes.paparazzi'
+    id("com.android.library")
+    id("kotlin-android")
+    id("org.jetbrains.dokka")
+    id("org.jetbrains.kotlin.kapt")
+    id("dev.chrisbanes.paparazzi")
 }
 
 android {
-    compileSdkVersion 33
+    compileSdkVersion = 33
 
     defaultConfig {
-        minSdk 25
-        targetSdk 30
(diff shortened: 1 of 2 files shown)


oas004:
If you guys feel that this approach will lead to to many small PRs I can of course change it :)

audio-ui/build.gradle
```diff
@@ -15,79 +15,78 @@
...
+            res.srcDirs += "src/sharedTest/res"
         }
     }
     lint {
-        checkReleaseBuilds false
-        textReport true
-        disable 'MissingTranslation', 'ExtraTranslation'
+        checkReleaseBuilds = false
+        textReport = true
+        disable "MissingTranslation", "ExtraTranslation"
```
luizgrp, line 83: I wonder if this is a function invocation?

luizgrp (approved):
LGTM

audio-ui/build.gradle
oas004: Yeah I think you are right! Can I change it when I convert the file to .kts? :)

luizgrp: sure!

PR #847, opened 2022-12-19
Prepare auth modules to be converted to Kotlin DSL

oas004:
#### WHAT
Following up #846 and preparing auth modules to be converted to Kotlin DSL

#### WHY
#833 

#### HOW
Follow https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html#prepare_your_groovy_scripts to prepare the Groovy scripts to more easily be converted to Kotlin DSL.

#### Checklist :clipboard:
- [x] Add explicit visibility modifier and explicit return types for public declarations
- [x] Run spotless check
- [x] Run tests
- [x] Update metalava's signature text files

PR #848, opened 2022-12-19
Prepare base-ui to be converted to Kotlin DSL

oas004:
#### WHAT
Following up #847 and preparing the base-ui for being converted to Kotlin DSL

#### WHY
#833 

#### HOW
Following https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html#prepare_your_groovy_scripts to make it easier to do the actual conversion to Kotlin DSL

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/base-ui/build.gradle
+++ b/base-ui/build.gradle
@@ -15,44 +15,44 @@
  */
 
 plugins {
-    id 'com.android.library'
-    id 'kotlin-android'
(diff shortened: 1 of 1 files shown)


oas004:
Rebase with main ☝️

PR #854, opened 2022-12-19
Prepare sample module gradle file for being converted to Kotlin DSL

oas004:
#### WHAT
Prepare the sample module' gradle file to be converted easily to Kotlin DSL

#### WHY
Following up #846  relating to #833 

#### HOW
Following https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html#prepare_your_groovy_scripts

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -14,49 +14,51 @@
  * limitations under the License.
  */
 
-apply plugin: 'com.android.application'
-apply plugin: 'kotlin-android'
-apply plugin: "com.google.protobuf"
-apply plugin: "com.google.android.libraries.mapsplatform.secrets-gradle-plugin"
+plugins {
+    id("com.android.application")
+    id("kotlin-android")
+    id("com.google.protobuf")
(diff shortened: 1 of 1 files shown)


sample/build.gradle
```diff
@@ -14,49 +14,49 @@
  * limitations under the License.
  */
 
-apply plugin: 'com.android.application'
-apply plugin: 'kotlin-android'
+apply plugin: "com.android.application"
+apply plugin: "kotlin-android"
 apply plugin: "com.google.protobuf"
 apply plugin: "com.google.android.libraries.mapsplatform.secrets-gradle-plugin"
```
oas004, line 20: Should I maybe also migrate this to use the plugin lambda?

luizgrp: I think so, as it's strongly recommended in that doc!

oas004: Okay, I'll fix it :)

luizgrp (approved):
LGTM

PR #855, opened 2022-12-19
Prepare tiles gradle file to be converted to Kotlin DSL

oas004:
#### WHAT
Prepare the tiles module' gradle file to be easily converted to Kotlin DSL

#### WHY
Following up #846 related to #833 

#### HOW

Following https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html#prepare_your_groovy_scripts to make it easy to convert the gradle script to Kotlin DSL

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/tiles/build.gradle
+++ b/tiles/build.gradle
@@ -15,31 +15,31 @@
  */
 
 plugins {
-    id 'com.android.library'
-    id 'kotlin-android'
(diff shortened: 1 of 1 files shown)


luizgrp (approved):
LGTM

PR #858, opened 2022-12-19
Convert audio module gradle from Groovy DSL to Kotlin DSL

oas004:
#### WHAT
Convert the gradle file in the `:audio` module to Kotlin DSL

#### WHY
Ref. #833 

#### HOW
Changing the file name and then converting it.

Some of the features here are incubating, is that okay? 

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/audio/build.gradle.kts
+++ b/audio/build.gradle.kts
@@ -19,10 +19,11 @@ plugins {
     id("kotlin-android")
     id("org.jetbrains.dokka")
     id("org.jetbrains.kotlin.kapt")
+    id("me.tylerbwong.gradle.metalava")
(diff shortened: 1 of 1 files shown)


yschimke:
Yep, they are already incubating in groovy, just not flagged.  And most remain incubating in 7.4.

targetSdk is the one exception, we can move that to the manifest

```
    <uses-sdk android:minSdkVersion="25" android:targetSdkVersion="30" tools:ignore="GradleOverrides,OldTargetApi" />
```

yschimke (approved):

PR #861, opened 2022-12-20
Convert audio-ui module gradle file from Groovy DSL to Kotlin DSL

oas004:
#### WHAT
Convert the `:audio-ui` module' gradle file to Kotlin DSL from Groovy DSL.

#### WHY

#### HOW
Converting the file by changing the filename to use `.kts` then converting the content. Then syncing and running all the required gradle tasks. 

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/audio-ui/build.gradle.kts
+++ b/audio-ui/build.gradle.kts
@@ -20,10 +20,11 @@ plugins {
     id("org.jetbrains.dokka")
     id("org.jetbrains.kotlin.kapt")
     id("dev.chrisbanes.paparazzi")
+    id("me.tylerbwong.gradle.metalava")
 }
 
 android {
-    compileSdkVersion = 33
+    compileSdk = 33
 
     defaultConfig {
         minSdk = 25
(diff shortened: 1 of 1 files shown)


audio-ui/build.gradle.kts
```diff
@@ -50,37 +51,36 @@ android {
...
+    sourceSets.getByName("androidTest") {
+        java.srcDir("src/sharedTest/kotlin")
+        res.srcDir("src/sharedTest/res")
+    }
+
+    lintOptions {
+        disable("MissingTranslation")
+        disable("ExtraTranslation")
+        isCheckReleaseBuilds = false
         textReport = true
```
oas004, line 83: Is this correct? :)

audio-ui/build.gradle.kts
```diff
@@ -50,37 +51,36 @@ android {
...
+    sourceSets.getByName("test") {
+        java.srcDir("src/sharedTest/kotlin")
+        res.srcDirs("src/sharedTest/res")
     }
-    lint {
-        checkReleaseBuilds = false
+
+    sourceSets.getByName("androidTest") {
+        java.srcDir("src/sharedTest/kotlin")
+        res.srcDir("src/sharedTest/res")
```
luizgrp, line 76: nit: we don't have those `sharedTest` folders in any module so they can be ignored while porting

audio-ui/build.gradle.kts
```diff
@@ -50,37 +51,36 @@ android {
     }
     packagingOptions {
         resources {
-            excludes += [
-                "/META-INF/AL2.0",
-                "/META-INF/LGPL2.1"
-            ]
+            excludes.addAll(listOf("/META-INF/AL2.0", "/META-INF/LGPL2.1"))
```
luizgrp, line 54: nit: `excludes += listOf` should be the same

audio-ui/build.gradle.kts
oas004: I think this is deprecated. Can I do something like this

```
with(lint) {
        disable.addAll(kotlin.collections.listOf("MissingTranslation", "ExtraTranslation"))
        checkReleaseBuilds = false
        textReport = true
 }
```

Instead?

luizgrp: You can check any of these implementations removing any string from `res/values-ar/strings.xml` and running gradle task `:audio-ui:lintDebug`. It's expected to not fail with missing translation! 

Then comment the and line where `MissingTranslation` is added to `disable` and run `:audio-ui:lintDebug` again. This time it should fail.

oas004: That is a great tip! Thank you! :) I tested it out with 

```
lint {
  disable.addAll(kotlin.collections.listOf("MissingTranslation", "ExtraTranslation"))
  checkReleaseBuilds = false
  textReport = true
 }
```

And it fails without the disabled section if I have commented out one of the translations :)

audio-ui/build.gradle.kts
oas004: Aha okay that makes sense! Thanks!  :)

audio-ui/build.gradle.kts
oas004: Aha alright, so I can remove the androidTest source sets section? :)

luizgrp: and the `test` too!

PR #862, opened 2022-12-20
Convert auth-composable module from Groovy DSL to Kotlin DSL

oas004:
#### WHAT
Converting the `:auth-composable` module' gradle file from Groovy to Kotlin DSL

#### WHY

#### HOW
Converting the file to `.kts` then converting the content and checking if it syncs. Then running all the required gradle tasks.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
LGTM

PR #864, opened 2022-12-21
Convert auth-data module gradle file to Kotlin DSL

oas004:
#### WHAT
Convert the `:auth-data` module' gradle file to Kotlin DSL.

#### WHY

#### HOW
Converting the file to`.kts` then converting the content. Making sure that it syncs properly and that the required gradle tasks pass.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth-data/build.gradle.kts
+++ b/auth-data/build.gradle.kts
@@ -15,14 +15,15 @@
  */
 
 plugins {
-    id ("com.android.library")
-    id ("kotlin-android")
-    id ("org.jetbrains.dokka")
-    id ("com.google.devtools.ksp")
+    id("com.android.library")
+    id("kotlin-android")
+    id("org.jetbrains.dokka")
+    id("com.google.devtools.ksp")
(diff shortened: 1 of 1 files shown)


auth-data/build.gradle.kts
```diff
@@ -15,14 +15,15 @@
...
 plugins {
-    id ("com.android.library")
-    id ("kotlin-android")
-    id ("org.jetbrains.dokka")
-    id ("com.google.devtools.ksp")
+    id("com.android.library")
+    id("kotlin-android")
+    id("org.jetbrains.dokka")
+    id("com.google.devtools.ksp")
+    id("me.tylerbwong.gradle.metalava")
```
yschimke, line 22: Per https://github.com/google/horologist/pull/865/files, do we need this and apply below?

oas004: Good point! I will remove the one below :)

luizgrp (approved):
LGTM

PR #865, opened 2022-12-21
Convert auth-ui module gradle file to Kotlin DSL

oas004:
#### WHAT
Convert the `:auth-ui` module' gradle file to Kotlin DSL

#### WHY

#### HOW
Converting the file to `.kts` then converting the content. Then checking that it syncs correctly and that the required gradle tasks passes.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth-ui/build.gradle.kts
+++ b/auth-ui/build.gradle.kts
@@ -19,10 +19,11 @@ plugins {
     id("kotlin-android")
     id("org.jetbrains.dokka")
     id("dev.chrisbanes.paparazzi")
+    id("me.tylerbwong.gradle.metalava")
 }
 
 android {
-    compileSdkVersion = 33
+    compileSdk = 33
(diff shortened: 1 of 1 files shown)


yschimke (approved):

auth-ui/build.gradle.kts
```diff
@@ -19,10 +19,11 @@ plugins {
     id("kotlin-android")
     id("org.jetbrains.dokka")
     id("dev.chrisbanes.paparazzi")
+    id("me.tylerbwong.gradle.metalava")
```
yschimke, line 22: Doesn't this apply as well? Do we need this and below?

oas004: Good point! I can remove the one below. Thanks! :)

PR #867, opened 2022-12-21
Convert composables module gradle file to Kotlin DSL

oas004:
#### WHAT
Convert the `:composables` module' gradle file to Kotlin DSL

#### WHY

#### HOW
Converting the file type to `.kts` then converting the content. Then checking that it syncs correctly and that the gradle tasks that is required are passing.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
LGTM

PR #868, opened 2022-12-21
Convert compose-layout module gradle file to Kotlin DSL

oas004:
#### WHAT
Convert the `:compose-layout` gradle file to Kotlin DSL

#### WHY
Following up #846  relating to #833 

#### HOW
Converting the filetype to `.kts` then converting the content. Then further, checking that it syncs correctly and that the required gradle tasks are passing.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/compose-layout/build.gradle.kts
+++ b/compose-layout/build.gradle.kts
@@ -19,10 +19,11 @@ plugins {
     id("kotlin-android")
     id("org.jetbrains.dokka")
     id("org.jetbrains.kotlin.kapt")
+    id("me.tylerbwong.gradle.metalava")
 }
 
 android {
-    compileSdkVersion = 33
+    compileSdk = 33
 
     defaultConfig {
         minSdk = 25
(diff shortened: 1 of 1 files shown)


compose-layout/build.gradle.kts
```diff
@@ -49,31 +50,20 @@ android {
...
-    sourceSets {
-        test {
-            java.srcDirs += "src/sharedTest/kotlin"
-            res.srcDirs += "src/sharedTest/res"
-        }
-        androidTest {
-            java.srcDirs += "src/sharedTest/kotlin"
-            res.srcDirs += "src/sharedTest/res"
-        }
-    }
```
oas004, line 76: I understand it from @luizgrp that we do not use these sharedTest and that I can remove them when converting. Please let me know if that is not correct :)

luizgrp (approved):
LGTM

PR #870, opened 2022-12-21
Convert datalayer module gradle file to Kotlin DSL

oas004:
#### WHAT
Convert the `:datalayer` module' gradle file to Kotlin DSL

#### WHY

#### HOW
Converting the file to `.kts` then converting the content. Then checking that the project syncs and that the required gradle tasks passes.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/build.gradle.kts
+++ b/datalayer/build.gradle.kts
@@ -21,7 +21,7 @@ plugins {
 }
 
 android {
-    compileSdkVersion = 33
+    compileSdk = 33
 
     defaultConfig {
         minSdk = 21
@@ -47,25 +47,25 @@ android {
     }
     packagingOptions {
         resources {
-            excludes += [
+            excludes += listOf(
                 "/META-INF/AL2.0",
                 "/META-INF/LGPL2.1"
-            ]
+            )
         }
     }
-
 
     testOptions {
         unitTests {
-            includeAndroidResources = true
+            isIncludeAndroidResources = true
         }
     }
+
     lint {
         checkReleaseBuilds = false
         textReport = true
-
-        baseline(file("quality/lint/lint-baseline.xml"))
+        baseline = file("quality/lint/lint-baseline.xml")
     }
+
     namespace = "com.google.android.horologist.datalayer"
 }
 
@@ -91,4 +91,4 @@ dependencies {
     androidTestImplementation(libs.truth)
 }
 
-apply plugin: "com.vanniktech.maven.publish"
+apply(plugin = "com.vanniktech.maven.publish")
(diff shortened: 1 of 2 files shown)


datalayer/build.gradle.kts
```diff
@@ -47,25 +47,25 @@ android {
...
+            isIncludeAndroidResources = true
         }
     }
+
     lint {
         checkReleaseBuilds = false
         textReport = true
-
-        baseline(file("quality/lint/lint-baseline.xml"))
+        baseline = file("quality/lint/lint-baseline.xml")
```
oas004, line 66: I think this was not the correct way to provide a baseline 😅

oas004: But it works locally though 🤔 That was strange

luizgrp: 🤔 try:

```kotlin
baseline = file("${project.path}/quality/lint/lint-baseline.xml")
```

oas004: Good idea! Synced locally, testing it now :)

luizgrp: oops, meant to be `project.projectDir` instead of `project.path`

oas004: Aha okay makes sense! Trying that :) Thanks! :)

yschimke (approved):

datalayer/build.gradle.kts
luizgrp:
```
> Task :datalayer:lintDebug FAILED
/home/runner/work/horologist/horologist/datalayer/quality/lint/lint-baseline.xml: Information: 1 errors/warnings were listed in the baseline file (quality/lint/lint-baseline.xml) but not found in the project; perhaps they have been fixed? Another possible explanation is that lint recently stopped analyzing (and including results from) dependent projects by default. You can turn this back on with android.lintOptions.checkDependencies=true. Unmatched issue types: ExpiredTargetSdkVersion [LintBaseline]
```

After reading the error message again and now paying attention to what is saying, I don't think it's an issue with the location of the baseline file, so `baseline = file("quality/lint/lint-baseline.xml")` should be fine.

The problem is with the file name that has been changed from `build.gradle` to `build.gradle.kts` so it should also be updated in the [lint-baseline](https://github.com/google/horologist/blob/a16cb844027bd681807186ece6cc6fd416b4d12f/datalayer/quality/lint/lint-baseline.xml#L10) file.

It's odd that the `:datalayer:lintDebug:` task is passing locally in your machine.

If you still find this PR failing after doing this change, it might be worth just deleting `lint-baseline.xml` file and running `:datalayer:lintDebug:` task to generate it again.

oas004: Aha okay, I tried locally with and without the baseline change and it passes. Will try on CI with the baseline change :) Thanks for the help! :)

oas004: Seems like that worked! Thanks for the help! :)

PR #871, opened 2022-12-21
Convert the media module gradle file to Kotlin DSL

oas004:
#### WHAT
Convert the `:media` module' gradle file to Kotlin DSL.

#### WHY

#### HOW
Converting the file type to `.kts` then converting the content. Then checking that it syncs and that the required gradle tasks are passing.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
LGTM

PR #872, opened 2022-12-21
Convert media-benchmark module gradle file to Kotlin DSL

oas004:
#### WHAT
Convert the `:media-benchmark` module gradle file to Kotlin DSL.

#### WHY

#### HOW
Converting the file to `.kts` then converting the content and checking that it syncs and that the required gradle tasks are passing.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media-benchmark/build.gradle.kts
+++ b/media-benchmark/build.gradle.kts
@@ -17,10 +17,11 @@
 plugins {
     id("com.android.library")
     id("kotlin-android")
+    id("me.tylerbwong.gradle.metalava")
 }
 
 android {
-    compileSdkVersion = 33
+    compileSdk = 33
 
     defaultConfig {
         minSdk = 28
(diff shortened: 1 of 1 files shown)


media-benchmark/build.gradle.kts
```diff
@@ -41,46 +42,52 @@ android {
...
 }
 
 dependencies {
-    api(projectOrDependency(":media-lib-session", libs.androidx.media3.session))
+
+    if (project.findProject(":media-lib-session") != null) {
+        api(project(":media-lib-session"))
+    } else {
+        api(libs.androidx.media3.session)
+    }
```
oas004, line 89: I had some issues interopping with this function in the project gradle since that is still in groovy 🤔 Should I make an extension function for this?

oas004: I noticed that this is used in some other modules as well, I think trying to interop with the groovy project file is a better approach 🤔

luizgrp: maybe porting that `projectOrDependency` to kotlin?

I think it can be inlined as it is:
```kotlin
api(project.findProject(":media-lib-session")?: libs.androidx.media3.session)
```

happy to have it addressed in another pr

luizgrp (approved):
LGTM

media-benchmark/build.gradle.kts
oas004: Maybe I can inline it until I have converted the project gradle to Kotlin DSL and then make a PR changing it on all places? :)

PR #880, opened 2022-12-22
Convert paparazzi module to Kotlin DSL

oas004:
#### WHAT
Converting the `:paparazzi` module gradle file to Kotlin DSL.

#### WHY
Following up #846 related to #833 

#### HOW
This module did not have anything that was different from the groovy prep phase, so here I only renamed the files. Then checked if it synced and that the required gradle tasks where passing.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Surprising but nice.

PR #889, opened 2022-12-26
Convert project gradle file to Kotlin DSL

### LINKED ISSUE #833: Convert Groovy gradle files to Kotlin DSL

oas004:
#### WHAT
Converting the project gradle file to Kotlin DSL.

#### WHY
Following up #846 relating to #833. This resolves #833 

#### HOW
Converting the file type to `.kts`. Then converting the content and checking that it syncs and that the required gradle tasks are passing. 

In this file, there was a lot I was not 100% sure about. Leaving this as a draft until I have made sure that everything is working correctly as expected. I think also I need some help with the `mavenPublish`stuff. This was not very good documented for Kotlin DSL, and I don't quite know how to test it. 

I am also a bit unsure if I have correctly converted the `dependencyUpdate` stuff.

Before there was some logic to check the plugin manager if there is a version catalog update plugin before we apply it in the project gradle file. That did not quite work as the `versionCatalogUpdate` lambda block would become unresolved.  So now I am applying it, but I don't know if that is the correct solution here. Any input would be highly appreciated :)  

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


build.gradle.kts
```diff
@@ -286,23 +307,19 @@ subprojects {
...
-        (it.candidate.version ==~ /.*-alpha.*/ && !(it.currentVersion ==~ /.*-alpha.*/)) ||
-          (it.candidate.version ==~ /.*-beta.*/ && !(it.currentVersion ==~ /.*-(beta|alpha).*/))
+        (candidate.version.matches("/.*-alpha.*/".toRegex()) && !(currentVersion.matches("/.*-alpha.*/".toRegex()))) ||
+            (candidate.version.matches("/.*-beta.*/".toRegex()) && !(currentVersion.matches("/.*-(beta|alpha).*/".toRegex())))
     }
 }
 
-if (!getPluginManager().hasPlugin("nl.littlerobots.version-catalog-update")) {
-    apply(plugin: "nl.littlerobots.version-catalog-update")
+if (!pluginManager.hasPlugin("nl.littlerobots.version-catalog-update")) {
```
yschimke, line 317: You can remove the if, and just apply it normally. It's there because I had thus already applied in a global config. But I'll remove that.

build.gradle.kts
```diff
@@ -286,23 +307,19 @@ subprojects {
...
-def projectOrDependency(module, dependency = null) {
-    return findProject(module) ?: dependency
-}
-
-tasks.named("dependencyUpdates").configure {
+tasks.withType<com.github.benmanes.gradle.versions.updates.DependencyUpdatesTask> {
     rejectVersionIf {
-        (it.candidate.version ==~ /.*-alpha.*/ && !(it.currentVersion ==~ /.*-alpha.*/)) ||
-          (it.candidate.version ==~ /.*-beta.*/ && !(it.currentVersion ==~ /.*-(beta|alpha).*/))
+        (candidate.version.matches("/.*-alpha.*/".toRegex()) && !(currentVersion.matches("/.*-alpha.*/".toRegex()))) ||
```
yschimke, line 312: I think drop the surrounding /

build.gradle.kts
```diff
@@ -93,22 +98,22 @@ allprojects {
...
+            maven(url = uri("https://androidx.dev/snapshots/builds/$composeSnapshot/artifacts/repository/"))
         }
     }
 
     plugins.withId("com.vanniktech.maven.publish") {
-        mavenPublish {
-            androidVariantToPublish = "release"
-            sonatypeHost = "DEFAULT"
+        mavenPublishing {
+//            androidVariantToPublish.set("release")
```
yschimke, line 109: I think we will need this

yschimke:
Don't worry about the dependency update parts. Can fix next time we run it. 

It runs as part of

`./gradlew versionCatalogUpdate`

build.gradle.kts
yschimke: It may be ANDROID_VARIANT_TO_PUBLISH, but also possible it's no longer required.

I suggest landing it and I'll ty a release and then drop it.

yschimke (approved):

build.gradle.kts
oas004: Alright, I removed the surrounding if statement :)

build.gradle.kts
oas004: I might be mistaken here, but according to the release notes (https://github.com/vanniktech/gradle-maven-publish-plugin/blob/master/CHANGELOG.md) the release version will be chosen by default. I could not quite find how to set the release `ANDROID_VARIANT_TO_PUBLISH`. I think that it might be a bit redundant to add 🤔 

```
Unless another variant is specified by setting the ANDROID_VARIANT_TO_PUBLISH Gradle property the release variant will be published.
```

yschimke: That is in the release notes for 0.21.0. We are on 0.20.0.

> Since version 0.19.0 the plugin was publishing a multi variant library by default for Android projects. Due to [a bug in Android Studio](https://issuetracker.google.com/issues/197636221) that will cause it to not find the sources for libraries published this way the plugin will temporarily revert to publishing single variant libraries again. Unless another variant is specified by setting the ANDROID_VARIANT_TO_PUBLISH Gradle property the release variant will be published.

I don't suggest updating to 0.22.0 in this PR.  So suggest adding that entry to `gradle.properties`

yschimke: Confirmed with `./gradlew publishToMavenLocal`.  You should be able to test locally with this in gradle.properties

```
RELEASE_SIGNING_ENABLED=false
ANDROID_VARIANT_TO_PUBLISH=release
```


```
$ ls ~/.m2/repository/com/google/android/horologist/horologist-audio/0.2.6/horologist-audio-0.2.6
horologist-audio-0.2.6-debug-javadoc.jar        horologist-audio-0.2.6-release-javadoc.jar      horologist-audio-0.2.6.module                 
horologist-audio-0.2.6-debug-javadoc.jar.asc    horologist-audio-0.2.6-release-javadoc.jar.asc  horologist-audio-0.2.6.module.asc             
horologist-audio-0.2.6-debug-sources.jar        horologist-audio-0.2.6-release-sources.jar      horologist-audio-0.2.6.pom                    
horologist-audio-0.2.6-debug-sources.jar.asc    horologist-audio-0.2.6-release-sources.jar.asc  horologist-audio-0.2.6.pom.asc                
horologist-audio-0.2.6-debug.aar                horologist-audio-0.2.6-release.aar                                                            
horologist-audio-0.2.6-debug.aar.asc            horologist-audio-0.2.6-release.aar.asc                   
```

build.gradle.kts
```diff
@@ -93,22 +98,21 @@ allprojects {
...
+            maven(url = uri("https://androidx.dev/snapshots/builds/$composeSnapshot/artifacts/repository/"))
         }
     }
 
     plugins.withId("com.vanniktech.maven.publish") {
-        mavenPublish {
-            androidVariantToPublish = "release"
-            sonatypeHost = "DEFAULT"
+        mavenPublishing {
+            publishToMavenCentral(SonatypeHost.DEFAULT)
```
yschimke, line 109: Should we remove this from gradle.properties?

```
SONATYPE_HOST=DEFAULT
```

build.gradle.kts
oas004: Oh, you are totally right! Thanks for catching this! And for showing me how to verify it! Added the `ANDROID_VARIANT_TO_PUBLISH=release` to the gradle.properties file now :)

build.gradle.kts
oas004: Good call! Removed it now :) Thanks!

Issue #892, opened 2022-12-29
Add resourcePrefix to Android modules


luizgrp:
#### WHAT

Set [resourcePrefix](https://cs.android.com/android-studio/platform/tools/base/+/mirror-goog-studio-main:build-system/gradle-api/src/main/java/com/android/build/api/dsl/CommonExtension.kt;l=633?q=CommonExtension) in the Android library modules of the project - unpublished modules and modules for sample apps are not needed.

#### WHY

Avoid naming collisions with similar resources that the consumer defines.

#### HOW

- Add `resourcePrefix = "horologist_"` to `build.gradle.kts`;
- Fix resource names;

yschimke:
Could this break translations?

RotBolt:
Can I pick this up to work or contribute to?

luizgrp:
@RotBolt that would be great!

RotBolt:
I will get started this weekend

yschimke:
I think we have this

luizgrp:
We only have this for newer modules. But happy to close it as low priority.

PR #895, opened 2022-12-29
Add maven publish plugin to auth modules

luizgrp:
#### WHAT

Add maven publish plugin to auth modules.

#### WHY

So they are published to maven in the next release for early API feedback.

#### HOW

- Uncomment line to apply maven publish plugin;
- Add missing configuration files;
- Unrelated: add [resourcePrefix](https://github.com/google/horologist/issues/892);

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth-composables/build.gradle.kts
+++ b/auth-composables/build.gradle.kts
@@ -75,6 +75,8 @@ android {
         textReport = true
     }
 
+    resourcePrefix = "horologist_"
+
     namespace = "com.google.android.horologist.auth.composables"
 }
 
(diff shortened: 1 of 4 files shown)


auth-composables/build.gradle.kts
```diff
@@ -75,6 +75,8 @@ android {
         textReport = true
     }
 
+    resourcePrefix = "horologist_"
```
yschimke, line 78: Why this change?

yschimke (approved):

auth-composables/build.gradle.kts
luizgrp: https://github.com/google/horologist/issues/892

PR #903, opened 2023-01-03
Fix unstable lambdas

fstanis:
#### WHAT
Changes lambdas and `PlayerUiController` to be stable (by compose definition).

#### WHY
This reduces the number of recompositions, improving performance.

#### HOW
By passing methods directly instead of creating lambdas and explicitly marking `PlayerUiController` as `@Stable`.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


luizgrp (approved):
nice one!

PR #914, opened 2023-01-09
Makes onRotaryInputAccumulated customizable

fstanis:
#### WHAT
Makes it possible to fine-tune `onRotaryInputAccumulated` thresholds.

#### WHY
To make it possible to customize rotary behavior. For example, volume controls by default use a different threshold to reduce sensitivity.

#### HOW
Moves `RotaryInputAccumulator` constructor parameters into `onRotaryInputAccumulated`.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


clickxu (approved):
LGTM

PR #922, opened 2023-01-16
Standardise the use of ScalingLazyColumn and fillMaxSize

yschimke:
#### WHAT

Standardise the use of ScalingLazyColumn and fillMaxSize.

#### WHY

Problems with ScalingLazyColumn width only shows up with narrow content, so apply preemptively.  

#### HOW

Add a modifier.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
thanks!

PR #928, opened 2023-01-18
Upgrade Android Gradle plugin

luizgrp:
#### WHAT

Upgrade Android Gradle plugin.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):

yschimke:
Can follow up, and you might be working through it, but I think we should make sure we are on 11 everywhere also.

https://github.com/android/nowinandroid/pull/531/files

yschimke:
Nice!

PR #930, opened 2023-01-18
Exposes playback speed method

kul3r4:
#### WHAT
Currently playback speed can be set but cannot be read from using the interface

#### WHY
So that the playback speed can be read and the value can be represented in the UI

#### HOW
Add a get method to the interface

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/media/src/main/java/com/google/android/horologist/media/repository/PlayerRepository.kt
+++ b/media/src/main/java/com/google/android/horologist/media/repository/PlayerRepository.kt
@@ -68,6 +68,11 @@ public interface PlayerRepository {
     public val seekForwardIncrement: StateFlow<Duration?>
 
     /**
+     * Returns the current [playbackSpeed].
+     */
+    public val playbackSpeed: StateFlow<Float>
+
+    /**
      * Prepares the player. E.g. player will start acquiring all the required resources to play.
      */
     public fun prepare()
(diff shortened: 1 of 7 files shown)


media/src/main/java/com/google/android/horologist/media/repository/PlayerRepository.kt
```diff
@@ -198,4 +198,9 @@ public interface PlayerRepository {
      * Set the playback speed.
      */
     public fun setPlaybackSpeed(speed: Float)
+
+    /**
+     * Returns the current playback speed.
+     */
+    public fun getPlaybackSpeed(): Float
```
luizgrp, line 205: what about adding [playbackSpeed](https://github.com/google/horologist/blob/8482a1f297a5d3c95151b5c2ff3f65d0f934a221/media-data/src/main/java/com/google/android/horologist/media/data/repository/PlayerRepositoryImpl.kt#L89) property to PlayerRepository like we do with these [other](https://github.com/google/horologist/blob/8482a1f297a5d3c95151b5c2ff3f65d0f934a221/media/src/main/java/com/google/android/horologist/media/repository/PlayerRepository.kt#L33-L68) properties?

kul3r4: Thank you, I updated the code

media/src/main/java/com/google/android/horologist/media/repository/PlayerRepository.kt
```diff
@@ -67,6 +67,11 @@ public interface PlayerRepository {
      */
     public val seekForwardIncrement: StateFlow<Duration?>
 
+    /**
+     * Returns the current [playbackSpeed].
+     */
+    public val playbackSpeed: StateFlow<Float?>
```
luizgrp, line 73: do we want to allow `null` to be returned? the implementation is defined as non-nullable (`StateFlow<Float>`), just wondering if there is a use case for this

kul3r4: agree, removed null

luizgrp (approved):
LGTM!

PR #934, opened 2023-01-19
Add Auth Sample wear app

luizgrp:
#### WHAT

Add Auth Sample wear app.

#### WHY

In order to be separate from the samples app.

#### HOW

Add `auth-sample-wear` module and move files from `sample` module.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/di/SampleAppDI.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import com.google.android.horologist.auth.sample.SampleApplication
+import com.squareup.moshi.Moshi
+import okhttp3.OkHttpClient
+import okhttp3.logging.HttpLoggingInterceptor
+
+/**
+ * Simple DI implementation.
+ */
+object SampleAppDI {
```
yschimke, line 27: Sorry about this DI pattern.  This must be worse than Hilt, or Koin.  Maybe we should just adopt Koin for samples apart from media-sample?

yschimke (approved):

auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/di/SampleAppDI.kt
luizgrp: will come back to this and iterate 👍

PR #937, opened 2023-01-20
Use releaseCompileOnly scope for compose.tools

yschimke:
#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
nice! 

maybe we should improve the docs of `compose-tools` to provide guidance on this?

PR #939, opened 2023-01-20
Revamps PlayerState to be less tied to media3

fstanis:
#### WHAT
Changes `PlayerState` to be simpler (`Ended` removed) and less tied to media3 internal state.

#### WHY
To make it easier to reason around and make non-standard players easier to implement.

#### HOW
See https://github.com/fstanis/horologist/blob/newstate/media-data/src/main/java/com/google/android/horologist/media/data/mapper/PlayerStateMapper.kt

The new mapping is such that:
- `PlayerState.Playing` is guaranteed to match `isPlaying`
- `PlayerState.Stopped` means pressing play is possible and should work
- `PlayerState.Idle` can now only happen when the player isn't yet initialized (e.g. empty timeline)
- `PlayerState.Loading`is largely unchaged, matching the buffering state

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


yschimke (approved):
Seems much more intuitive reading it through.

media-data/src/main/java/com/google/android/horologist/media/data/mapper/PlayerStateMapper.kt
```diff
@@ -25,26 +25,29 @@ import com.google.android.horologist.media.model.PlayerState
...
-            ) &&
-            player.playWhenReady
-        ) {
-            PlayerState.Playing
-        } else if (player.isLoading) {
-            PlayerState.Loading
-        } else {
-            map(player.playbackState)
-        }
+    public fun map(player: Player): PlayerState = when (player.playbackState) {
```
luizgrp, line 28: do you think it's worth improving test coverage in `PlayerStateMapperTest` or `PlayerStateMapperTest` providing enough for this class?

luizgrp (approved):
this is a great improvement! thanks!

media-data/src/main/java/com/google/android/horologist/media/data/mapper/PlayerStateMapper.kt
fstanis: Added another test - in practice, I like the coverage provided by `PlayerRepositoryImplTest` because of how important it is to test this together with the `play()` command.

PR #941, opened 2023-01-23
Add artworkColor field

kpaikena:
#### WHAT

Add field ‘artworkColor', that represents the dominant color of the album artwork, to Media and MediaUiModel classes.

#### WHY

So that we can use 'artworkColor' field to apply tint to UMO depending on the media playing.

#### HOW

Add 'artworkColor' field to Media and MediaUiModel classes, and to MediaUiModelMapper to convert from Media to MediaUiModel.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/MediaUiModel.kt
```diff
@@ -16,12 +16,14 @@
...
 import com.google.android.horologist.media.ui.ExperimentalHorologistMediaUiApi
 
 @ExperimentalHorologistMediaUiApi
 public data class MediaUiModel(
     val id: String,
     val title: String? = null,
     val subtitle: String? = null,
-    val artworkUri: String? = null
+    val artworkUri: String? = null,
+    @ColorInt val artworkColor: Int? = null
```
yschimke, line 28: Should we use androidx.compose.ui.graphics.Color?

fstanis: +1 since that's the type that `ColorBackground` composable currently accepts.

kpaikena: We could use androidx.compose.ui.graphics.Color for MediaUiModel but I think Media class doesn't have that dependency since it is in Java library? So we would have to convert from Int to Color in MediaUiModelMapper.

fstanis: > So we would have to convert from Int to Color in MediaUiModelMapper.

Yes, I think that's the correct place to do it - the data in the model should be as close as possible to what we'll render with Compose components and they take `androidx.compose.ui.graphics.Color` so this should be that type and the `MediaUiModelMapper` would produce it.

media/src/main/java/com/google/android/horologist/media/model/Media.kt
```diff
@@ -28,5 +28,6 @@ public data class Media(
     val title: String,
     val artist: String,
     val artworkUri: String? = null,
+    val artworkColor: Int? = null,
```
yschimke, line 31: If this is staying, should it have `@ColorInt`

luizgrp: this class is in java module and shouldn't have android dependencies

kpaikena: Added @ColorInt annotation.

media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/MediaUiModel.kt
kpaikena: Changed type in `MediaUiModel` to `androidx.compose.ui.graphics.Color`.

yschimke (approved):

PR #946, opened 2023-01-24
API Fixes

yschimke:
#### WHAT

API Fixes

#### WHY

Seemingly broken in previous diff.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
> metalava didn't flag these differences?

I'm not sure.  it generates the file, the checks pass. Something feels wrong.

PR #950, opened 2023-01-25
Define Android Studio baseline

yschimke:
#### WHAT

Define Android Studio baseline

#### WHY

Need it stated what we support so we have a line to separate issues to fix.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ an issue https://github.com/google/horologist/issues with the label "Feature Req
 
 ## Development
 
-The project should work immediately from a fresh checkout in Android Studio (Electric Eel or newer) or or with Gradle (./gradlew).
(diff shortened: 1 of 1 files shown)


luizgrp (approved):

CONTRIBUTING.md
```diff
@@ -9,7 +9,7 @@ an issue https://github.com/google/horologist/issues with the label "Feature Req
 
 ## Development
 
-The project should work immediately from a fresh checkout in Android Studio or with Gradle.
+The project should work immediately from a fresh checkout in Android Studio (Electric Eel or newer) or Gradle (./gradlew).
```
luizgrp, line 12: should we make it more dynamic saying "stable" version and maybe place a link to https://developer.android.com/studio/releases?

PR #951, opened 2023-01-25
Fix kotlin android plugin name

yschimke:
#### WHAT

Fix kotin android plugin name

#### WHY

AS EE suggested this, and seems to be the full name.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/audio-ui/build.gradle.kts
+++ b/audio-ui/build.gradle.kts
@@ -16,11 +16,11 @@
 
 plugins {
     id("com.android.library")
-    id("kotlin-android")
     id("org.jetbrains.dokka")
     id("org.jetbrains.kotlin.kapt")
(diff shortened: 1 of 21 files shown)


audio-ui/build.gradle.kts
```diff
@@ -115,6 +115,7 @@ dependencies {
     implementation(libs.lottie.compose)
 
     implementation(libs.compose.ui.toolingpreview)
+    implementation("androidx.core:core-ktx:+")
```
yschimke, line 118: Ooops

yschimke:
Discussion here https://stackoverflow.com/questions/69297003/are-kotlin-android-and-org-jetbrains-kotlin-android-the-same

PR #953, opened 2023-01-26
Add lib for token sharing between wear and phone

luizgrp:
#### WHAT

Add lib for token sharing between wear and phone.

#### WHY

Provide default implementation for recommended authentication method: [Pass tokens using the data layer](https://developer.android.com/training/wearables/apps/auth-wear#tokens).

#### HOW

- Add `auth-data-phone` module and provide `TokenRepository` to be used to send token from the phone app;
- Add `TokenListenerService` to `auth-data` module to be used to receive token from the watch app;
- Add `auth-sample-phone` with a phone auth sample app to share token with the wear auth sample app;
- Update `auth-sample-wear` to use `TokenListenerService`;

Apologies for the large size PR. Improvements to the phone sample app will be added in a separate PR.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


auth-data/src/main/java/com/google/android/horologist/auth/data/tokenshare/TokenListenerService.kt
```diff
@@ -0,0 +1,79 @@
...
+ *   android:exported="true">
+ *   <intent-filter>
+ *     <action android:name="com.google.android.gms.wearable.DATA_CHANGED" />
+ *     <data android:scheme="wear" android:host="*" android:pathPrefix="/horologist_auth" />
+ *   </intent-filter>
+ * </service>
+ * ```
+ */
+@ExperimentalHorologistAuthDataApi
+public abstract class TokenListenerService : WearableListenerService() {
```
yschimke, line 47: How often do you think the app needs to be woken up on a new token? WearableListenerService is presumably a cost in terms of accounting.  If the mobile app updates the token routinely (maybe a scheduled job, and < 30 minute token refresh), then our wear app will wake up a lot also.

If mostly a fresh token is needed while app is running, that seems simpler.

Maybe this is just documentation to advise against implementing TokenListenerService unless you really need it.

auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenRepositoryImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+ */
+@ExperimentalHorologistAuthDataPhoneApi
+public class TokenRepositoryImpl(private val dataClient: DataClient) : TokenRepository {
+
+    @SuppressLint("VisibleForTests") // https://issuetracker.google.com/issues/239451111
+    override suspend fun updateToken(token: String) {
+        val putDataReq: PutDataRequest = PutDataMapRequest.create(SERVICE_PATH_PREFIX).run {
+            dataMap.putString(KEY_TOKEN, token)
+            asPutDataRequest()
+        }
```
yschimke, line 44:
```suggestion
        val putDataReq: PutDataRequest = PutDataMapRequest.create(SERVICE_PATH_PREFIX).apply {
            dataMap.putString(KEY_TOKEN, token)
        }.asPutDataRequest()
```

Maybe personal style, feel free to ignore.

auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenRepositoryImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        if (putDataTask.isSuccessful) {
+            Log.d(TAG, "DataItem updated successfully")
+        } else {
+            putDataTask.exception?.let { exception ->
+                Log.e(TAG, "DataItem update failed with: ${exception.message}", exception)
+            } ?: run {
+                Log.e(
+                    TAG,
+                    "DataItem update failed. [isCancelled: ${putDataTask.isCanceled}] [isComplete: ${putDataTask.isComplete}]"
```
yschimke, line 57: Can it have an exception and be complete?

auth-data/src/main/java/com/google/android/horologist/auth/data/tokenshare/TokenListenerService.kt
luizgrp: thanks for this perspective! 

in the next PR then I will add a default implementation for `DataClient.OnDataChangedListener` to be used as option, and improve the documentation of `DataClient.OnDataChangedListener` recommending to use that

auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenRepositoryImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    @SuppressLint("VisibleForTests") // https://issuetracker.google.com/issues/239451111
+    override suspend fun updateToken(token: String) {
+        val putDataReq: PutDataRequest = PutDataMapRequest.create(SERVICE_PATH_PREFIX).run {
+            dataMap.putString(KEY_TOKEN, token)
+            asPutDataRequest()
+        }
+        val putDataTask: Task<DataItem> = dataClient.putDataItem(putDataReq)
+
+        putDataTask.await()
```
yschimke, line 47: Won't any exception be throw here? IIRC await strips Task<T> to T or a thrown exception.

yschimke: https://cs.android.com/android/platform/superproject/+/master:external/kotlinx.coroutines/integration/kotlinx-coroutines-play-services/src/Tasks.kt;l=139?q=kotlinx.coroutines.tasks

auth-data/src/main/java/com/google/android/horologist/auth/data/tokenshare/TokenListenerService.kt
yschimke: Probably for a follow up, it would seem useful to have a way to use this without the service and manifest entries.  But no issue with this as is.

yschimke (approved):

garanj:
This looks great! I have a question:

Do you think that there is the need to support more than one token? If I'm reading it right, then this currently supports one token between phone and watch?

I am wondering, are there scenarios where more than one token might be needed? e.g.
- There is one phone, but multiple watches, which need separate tokens
- There is one phone, one watch, but multiple services being used, so multiple tokens needed?

(I'm only throwing these ideas out there, I don't have concrete examples of either right now)

auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenRepositoryImpl.kt
luizgrp: I like that, thanks!

auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenRepositoryImpl.kt
luizgrp: Reading the [kdocs](https://cs.android.com/android/platform/superproject/+/master:external/kotlinx.coroutines/integration/kotlinx-coroutines-play-services/src/Tasks.kt;l=97-104?q=kotlinx.coroutines.tasks) of the function called it only mentions `CancellationException` which I don't think it should be handled. What are you suggesting?

luizgrp:
@garanj good questions, I guess we could improve the api to associate a key to each token stored.

Will merge this and submit follow up PRs addressing comments and improvements.

auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenRepositoryImpl.kt
luizgrp: https://github.com/google/horologist/pull/955

auth-data/src/main/java/com/google/android/horologist/auth/data/tokenshare/TokenListenerService.kt
luizgrp: https://github.com/google/horologist/pull/955

PR #955, opened 2023-01-26
Improve token sharing solution and samples

luizgrp:
#### WHAT

Improve token sharing solution and samples.

#### HOW

- Add `TokenListener` to `auth-data`;
- Add screens to wear app to display token shared from phone app;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


auth-data/src/main/java/com/google/android/horologist/auth/data/tokenshare/TokenListenerService.kt
```diff
@@ -46,34 +44,22 @@ import kotlinx.coroutines.launch
...
-                    .dataMap.getString(KEY_TOKEN)
-                    ?.let { token ->
-                        getCoroutineScope().launch {
-                            onTokenReceived(token)
-                        }
-                    }
-                    ?: run {
-                        Log.d(TAG, "Token not found in $uri.")
-                    }
+        OnDataChangedHandler.handle(dataEventBuffer) { token ->
```
yschimke, line 59: nice cleanup

auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/screens/tokenshare/listener/TokenShareListenerScreen.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.google.android.horologist.compose.layout.ScalingLazyColumnState
+
+@Composable
+fun TokenShareListenerScreen(
+    columnState: ScalingLazyColumnState,
+    modifier: Modifier = Modifier,
+    viewModel: TokenShareListenerViewModel = viewModel()
+) {
+    val context = LocalContext.current
+    val dataClient = remember { Wearable.getDataClient(context) }
```
yschimke, line 52: I wonder if this is better in the ViewModel?

yschimke (approved):

auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/screens/tokenshare/listener/TokenShareListenerScreen.kt
luizgrp: Yeah I'd like to see it in the VM too. Then we would have to make a call to the VM on onResume and onPause in order to add/remove the listener. Not entirely sure about this pattern of letting the VM know the activity lifecycle events. Might call the functions in the VM with different names. Thoughts? 

Will merge this PR and come back to this matter in another PR.

PR #957, opened 2023-01-27
Add StreamlineSignInScreen

luizgrp:
#### WHAT

Add `StreamlineSignInScreen`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth-sample-wear/src/main/res/values/strings.xml
+++ b/auth-sample-wear/src/main/res/values/strings.xml
@@ -24,7 +24,8 @@
     <string name="auth_menu_oauth_device_grant_sign_in_item">Sign in</string>
     <string name="auth_menu_oauth_device_grant_sign_out_item">Sign out</string>
     <string name="auth_menu_google_sign_in_header">Google Sign-In</string>
-    <string name="auth_menu_google_sign_in_item">Sign in</string>
+    <string name="auth_menu_google_sign_in_prompt_item">Sign-in Prompt</string>
+    <string name="auth_menu_google_streamline_sign_in_item">Streamline Sign-in</string>
     <string name="auth_menu_google_sign_out_item">Sign out</string>
     <string name="auth_menu_token_share_header">Token Sharing</string>
     <string name="auth_menu_token_share_listener_item">Via Listener</string>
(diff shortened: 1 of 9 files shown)


auth-ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/streamline/GoogleStreamlineSignInViewModelFactory.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.google.android.horologist.auth.data.googlesignin.GoogleSignInAuthUserRepository
+import com.google.android.horologist.auth.ui.ExperimentalHorologistAuthUiApi
+import com.google.android.horologist.auth.ui.common.screens.streamline.StreamlineSignInViewModel
+
+/**
+ * A [factory][ViewModelProvider.Factory] to create a [StreamlineSignInViewModel] with dependencies
+ * with implementation for the Google Sign-In authentication method.
+ */
+@ExperimentalHorologistAuthUiApi
+public val GoogleStreamlineSignInViewModelFactory: ViewModelProvider.Factory = viewModelFactory {
```
yschimke, line 31: Why a separate class? You seem to have a val in other places.

    companion object {

        private val TAG = GoogleSignOutViewModel::class.java.simpleName

        val Factory: ViewModelProvider.Factory = viewModelFactory {

auth-sample-wear/src/main/res/values/strings.xml
```diff
@@ -35,4 +36,5 @@
     <string name="google_sign_in_prompt_message">Send messages and create chat groups with your friends</string>
     <string name="google_sign_out_success_message">Signed out successfully!</string>
     <string name="google_sign_in_prompt_already_signed_in_message">Already signed-in!</string>
+    <string name="google_streamline_sign_in_already_signed_in_message">Already signed-in!</string>
```
yschimke, line 39: Is it worth some of these being in the auth-ui module to provide default translations?

yschimke (approved):

auth-ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/streamline/GoogleStreamlineSignInViewModelFactory.kt
luizgrp: It's in a separate class because `StreamlineSignInScreen` can be used with any auth method, not just Google Sign-In. Let me know if it clarifies it.

auth-sample-wear/src/main/res/values/strings.xml
luizgrp: 100%, but these designs/texts were not in the redlines, so I'd rather wait until get UX involved/approval

PR #959, opened 2023-01-27
Cleanup build files

yschimke:
#### WHAT

Cleanup build files. target sdk removed from library modules. Suppress incubating warnings.

#### WHY

Lot of warnings on latest AGP

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


audio-ui/build.gradle.kts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+@file:Suppress("UnstableApiUsage")
```
luizgrp, line 17: thanks for adding this!

luizgrp (approved):

PR #960, opened 2023-01-27
Add AuthUser to onAlreadySignedIn callback

luizgrp:
#### WHAT

Add `AuthUser` to `onAlreadySignedIn` callback.

#### WHY

So it can be used to display information on the screen that is navigating to.

#### HOW

Add snapshot and unit tests;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth-ui/src/test/java/com/google/android/horologist/auth/ui/common/screens/streamline/StreamlineSignInScreenTest.kt
+++ b/auth-ui/src/test/java/com/google/android/horologist/auth/ui/common/screens/streamline/StreamlineSignInScreenTest.kt
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2023 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@file:OptIn(
+    ExperimentalHorologistPaparazziApi::class,
+    ExperimentalHorologistComposeToolsApi::class,
+    ExperimentalHorologistAuthUiApi::class
+)
+
+package com.google.android.horologist.auth.ui.common.screens.streamline
+
+import com.google.android.horologist.auth.data.common.model.AuthUser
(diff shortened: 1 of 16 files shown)


auth-ui/src/test/java/com/google/android/horologist/auth/ui/common/screens/streamline/StreamlineSignInScreenTest.kt
```diff
@@ -0,0 +1,86 @@
...
+                onSignedOut = { }
+            )
+        }
+    }
+
+    @Test
+    fun signedIn() {
+        paparazzi.snapshotInABox {
+            StreamlineSignInScreen(
+                state = StreamlineSignInScreenState.SignedIn(AuthUser()),
```
yschimke, line 67: The screenshots don't seem to match?

yschimke (approved):

auth-ui/src/test/java/com/google/android/horologist/auth/ui/common/screens/streamline/StreamlineSignInScreenTest.kt
luizgrp: the screen only provides the "loading" content, after the SignedIn state is emitted is up to the caller of the screen to navigate to another screen

yschimke: So broken in non interactive previews as well?

luizgrp: hmm I don't think there is anything broken, the screenshots are matching my expectations, there are no other screens to be displayed here because the `onAlreadySignedIn` param on the line below is empty

yschimke: I guess I'm wondering whether it's worth changing the test, even refactoring the Composable API slightly so you can screenshot test each of the states.  4 identical screenshots with "Signing In" seems less valuable that testing the layout of those states.

yschimke: Your call.

luizgrp: The layout of the other states in dynamic, according to what screen the app is redirecting the user to. I could add a layout for the tests and the screenshot will be of that layout. 
However not sure if paparazzi even support that. The composable api is already changed in order to make it testable, hence the internal composable function which takes a state instead of viewmodel. Not sure how further I would have to go to enable paparazzi to take a screenshot of the destination of those callbacks. Happy to hear suggestions, and will come back to it in another point.

PR #962, opened 2023-01-30
Accessibility changes for Pickers

rajat4064g:
#### WHAT
Fixing accessibility in DatePicker and TimePicker.

#### WHY
Accessibility wasn't working properly in Pickers.

#### HOW
Handled Double Tap events appropriately to move from on picker to another and scrolling on the selected picker only irrespective of where it was performed on screen.


composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -107,28 +121,71 @@ public fun TimePicker(
...
-        val focusRequester3 = remember { FocusRequester() }
-        Box(modifier = modifier.fillMaxSize()) {
+        val focusRequesterHours = remember { FocusRequester() }
+        val focusRequesterMinutes = remember { FocusRequester() }
+        val focusRequesterSeconds = remember { FocusRequester() }
+        val focusRequesterConfirmButton = remember { FocusRequester() }
+
+        val hourContentDescription by remember {
+            derivedStateOf {
+                createDescription(focusedElement, hourState.selectedOption, "hours")
```
BowerSteve, line 148: "hours", "minutes", "seconds" should come from string resources too, see content description strings below.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -465,6 +679,126 @@ internal fun PickerWithRSB(
...
+    return accessibilityManager
+}
+
+private fun touchExplorationStateListener(): AccessibilityManager.TouchExplorationStateChangeListener {
+    touchExplorationStateChangeListener =
+        AccessibilityManager.TouchExplorationStateChangeListener { isEnabled ->
+            touchExplorationEnabled = isEnabled
+        }
+    return touchExplorationStateChangeListener
+}
```
BowerSteve, line 759: Looks like these could be shared between DatePicker/TimePicker if made internal.

BowerSteve (changes requested):

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
rajat4064g: `rememberTalkBackState` is made internal as that is required to be shared between files. These private methods are only needed here.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
rajat4064g: Done

BowerSteve (approved):

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -78,7 +83,7 @@ public fun DatePicker(
         verifyDates(date, fromDate, toDate)
     }
 
-    val datePickerState = remember {
+    val datePickerState = remember(date) {
```
yschimke, line 86: This makes sense.  But do we have API guidance on the general class of "Pickers" for how to respond to params changing.

Maybe not relevant in a wear app when used as full screen lifecycle.

But when implementing these pickers, is it clear that all state should be reset when the input changes?

cc @BowerSteve

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -194,26 +290,57 @@ public fun DatePicker(
...
-                listOf(focusRequester1, focusRequester2, focusRequester3)[selectedColumn]
-                    .requestFocus()
+            LaunchedEffect(focusedElement) {
+                if (focusedElement != FocusableElementDatePicker.NONE) {
+                    listOf(
+                        focusRequesterDay,
+                        focusRequesterMonth,
+                        focusRequesterYear,
+                        focusRequesterConfirmButton
+                    )[focusedElement.index].requestFocus()
```
yschimke, line 342: Now that we have focus in Wear Foundation.

I think this should be using androidx.wear.compose.foundation.HierarchicalFocusCoordinator and friends.  If this Picker is in a Pager, it shouldn't request focus here.

yschimke (approved):
LGTM - one existing problem to resolve with focus, can be a follow up.

PR #974, opened 2023-02-01
Fix VolumeViewModel creation

yschimke:
#### WHAT

Fix VolumeViewModel creation

#### WHY

The library methods like VolumeScreen, are super tolerant, creating the ViewModel with viewModel and a factor on the fly.  Instead we should use our ViewModel created from hilt in case it has different configuration.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


laiyichin (approved):
LGTM

PR #976, opened 2023-02-02
Move scroll logic to player screen

yschimke:
#### WHAT

Move scroll logic to player screen

#### WHY

Good suggestion by @laiyichin , no need to be in the scaffold since we now have Focus APIs

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/screens/player/PlayerScreen.kt
```diff
@@ -66,6 +71,7 @@ public typealias PlayerBackground = @Composable BoxScope.(playerUiState: PlayerU
 @Composable
 public fun PlayerScreen(
     playerViewModel: PlayerViewModel,
+    volumeViewModel: VolumeViewModel,
```
luizgrp, line 74: not sure about the two VMs per view - https://stackoverflow.com/a/73553632/1401164

luizgrp (approved):

PR #979, opened 2023-02-03
Metalava upgrade

yschimke:
#### WHAT

Metalava upgrade

#### WHY

General health and https://github.com/tylerbwong/metalava-gradle/issues/61

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
thanks, I think all the files from the auth modules will have to be regenerated

PR #985, opened 2023-02-03
Improve Auth Token Share 

luizgrp:
#### WHAT

Improve Auth Token Share.

#### WHY

Address https://github.com/google/horologist/pull/953#issuecomment-1405044082

#### HOW

- Change it use `datalayer` module;
- Change factory functions for `TokenBundleRepositoryImpl` to require a `key`, in order to allow multiple bundle of tokens to be used, when required;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenBundleRepositoryImpl.kt
```diff
@@ -0,0 +1,73 @@
...
+         * order to identify each one, otherwise the same [default][DEFAULT_TOKEN_BUNDLE_KEY] key
+         * will be used and only a single token bundle will be persisted.
+         */
+        public fun <TokenBundle> create(
+            key: String = DEFAULT_TOKEN_BUNDLE_KEY,
+            applicationContext: Context,
+            coroutineScope: CoroutineScope,
+            serializer: Serializer<TokenBundle>
+        ): TokenBundleRepositoryImpl<TokenBundle> = TokenBundleRepositoryImpl(
+            WearDataLayerRegistry.fromContext(
```
yschimke, line 56: I think WearDataLayerRegistry should be a singleton, but if this clean things up I'm ok with it. We can revisit once we have multiple users of datalayer in a typical app.

auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenBundleRepositoryImpl.kt
```diff
@@ -0,0 +1,73 @@
...
+                application = applicationContext,
+                coroutineScope = coroutineScope
+            ).protoDataStore(
+                path = buildPath(key),
+                coroutineScope = coroutineScope,
+                serializer = serializer
+            )
+        )
+
+        private fun buildPath(key: String) =
```
yschimke, line 66: What about just documenting and being stricter here? Failing on one or t'other.

auth-sample-shared/src/main/java/com/google/android/horologist/auth/sample/shared/TokenSerializer.kt
```diff
@@ -14,19 +14,24 @@
...
 
-    private val _uiState = MutableStateFlow<List<String>>(emptyList())
-    public val uiState: StateFlow<List<String>> = _uiState
+    override suspend fun readFrom(input: InputStream): String =
+        InputStreamReader(input).readText()
 
-    override fun onTokenReceived(token: String) {
-        _uiState.value += token
+    override suspend fun writeTo(t: String, output: OutputStream) {
+        withContext(Dispatchers.IO) {
```
yschimke, line 33: We should probably have Dispatchers.IO for both or neither.

Both if we expect a real IO stream, neither if we know they are ByteArrayOutputStream or similar.

yschimke (approved):
Looks nice, thanks for battling through this.

auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenBundleRepositoryImpl.kt
luizgrp: Not sure I follow, is my understand correct that you are proposing to:
1) improve documentation of the create function above to say that the key should start with a forward slash
2) throw an exception if the key does not start with forward slash

For 1, if the forward slash is not provided, it fails at runtime by the wear data layer with a message about the "path". This repository is proposing the concept of "keys" and not "path" so the the runtime error can be confusing to the user, hence I decided to verify and add the forward slash for the user.

yschimke: ok, then maybe fail if it doesn't match some pattern aaaa_aaaa? Ban key including a /?

yschimke: I'm arguing against Postels' Law. But I don't think I agree with it. I prefer fail fast.

yschimke: Your call, just explaining my opinion.

luizgrp: sorry, I'm lost, maybe some code could help

auth-sample-shared/src/main/java/com/google/android/horologist/auth/sample/shared/TokenSerializer.kt
luizgrp: I intend to improve this class in a follow up PR to have a data class instead of a simple String so this will be addressed.

yschimke:
Woot!

auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenBundleRepositoryImpl.kt
yschimke: One of these two for example.

```
        private fun buildPathFromKey(key: String): String {
            check(!key.contains('/'))
            return "/$key"
        }
```

```
        private fun buildPathStrict(key: String): String {
            check(key.startsWith('/'))
            return key
        }
```

auth-sample-shared/src/main/java/com/google/android/horologist/auth/sample/shared/TokenSerializer.kt
luizgrp: https://github.com/google/horologist/pull/996

PR #986, opened 2023-02-06
Show login screens for Uamp

yschimke:
#### WHAT

Show login screens for Uamp

#### WHY

Test the implementation effort for auth in a typical standalone app.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


auth-data/src/main/java/com/google/android/horologist/auth/data/googlesignin/GoogleSignInAuthUserRepository.kt
```diff
@@ -18,18 +18,43 @@ package com.google.android.horologist.auth.data.googlesignin
...
-) : AuthUserRepository {
+    private val applicationContext: Context,
+    private val googleSignInClient: GoogleSignInClient
+) : AuthUserRepository, GoogleSignInEventListener {
+    // simple way to trigger refreshes to the sync GoogleSignIn.getLastSignedInAccount
+    private val _authTrigger = MutableStateFlow(0)
 
-    override suspend fun getAuthenticated(): AuthUser? =
-        AuthUserMapper.map(GoogleSignIn.getLastSignedInAccount(applicationContext))
+    public val authState: Flow<AuthUser?> = _authTrigger.map { getAuthenticated() }
```
yschimke, line 43: I need somewhere, vaguely reactive I can listen to logged in state.

auth-data/src/main/java/com/google/android/horologist/auth/data/googlesignin/GoogleSignInEventListener.kt
```diff
@@ -30,7 +30,7 @@ public interface GoogleSignInEventListener {
      *
      * @param account account that signed in.
      */
-    public suspend fun onSignedIn(account: GoogleSignInAccount): Unit
+    public suspend fun onSignedIn(account: GoogleSignInAccount?): Unit
```
yschimke, line 33: This is to allow listening to log in and log out events.

auth-ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/signin/GoogleSignInScreen.kt
```diff
@@ -83,16 +82,9 @@ public fun GoogleSignInScreen(
                     viewModel.onAccountSelected(account)
                 }
             } ?: run {
-                val googleSignInClient = GoogleSignIn.getClient(
-                    context,
-                    GoogleSignInOptions.Builder(GoogleSignInOptions.DEFAULT_SIGN_IN)
-                        .requestEmail()
```
yschimke, line 89: I don't think we can hardcode the signin options here, so pulling GoogleSignInClient out to something like a singleton.

yschimke:
@luizgrp not sure if I'm doing this right, so happy to refactor it heavily.

auth-ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/signin/GoogleSignInScreen.kt
```diff
@@ -83,16 +82,9 @@ public fun GoogleSignInScreen(
...
-                        .build()
-                )
-
                 val signInRequestLauncher = rememberLauncherForActivityResult(
                     contract = GoogleSignInContract(
-                        googleSignInClient
+                        viewModel.googleSignInClient
```
luizgrp, line 87: maybe pass the `GoogleSignInClient` in `GoogleSignInScreenState.SelectAccount` state instead of making it public

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/auth/signout/UampGoogleSignOutViewModel.kt
```diff
@@ -44,7 +45,8 @@ class UampGoogleSignOutViewModel @Inject constructor(
         ) {
             viewModelScope.launch {
                 try {
-                    googleSignInClient.signOut().await()
+                    googleSignInAuthUserRepository.signOut()
+                    googleSignInEventListener.onSignedIn(null)
```
luizgrp, line 49: This line is redundant, `GoogleSignInAuthUserRepository` is already calling `onSignedIn(null)` on 
https://github.com/google/horologist/pull/986/files#diff-3cdf90df9c448c4456ab2a8c323959a3ff0e0ecc1f42a638dcf72adcbbcc6f61R58

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/settings/UampSettingsScreen.kt
```diff
@@ -121,3 +138,29 @@ fun ActionSetting(
         contentPadding = ChipDefaults.ContentPadding
     )
 }
+
+@Composable
+fun CheckedSetting(
+    value: Boolean,
+    text: String,
+    enabled: Boolean = true,
+    onCheckedChange: (Boolean) -> Unit
```
luizgrp, line 147: nit: I know it wasn't added by this PR, but to follow the [compose guidelines](https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md) the signature has to be changed to:

```kotlin
@Composable
fun CheckedSetting(
    value: Boolean,
    text: String,
    onCheckedChange: (Boolean) -> Unit
    enabled: Boolean = true
)
```

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/app/UampWearApp.kt
```diff
@@ -269,6 +271,15 @@ fun UampWearApp(
                 navController.navigateToLibrary()
             })
         }
+
+        if (appViewModel.requireLogin()) {
+            if (!appViewModel.isLoggedIn()) {
```
luizgrp, line 276: nit: this step should be redundant as `GoogleSignInPromptScreen` will do the same check and dismiss itself when it is already signed in. `GoogleSignInPromptScreen` will display a "Signing in.." screen while doing that check, just curious if the the reason you added this check here is to avoid that screen?

auth-data/src/main/java/com/google/android/horologist/auth/data/googlesignin/GoogleSignInAuthUserRepository.kt
luizgrp: Then it looks like `GoogleSignInEventListener` can be removed.

The initial idea was:
- `AuthUserRepository` - users of this library should provide an implementation of how to check if  there is an authenticated user.

- `GoogleSignInEventListener`- users of this library can opt in to receive events when a user is logged in.

If this `authState` is exposed, then the users could just listen to it instead.

luizgrp (commented):
I think our library cannot deliver a reliable "sign out" event that the users can listen to, given the implementation is out of its control. If you think it is good enough for the media sample then I'd suggest you to move the code to a custom implementation of `AuthUserRepository` in `media-sample`, instead of modifying the `GoogleSignInAuthUserRepository` provided by the library.

I like the extraction of `GoogleSignInClient` out of `GoogleSignInScreen`. In fact we could go one step further to simplify the class by moving the call to `GoogleSignIn.getLastSignedInAccount` from there to the ViewModel.

auth-data/src/main/java/com/google/android/horologist/auth/data/googlesignin/GoogleSignInAuthUserRepository.kt
yschimke: What's your advice? Go with this repository?

auth-data/src/main/java/com/google/android/horologist/auth/data/googlesignin/GoogleSignInEventListener.kt
yschimke: Yep, do we remove this listener? or add onSignedOut?  Seems like we remove?

.gitignore
```diff
@@ -49,3 +49,4 @@ checksum.txt
 
 # Paparazzi failures output folder
 out/
+/secrets.properties
```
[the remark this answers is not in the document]
yschimke: This isn't needed as is. We don't require secrets for this.

auth-data/src/main/java/com/google/android/horologist/auth/data/googlesignin/GoogleSignInAuthUserRepository.kt
luizgrp: I wouldn't set expectations on something we can't fulfil (provide a reliable way of listening to sign out), so I would not add this repository to the library. If you think it's good enough for the media sample app, then I'd move the code to a custom implementation of `AuthUserRepository` in `media-sample` module instead of changing this class.

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/app/UampWearApp.kt
yschimke: OK, I'll test that.  I want 

not signed in, not guest mode -> show login prompt
signed in, next launch -> show confirmation

So I think I still need two branches.

.gitignore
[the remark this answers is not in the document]
luizgrp: ok then this line can be removed, just move the file in your local folder from `/secrets.properties` to `/auth-sample-wear/secrets.properties`

yschimke:
@luizgrp Any advice on the failing test GoogleSignInViewModelTest, it now requires GoogleSigninClient

Use Robolectric and create a TestActivity?

luizgrp:
> Any advice on the failing test GoogleSignInViewModelTest, it now requires GoogleSigninClient
> Use Robolectric and create a TestActivity?

I think GoogleSigninClient can be created with just a context, just add `@RunWith(AndroidJUnit4::class)` to `GoogleSignInViewModelTest` and:

```kotlin
        val googleSignInClient = GoogleSignIn.getClient(
            ApplicationProvider.getApplicationContext() as Context,
            GoogleSignInOptions.Builder(GoogleSignInOptions.DEFAULT_SIGN_IN)
                .requestEmail()
                .build()
        )
```

auth-data/src/main/java/com/google/android/horologist/auth/data/googlesignin/GoogleSignInAuthUserRepository.kt
yschimke: I think it's ok if we are in the right state at

- startup
- after successful login
- after signout
- (maybe) when getAuthenticated is requested.

auth-data/src/main/java/com/google/android/horologist/auth/data/googlesignin/GoogleSignInEventListener.kt
yschimke: I'm confused between this comment and previous one 

> Then it looks like GoogleSignInEventListener can be removed.

GoogleSignInEventListener is only useful to me personally if it tells me both

- User signed in
- User signed out

Otherwise you are guaranteed to need the smart repository, so what's the point of the listener.

luizgrp: Apologies for the confusion. The comment to remove `GoogleSignInEventListener` was left in case we decide to proceed with these changes in the library, which I'm opposed to. 

Agreed that it would be nice if there was a way of listening to signed out events, however the gmscore library doesn't seem to provide it, so I don't think is feasible to provided it in horologist. 

`GoogleSignInEventListener` is a way for the user to be able to listen to the account that was selected in `GoogleSignInScreen`. It was not meant to be a repository of signed in / signed out events.

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/settings/UampSettingsScreen.kt
yschimke: Yep, onCheckedChange will be last because trailing lambda

auth-ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/signin/GoogleSignInScreen.kt
yschimke: This feels wrong also. It turns the `object` into a `data class`, maybe need to research this and find out how to use such things from a composable.

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/settings/UampSettingsScreen.kt
luizgrp: My impression from the [guidelines](https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#compose-ui-layouts) is that the only non-mandatory param that could be in used in the trailing lambda should be a composable function

auth-ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/signin/GoogleSignInScreen.kt
luizgrp: do you mean that it feels wrong from turning the `object` to `data class`? we've been using it like that in many places e.g. https://github.com/google/horologist/blob/0dba6446fbf752d5431817796c1b2cf81f9f4cc5/media-ui/src/main/java/com/google/android/horologist/media/ui/screens/browse/BrowseScreen.kt#L235-L237

yschimke:
@luizgrp I think ready for review

auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/screens/googlesignin/signin/GoogleSignInSampleViewModel.kt
```diff
@@ -19,10 +19,21 @@ package com.google.android.horologist.auth.sample.screens.googlesignin.signin
...
+        val application = this[ViewModelProvider.AndroidViewModelFactory.APPLICATION_KEY]!!
+
+        val googleSignInClient = GoogleSignIn.getClient(
+            application,
+            GoogleSignInOptions.Builder(GoogleSignInOptions.DEFAULT_SIGN_IN)
+                .requestEmail()
+                .build()
+        )
+
+        GoogleSignInViewModel(googleSignInClient, GoogleSignInEventListenerSample)
```
luizgrp, line 37: nit: maybe we should require only the `GoogleSignInOptions` to be built/passed by the user, in order to reduce complexity. Unless the creation of those `GoogleSignInClient` are costly

auth-data/build.gradle.kts
```diff
@@ -102,6 +102,7 @@ dependencies {
     implementation(libs.retrofit2.convertermoshi)
     implementation(libs.moshi.kotlin)
     ksp(libs.moshi.kotlin.codegen)
+    implementation(libs.kotlinx.coroutines.playservices)
```
luizgrp, line 105: nit: seems to be unnecessary now

auth-ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/signin/GoogleSignInScreen.kt
luizgrp: the fact that the screen is grabbing this instance from the viewmodel directly violates the [unidirectional data flow](https://developer.android.com/jetpack/compose/architecture#udf) in my view

luizgrp (approved):
Thanks for the feedback on this new API!

auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/screens/googlesignin/signin/GoogleSignInSampleViewModel.kt
yschimke: I'll leave that change to you.  This is sample app code.  It's hard to rewire without rebuilding Hilt.

luizgrp: No worries, I will come back to it in another PR.
The comment is on the sample code, but the changes are in the constructor of `GoogleSignInViewModel` of `auth-ui`.

PR #987, opened 2023-02-06
Adds ActiveDurationText to health-composables

garanj:
#### WHAT

Adds the first Composable to `health-composables`. This is the `ActiveDurationText` composable to make it easier to show a chronometer in an exercise.

#### WHY

When using `ExerciseClient` from Health Services, creating a chronometer that ticks correctly can be time consuming.

#### HOW

Creates a Composable that takes values from `ExerciseUpdate` to display a chronometer for active duration.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/health-composables/src/main/java/com/google/android/horologist/health/composables/ActiveDurationText.kt
+++ b/health-composables/src/main/java/com/google/android/horologist/health/composables/ActiveDurationText.kt
@@ -1,0 +1,196 @@
+/*
+ * Copyright 2023 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.health.composables
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.derivedStateOf
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalLifecycleOwner
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.style.TextAlign
+import androidx.health.services.client.data.ExerciseState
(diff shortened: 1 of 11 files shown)


health-composables/src/main/java/com/google/android/horologist/health/composables/ActiveDurationText.kt
```diff
@@ -0,0 +1,196 @@
...
+            val hours = activeSeconds / 3600
+            val minutes = (activeSeconds % 3600) / 60
+            val seconds = activeSeconds % 60
+            formatTemplate.format(hours, minutes, seconds)
+        }
+    }
+
+    // When the component is not visible, then the ticker should stop, and only resume when the UI
+    // component becomes visible again.
+    DisposableEffect(lifecycleOwner) {
```
yschimke, line 101: I'm happy to land this, but I think you can use `lifecycleOwner.lifecycle.repeatOnLifecycle(Lifecycle.State.RESUMED)` for this

health-composables/src/main/java/com/google/android/horologist/health/composables/ActiveDurationText.kt
```diff
@@ -0,0 +1,196 @@
...
+ * @param lifecycleOwner lifecycle owner.
+ */
+
+@ExperimentalHorologistHealthComposablesApi
+@Composable
+public fun ActiveDurationText(
+    checkpoint: ExerciseUpdate.ActiveDurationCheckpoint,
+    state: ExerciseState,
+    modifier: Modifier = Modifier,
+    formatTemplate: String = ActiveDurationDefaults.HH_MM_SS,
```
yschimke, line 75: I think we can have a cleaner separation here.  Perhaps make Text a Slot, but given the value on these important boundaries?

yschimke (approved):
I'm keen to land this as a starting point for health composables, but will follow up with some PRs for discussion on improving the approach.

health-composables/src/main/java/com/google/android/horologist/health/composables/ActiveDurationText.kt
garanj: Sounds good, so like `slot: (formattedDuration) -> Unit` ?

health-composables/src/main/java/com/google/android/horologist/health/composables/ActiveDurationText.kt
garanj: Ok cool, I need to dig into this better to understand this.

health-composables/src/main/java/com/google/android/horologist/health/composables/ActiveDurationText.kt
```diff
@@ -0,0 +1,196 @@
...
+private fun nextTimeForOffset(timeMillis: Long, offsetMillis: Long) =
+    ((timeMillis - offsetMillis) / 1000) * 1000 + 1000 + offsetMillis
+
+/**
+ * Calculates the active duration, taking into account any elapsed time since the
+ * [ExerciseUpdate.ActiveDurationCheckpoint] was delivered from Health Services.
+ *
+ * @param checkpoint The checkpoint from the most recent state change.
+ * @param state The current state of the exercise.
+ * @return The current active duration, in seconds since epoch.
```
ithinkihaveacat, line 184: nit: Can you return an actual java.time.Duration? Somewhat useful to retain/propagate the type information and the javadoc keeps talking about duration anyway.

PR #988, opened 2023-02-07
Add doc for sign in instructions

yschimke:
#### WHAT

Add doc for sign in instructions

#### WHY

Sign in will fail without this

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
thanks!

PR #993, opened 2023-02-07
Makes *MediaDisplay not need a full player state

fstanis:
This decouples the "loading" from the state, allowing the caller to specify different terms for when to be in the loading state.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


yschimke:
Seems like an improvement

yschimke (approved):

PR #995, opened 2023-02-08
A different approach to the ActiveDurationText

yschimke:
#### WHAT

A different approach to the ActiveDurationText

#### WHY

Introduce a slot API, and use repeatOnLifecycle

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
@garanj Thinking more about the lifecycle code.

We could remove that and turn the ActiveDurationCheckpoint into a Flow, that ticks every X.  IT should probably live in a ViewModel and the UI is just subscribed to it, while the UI is open.

Or we could take the approach from Media and turn the exercise state into some UI State, with a predictor function.

garanj (commented):
Awesome!

garanj:
> @garanj Thinking more about the lifecycle code.
> 
> We could remove that and turn the ActiveDurationCheckpoint into a Flow, that ticks every X. IT should probably live in a ViewModel and the UI is just subscribed to it, while the UI is open.
> 
> Or we could take the approach from Media and turn the exercise state into some UI State, with a predictor function.
> 
> Overkill for now, just discussion.

I had wondered about a flow and putting it in the ViewModel. Makes sense, I wondered if there might ever be a case where the app wouldn't be using a ViewModel?

yschimke:
> I wondered if there might ever be a case where the app wouldn't be using a ViewModel?

It's not exclusive. And the same patterns work if not, but in our MAD guidelines the ViewModel is a useful place to hoist state to.

PR #996, opened 2023-02-08
Improve token sharing sample

luizgrp:
#### WHAT

Improve token sharing sample.

#### WHY

- In order to have something more complex than just a simple string being shared between the devices;
- In order to demo the sharing of multiple token bundles;

#### HOW

Add `TokenBundle` proto definition in `auth-sample-shared` and update `TokenSerializer` to use it.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth-sample-shared/src/main/java/com/google/android/horologist/auth/sample/shared/TokenBundleSerializer.kt
+++ b/auth-sample-shared/src/main/java/com/google/android/horologist/auth/sample/shared/TokenBundleSerializer.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2023 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.auth.sample.shared
+
+import androidx.datastore.core.CorruptionException
+import androidx.datastore.core.Serializer
+import com.google.android.horologist.auth.sample.shared.model.TokenBundleProto.TokenBundle
+import com.google.protobuf.InvalidProtocolBufferException
+import java.io.InputStream
+import java.io.OutputStream
+
+public object TokenBundleSerializer : Serializer<TokenBundle?> {
+
+    override val defaultValue: TokenBundle? = null
+
+    override suspend fun readFrom(input: InputStream): TokenBundle? =
+        try {
+            TokenBundle.parseFrom(input)
+        } catch (exception: InvalidProtocolBufferException) {
+            throw CorruptionException("Cannot read proto.", exception)
+        }
+
+    override suspend fun writeTo(t: TokenBundle?, output: OutputStream) {
+        t?.writeTo(output)
+    }
+}
--- a/auth-sample-shared/src/main/proto/token_bundle.proto
+++ b/auth-sample-shared/src/main/proto/token_bundle.proto
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      https://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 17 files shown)


auth-sample-shared/src/main/proto/token_bundle.proto
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+syntax = "proto3";
+
+package com.google.android.horologist.auth.sample.shared.model;
+
+option java_package = "com.google.android.horologist.auth.sample.shared.model";
+option java_outer_classname = "TokenBundleProto";
+
+message TokenBundle {
```
yschimke, line 24: Is it worth moving this into the library as a default with the 3-4 fields we expect are typically required? But making it flexible to use a different serializer?

luizgrp: Good idea, can follow up in a another pr. Any idea of which fields would these be?

auth-sample-shared/src/main/java/com/google/android/horologist/auth/sample/shared/TokenBundleSerializer.kt
```diff
@@ -16,22 +16,25 @@
...
+    override val defaultValue: TokenBundle? = null
 
-    override suspend fun writeTo(t: String, output: OutputStream) {
-        withContext(Dispatchers.IO) {
-            output.write(t.toByteArray())
+    override suspend fun readFrom(input: InputStream): TokenBundle? =
+        try {
+            TokenBundle.parseFrom(input)
+        } catch (exception: InvalidProtocolBufferException) {
+            throw CorruptionException("Cannot read proto.", exception)
```
yschimke, line 34: Are you sure this does what you think?

```
        val inputStream: InputStream = ByteArrayInputStream(byteArrayOf())
        println(SettingsProto.Settings.parseFrom(inputStream))
```

# com.google.android.horologist.mediasample.domain.proto.SettingsProto$Settings@dd32a22

auth-sample-shared/src/main/java/com/google/android/horologist/auth/sample/shared/TokenBundleSerializer.kt
```diff
@@ -16,22 +16,25 @@
...
-import kotlinx.coroutines.withContext
+import com.google.android.horologist.auth.sample.shared.model.TokenBundleProto.TokenBundle
+import com.google.protobuf.InvalidProtocolBufferException
 import java.io.InputStream
-import java.io.InputStreamReader
 import java.io.OutputStream
 
-public object TokenSerializer : Serializer<String> {
-    override val defaultValue: String = ""
+public object TokenBundleSerializer : Serializer<TokenBundle?> {
```
yschimke, line 26: I wonder whether it's simpler just to treat a TokenBundle with all blank keys as missing.  But your call. If this works, modulo the readFrom, then it seems ok.

auth-sample-shared/src/main/java/com/google/android/horologist/auth/sample/shared/TokenBundleSerializer.kt
luizgrp: Not sure I follow. Are you commenting on line 34 or in the whole try/catch code? What SettingsProto has to do with this code?

The samples are working as expected?

yschimke: Don't you want this to turn empty into null?

auth-sample-shared/src/main/proto/token_bundle.proto
yschimke: Maybe ask 3P devs, but start with ones you know and expand? 

If *forced* to guess?

```
  string access_token = 1;
  string refresh_token = 2;
  string client_id = 3;
  string client_token = 4;
```

auth-sample-shared/src/main/proto/token_bundle.proto
```diff
@@ -0,0 +1,27 @@
...
+  string access_token = 1;
+  string refresh_token = 2;
```
yschimke, line 26: If you want these truly nullable, you can use google.protobuf.StringValue which can be null in generated code.

auth-sample-shared/src/main/java/com/google/android/horologist/auth/sample/shared/TokenBundleSerializer.kt
luizgrp: No, that wouldn't be desired but I see it is the result of that function when the ByteArray is empty 👍 

The samples are working as intended so it looks like when the default value is null, this function doesn't called.
I will leave as it is, unless you see any other further issue?

auth-sample-shared/src/main/java/com/google/android/horologist/auth/sample/shared/TokenBundleSerializer.kt
luizgrp: I think it's confusing to treat an object instance as "missing", it can lead to "tolerable" paths in the code where it should fail fast (e.g. crash during development as some info that should have been sent was not sent) or report (e.g. server returned empty data, app should log and report it instead of just treating as missing value).

If it had to go that way then I would add an extra boolean property to make it explicit that was marked as missing, or if explore a more complex data structure like sealed classes in kotlin. I think null serves its purpose here to mark as missing, if there are not other further implications.

If one wants to avoid nullable types in the app, then a domain model can fulfil that purpose. The proto model is part of the data layer, and a null value for it can be mapped into a "TokenBundle.Missing" class from the domain.

PR #997, opened 2023-02-08
Avoid StrictMode violation from reading on main thread

yschimke:
#### WHAT

Avoid StrictMode violation from reading on main thread

#### WHY

GoogleSignIn.getLastSignedInAccount is not main thread safe.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):

media-sample/src/main/java/com/google/android/horologist/mediasample/data/auth/GoogleSignInAuthUserRepository.kt
```diff
@@ -44,8 +46,8 @@ public class GoogleSignInAuthUserRepository(
 
     public val authState: Flow<AuthUser?> = _authTrigger.map { getAuthenticated() }
 
-    override suspend fun getAuthenticated(): AuthUser? {
-        return AuthUserMapper.map(GoogleSignIn.getLastSignedInAccount(applicationContext))
+    override suspend fun getAuthenticated(): AuthUser? = withContext(Dispatchers.Default) {
```
luizgrp, line 49: shouldn't use Dispatchers.IO given it's a DiskReadViolation?

yschimke: Yeah, good point.

PR #1002, opened 2023-02-09
Refactors metadata display and handling

fstanis:
#### WHAT
- Moves display-related components into new `components.display` package
- Moves default title and artist to mapper constructor
- Changes how metadata is determined to be more in line with media3 apps
- Makes determining loading state more explicit (but currently relies on empty metadata)

#### WHY
Simplifies customization of metadata extraction and makes displaying it more robust, especially when a loading state is needed.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests

### DIFF
--- a/media-data/src/test/java/com/google/android/horologist/media/data/mapper/MediaMapperTest.kt
+++ b/media-data/src/test/java/com/google/android/horologist/media/data/mapper/MediaMapperTest.kt
@@ -47,7 +47,6 @@ class MediaMapperTest {
         val id = "id"
         val uri = "uri"
         val title = "title"
-        val displayTitle = "displayTitle"
         val artist = "artist"
         val albumArtist = "albumArtist"
         val artworkUri = "artworkUri"
@@ -58,7 +57,6 @@ class MediaMapperTest {
             .setMediaMetadata(
                 MediaMetadata.Builder()
                     .setTitle(title)
-                    .setDisplayTitle(displayTitle)
                     .setArtist(artist)
                     .setAlbumArtist(albumArtist)
                     .setArtworkUri(Uri.parse(artworkUri))
@@ -67,12 +65,12 @@ class MediaMapperTest {
             .build()
 
         // when
-        val result = sut.map(mediaItem, null)
+        val result = sut.map(mediaItem, mediaItem.mediaMetadata)
 
         // then
         assertThat(result.id).isEqualTo(id)
         assertThat(result.uri).isEqualTo(uri)
-        assertThat(result.title).isEqualTo(displayTitle)
+        assertThat(result.title).isEqualTo(title)
         assertThat(result.artist).isEqualTo(artist)
         assertThat(result.artworkUri).isEqualTo(artworkUri)
     }
@@ -95,7 +93,7 @@ class MediaMapperTest {
             .build()
 
         // when
-        val result = sut.map(mediaItem, null)
+        val result = sut.map(mediaItem, mediaItem.mediaMetadata)
 
         // then
         assertThat(result.id).isEqualTo(id)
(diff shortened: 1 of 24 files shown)


media-sample/src/main/java/com/google/android/horologist/mediasample/di/DataModule.kt
```diff
@@ -171,8 +173,8 @@ class DataModule {
     fun mediaExtrasMapper(): MediaExtrasMapper = MediaExtrasMapperNoopImpl
 
     @Provides
-    fun mediaMapper(mediaExtrasMapper: MediaExtrasMapper): MediaMapper =
-        MediaMapper(mediaExtrasMapper)
+    fun mediaMapper(resourceProvider: ResourceProvider, mediaExtrasMapper: MediaExtrasMapper): MediaMapper =
```
luizgrp, line 176: this is something that I don't think we should keep propagating (https://github.com/google/horologist/pull/423/files#r926699559)

Sample of a scenario where it goes wrong: 
- a scope annotation (e.g. singleton) is added to this MediaMapper provider function;
- there are translations for "no_title" string;
- then once there is a locale changed, the MediaMapper instance will never get the correct string

media-data/src/test/java/com/google/android/horologist/media/data/mapper/MediaMapperTest.kt
```diff
@@ -35,10 +35,12 @@ import org.robolectric.annotation.Config
 class MediaMapperTest {
 
     private lateinit var sut: MediaMapper
+    private val defaultTitle = "defaultArtist"
```
luizgrp, line 38: nit: defaultTitle

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/player/PlayerScreen.kt
```diff
@@ -102,24 +97,15 @@ public fun PlayerScreen(
...
-        )
-    } else {
-        InfoMediaDisplay(
-            message = stringResource(R.string.horologist_nothing_playing),
-            modifier = modifier
-        )
-    }
+    MediaInfoDisplay(
+        media = playerUiState.media,
+        loading = !playerUiState.connected || playerUiState.media?.loading == true,
```
luizgrp, line 106: would it be possible to have a "loading" property in PlayerUiState and this logic inside PlayerUiStateMapper? You know better so please take my comment as a naive question

media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/MediaUiModel.kt
```diff
@@ -22,8 +22,11 @@ import com.google.android.horologist.media.ui.ExperimentalHorologistMediaUiApi
...
-    val title: String? = null,
-    val subtitle: String? = null,
+    val title: String,
+    val subtitle: String = "",
     val artworkUri: String? = null,
     val artworkColor: Color? = null
-)
+) {
+    // Consider making this a field
+    val loading: Boolean get() = title.isEmpty() && subtitle.isEmpty()
```
luizgrp, line 31: can this be a property and this logic be inside MediaUiModelMapper?

luizgrp (approved):
LGTM

media-ui/src/main/java/com/google/android/horologist/media/ui/screens/player/PlayerScreen.kt
fstanis: Maybe; I'm still unsure if this should be "metadata loading" or loading in general, so that's why I left it this way as a temporary solution.

media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/MediaUiModel.kt
fstanis: Again, really depends on what this will mean in the future, I added it this way just so it wouldn't be an API-breaking change once I know how to proceed.

PR #1003, opened 2023-02-10
Material assessment

laiyichin:
#### WHAT
- Include a background for play / pause button
- Disable volume button if nothing is playing

#### WHY
Matching material design

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

Screenshot 2023-02-10 at 16 07 46


media-sample/src/main/java/com/google/android/horologist/mediasample/ui/player/UampMediaPlayerScreen.kt
```diff
@@ -60,7 +60,7 @@ fun UampMediaPlayerScreen(
             UampSettingsButtons(
                 volumeState = volumeState,
                 onVolumeClick = onVolumeClick,
-                enabled = it.connected
+                enabled = it.connected && it.media != null,
```
fstanis, line 63: I feel this condition should really be something like `state != PlayerState.Idle` - but I now see we don't expose that in `PlayerUiState` and I think maybe we should in some way. Perhaps just add a TODO or so, so that we can revisit.

fstanis:
Doesn't this mean the regular `PlayPauseButton` (i.e. not `PlayPauseProgressButton`) would still render without a circle background? I assume both are supposed to look the same when disabled?

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/player/UampMediaPlayerScreen.kt
```diff
@@ -64,7 +64,7 @@ fun UampMediaPlayerScreen(
             UampSettingsButtons(
                 volumeState = volumeState,
                 onVolumeClick = onVolumeClick,
-                enabled = it.connected
+                enabled = it.connected && it.media != null
```
yschimke, line 67: I'm a bit surprised we have this condition. You can adjust volume even with nothing selected.  But I'll defer to you.

yschimke (approved):

media-sample/src/main/java/com/google/android/horologist/mediasample/ui/player/UampMediaPlayerScreen.kt
laiyichin: Added todo.

PR #1004, opened 2023-02-10
Vertically align MessageMediaDisplay

fstanis:


media-ui/src/main/java/com/google/android/horologist/media/ui/components/display/MessageMediaDisplay.kt
```diff
@@ -34,17 +34,19 @@ import com.google.android.horologist.media.ui.ExperimentalHorologistMediaUiApi
...
             text = message.orEmpty(),
             modifier = Modifier.fillMaxWidth(0.7f),
             color = MaterialTheme.colors.onBackground,
             textAlign = TextAlign.Center,
             overflow = TextOverflow.Ellipsis,
             maxLines = 2,
-            style = MaterialTheme.typography.body2
+            style = style
         )
+        Text("", style = style, minLines = 2)
```
yschimke, line 50: Why minLines?

media-ui/src/main/java/com/google/android/horologist/media/ui/components/display/MessageMediaDisplay.kt
```diff
@@ -34,17 +34,19 @@ import com.google.android.horologist.media.ui.ExperimentalHorologistMediaUiApi
 @Composable
 public fun MessageMediaDisplay(
     modifier: Modifier = Modifier,
-    message: String? = null
+    message: String
 ) {
-    Column(modifier = modifier, horizontalAlignment = Alignment.CenterHorizontally) {
+    Box(modifier = modifier, contentAlignment = Alignment.Center) {
```
yschimke, line 39: Does contentAlignment = Alignment.TopCenter, and skipping the second text work? Seems like it would let the text grow more.

media-ui/src/main/java/com/google/android/horologist/media/ui/components/display/MessageMediaDisplay.kt
fstanis: This is just a placeholder to set the height of the `Box` to be 2 lines.

media-ui/src/main/java/com/google/android/horologist/media/ui/components/display/MessageMediaDisplay.kt
fstanis: The issue is that the `Box`'s height matches the height of the `Text`, so the `Box`'s vertical alignment doesn't do anything here and it gets the alignment from the parent.

laiyichin (approved):
LGTM. Shall we add a screenshot to the PR?

fstanis:
`MediaPlayerStatesScreenTest_mediaPlayerScreen[0]_nomedia.png` does cover this; updated files. We don't have a two-line one though.

PR #1005, opened 2023-02-13
Rename generic type

luizgrp:
#### WHAT

Rename generic type

#### WHY

In order to make it more concise.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
LGTM

PR #1014, opened 2023-02-14
Persistent progress bar

laiyichin:
#### WHAT
Instead of hiding progress bar, keep it persistent (but set duration to zero) 

#### WHY
So there is no sudden jumps when skipping media items.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media-ui/src/main/java/com/google/android/horologist/media/ui/state/mapper/TrackPositionUiModelMapper.kt
```diff
@@ -38,7 +39,7 @@ public object TrackPositionUiModelMapper {
             return TrackPositionUiModel.Actual.ZERO
         }
         if (currentPositionMs == null || durationMs == null || durationMs <= 0) {
-            return TrackPositionUiModel.Hidden
+            return TrackPositionUiModel.Actual(0f, Duration.ZERO, Duration.ZERO)
```
fstanis, line 42: You could merge this with the `if` above which returns `TrackPositionUiModel.Actual.ZERO`. But I'm not sure this is the correct place to do this. If the position should be zero, then `PlaybackStateEvent` should be valid (i.e. non-null duration and position).

laiyichin:
Discussed offline, this change is not needed anymore.

PR #1015, opened 2023-02-14
Add a basic preview of media complications

yschimke:
#### WHAT

Add a basic preview of media complications.

image

#### WHY

https://github.com/google/horologist/pull/1011 prompted looking at how we test these.

From an ADS talk, had this complication preview but wasn't using it.

#### HOW

Copied an old prototype, and shoved in some sample data. 

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


ataulm:
>https://github.com/google/horologist/pull/1011 prompted looking at how we test these.

by "test" we mean, how can we see what these look like? Or test as in, how can we write tests for these?

I'll leave this with you for now (code looks fine). I'm philosophically wondering about the cost/value trade-off to complication devs to follow this pattern given that they don't get much choice in how complications are ultimately rendered.

yschimke:
@ataulm yeah, this is not library code.  I'm currently only using it for testing this media complication. I've found it useful, without it I get the icon tinting wrong.

ataulm (approved):

PR #1018, opened 2023-02-16
Moves onVolumeChangeByScroll to a modifier

fstanis:


media-ui/src/main/java/com/google/android/horologist/media/ui/screens/player/PlayerScreen.kt
```diff
@@ -202,3 +200,11 @@ public fun PlayerScreen(
         }
     }
 }
+
+private fun Modifier.onVolumeChangeByScroll(
```
yschimke, line 204: Nice improvement!

PR #1019, opened 2023-02-16
Improve StreamlineSignIn components and sample

luizgrp:
#### WHAT

Improve StreamlineSignIn components and sample.

#### HOW

- Improve Streamline sample to display confirmation dialog after selecting an account;
- Surface `AccountUiModel` to composable components instead of `AuthUser`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
I like the UI model, I've gotten used to them.

PR #1025, opened 2023-02-21
Improve StreamlineSignInDefaultScreen callback

luizgrp:
#### WHAT

Improve `StreamlineSignInDefaultScreen`'s `onSignedInConfirmationDialogDismissOrTimeout` callback to receive an `AccountUiModel`.

#### WHY

So callers can receive the account picked by the user.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/screens/common/streamline/StreamlineSignInSampleScreen.kt
+++ b/auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/screens/common/streamline/StreamlineSignInSampleScreen.kt
@@ -49,11 +49,11 @@ fun StreamlineSignInSampleScreen(
     modifier: Modifier = Modifier,
     viewModel: StreamlineSignInDefaultViewModel = viewModel(factory = StreamlineSignInSampleViewModelFactory)
 ) {
-    var showSignedInConfirmationDialog by rememberSaveable { mutableStateOf(false) }
+    var showNoAccountsAvailableDialog by rememberSaveable { mutableStateOf(false) }
 
     StreamlineSignInDefaultScreen(
-        onSignedInConfirmationDialogDismissOrTimeout = navController::popBackStack,
-        onNoAccountsAvailable = { showSignedInConfirmationDialog = true },
+        onSignedInConfirmationDialogDismissOrTimeout = { navController.popBackStack() },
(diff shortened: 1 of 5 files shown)


auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/screens/common/streamline/StreamlineSignInSampleScreen.kt
```diff
@@ -49,11 +49,11 @@ fun StreamlineSignInSampleScreen(
     modifier: Modifier = Modifier,
     viewModel: StreamlineSignInDefaultViewModel = viewModel(factory = StreamlineSignInSampleViewModelFactory)
 ) {
-    var showSignedInConfirmationDialog by rememberSaveable { mutableStateOf(false) }
+    var showNoAccountsAvailableDialog by rememberSaveable { mutableStateOf(false) }
 
     StreamlineSignInDefaultScreen(
-        onSignedInConfirmationDialogDismissOrTimeout = navController::popBackStack,
-        onNoAccountsAvailable = { showSignedInConfirmationDialog = true },
+        onSignedInConfirmationDialogDismissOrTimeout = { navController.popBackStack() },
```
yschimke, line 55: Isn't the method reference optimal? Maybe not here, but across compse codebases?

yschimke (approved):

auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/screens/common/streamline/StreamlineSignInSampleScreen.kt
luizgrp: yes, but the signature of `onSignedInConfirmationDialogDismissOrTimeout` was changed in this PR to have a param, so it doesn't match `popBackStack` signature anymore

future iteration would be to remember this lambda to avoid recompositions, but planning to run metrics for the whole module and try to fix all in one go

yschimke: Nice idea.  When you go through that, maybe post results as you go to horologist-dev. I'd love to see a run through that on a real library or app.

PR #1029, opened 2023-02-22
Add overloaded SignedInConfirmationDialog 

luizgrp:
#### WHAT

Add overloaded `SignedInConfirmationDialog` with `AccountUiModel` as parameter.

#### WHY

For convenience.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth-composables/src/main/java/com/google/android/horologist/auth/composables/dialogs/SignedInConfirmationDialog.kt
+++ b/auth-composables/src/main/java/com/google/android/horologist/auth/composables/dialogs/SignedInConfirmationDialog.kt
@@ -44,6 +44,7 @@ import androidx.wear.compose.material.dialog.DialogDefaults
 import coil.compose.rememberAsyncImagePainter
 import com.google.android.horologist.auth.composables.ExperimentalHorologistAuthComposablesApi
 import com.google.android.horologist.auth.composables.R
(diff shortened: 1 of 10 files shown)


auth-composables/src/main/java/com/google/android/horologist/auth/composables/dialogs/SignedInConfirmationDialog.kt
```diff
@@ -73,18 +74,40 @@ public fun SignedInConfirmationDialog(
...
     }
 }
 
+/**
+ * A [SignedInConfirmationDialog] that can display the name, email and avatar image of an
+ * [AccountUiModel].
+ */
+@ExperimentalHorologistAuthComposablesApi
+@Composable
+public fun SignedInConfirmationDialog(
```
yschimke, line 90: Do we still need the one above?

yschimke (approved):

auth-composables/src/main/java/com/google/android/horologist/auth/composables/dialogs/SignedInConfirmationDialog.kt
luizgrp: I'm afraid yes, it's needed in a follow up PR, breaking it down into smaller ones

PR #1030, opened 2023-02-22
Do not show Watch speaker as default route

kul3r4:
#### WHAT
Do not show watch speaker as default route when opening Volume screen before connecting to an Output device 

#### WHY
It is now confusing that "Watch speaker" is shown on the chip but there is no way to play on the watch speaker, you always have to tap on it in order to open the BT settings fragment

#### HOW
Add a "isPlayable" variable that is checked on the UI

#### Checklist :clipboard:
- [x] Run spotless check
- [ ] Run tests


yschimke:
As discussed, I think this change doesn't work if Watch Speaker playback is allowed. So I think it should be something like changing

https://github.com/google/horologist/blob/65d0be9499b9d89e67e77339ed39d3c5f5ff048b/audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/AudioOutputUi.kt#L36

to have some flag of whether this is a valid output.  And if not then we display a prompt instead of the device.

audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/AudioOutputUi.kt
```diff
@@ -42,9 +42,10 @@ public data class AudioOutputUi(
 @Composable
 public fun AudioOutput.toAudioOutputUi(): AudioOutputUi {
     return AudioOutputUi(
-        displayName = when (type) {
-            TYPE_WATCH -> stringResource(id = R.string.horologist_speaker_name)
-            TYPE_NONE -> stringResource(id = R.string.horologist_output_none)
+        displayName = when  {
+            type == TYPE_WATCH && isPlayable -> stringResource(id = R.string.horologist_speaker_name)
+            type == TYPE_WATCH && !isPlayable -> stringResource(id = R.string.choose_device)
```
yschimke, line 47: I think the `!isPlayable` can just be the first case, without TYPE_WATCH. If we can't play with this output we can tell them to choose a different output.

Then the rest remains as is.

yschimke:
Worth updating title and description.

audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/AudioOutputUi.kt
```diff
@@ -42,9 +42,10 @@ public data class AudioOutputUi(
...
-        displayName = when (type) {
-            TYPE_WATCH -> stringResource(id = R.string.horologist_speaker_name)
-            TYPE_NONE -> stringResource(id = R.string.horologist_output_none)
+        displayName = when {
+            !isPlayable -> stringResource(id = R.string.choose_device)
+            type == TYPE_WATCH -> stringResource(id = R.string.horologist_speaker_name)
+            type == TYPE_NONE -> stringResource(id = R.string.horologist_output_none)
             else -> name
         },
         imageVector = when (type) {
```
yschimke, line 51: Do we want to also check this to the same path as TYPE_NONE? or some other icon, or no icon?

yschimke: It makes me wonder whether it's nicer to have, thoughts?

```
if (isPlayable) {
  // current block
} else {
  AudioOutputUi(displayName = stringResource(id = R.string.choose_device), imageVector = null, isConnected = false)
}
```

yschimke (approved):
Pending review comments

PR #1031, opened 2023-02-22
Add avatar slots to StreamlineSignInScreen

luizgrp:
#### WHAT

Add avatar slots to `StreamlineSignInScreen`.

#### WHY

Client setup can't retrieve avatar uri at the moment and needs to pass 3rd party view component instead.

#### HOW

- Add overloaded `SignedInConfirmationDialog` composable that accepts a composable slot param for the avatar;
- Add overloaded `SelectAccountScreen` composable that accepts a composable slot param for the avatar and a generic type for accounts;
- Change `StreamlineSignInScreen` composable to accept a composable slot params for the select account screen avatar and for the signed in confirmation dialog avatar;
- Remove other versions of `StreamlineSignInScreen` to avoid maintenance overhead for now - if client overcome challenges to retrieve the avatar uri, we could bring back them, potentially even reverting this PR;
- Add `StreamlineAccountRepository` with a generic type for account;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/screens/common/streamline/StreamlineSignInSampleScreen.kt
```diff
@@ -47,18 +58,48 @@ fun StreamlineSignInSampleScreen(
...
+        factory = StreamlineSignInSampleViewModelFactory
+    )
 ) {
     var showNoAccountsAvailableDialog by rememberSaveable { mutableStateOf(false) }
 
-    StreamlineSignInDefaultScreen(
+    StreamlineSignInScreen(
+        accountName = { account -> account.name ?: "" },
+        accountEmail = { account -> account.email },
+        selectAccountAvatarContent = {
```
yschimke, line 70: I wonder whether these can be named in terms of the form, rather than the context? If you have to use an avatar twice in this screen, would you have two params or one?

Not blocking.

yschimke (approved):
Seems like a more flexible design, one longer term comment about how we avoid proliferation of very specific slots.

auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/screens/common/streamline/StreamlineSignInSampleScreen.kt
luizgrp: > If you have to use an avatar twice in this screen, would you have two params or one?

The reason for having two params is that the size of the avatars are different in the different screens, see [here](https://github.com/google/horologist/pull/1031/files#diff-2e7e93f05dcdf10f7dc3ce8c33313d331f7569e10490615b85ea7bf061a23c11R112) and [here](https://github.com/google/horologist/pull/1031/files#diff-d69435c85b82b0e2d4605b5be0f64dccbde72ecd957348203d96f41048f6fd08R82). It could accept a single param, ask for the largest size and then display it in a smaller size for the other screen, but didn't explore that far. I should put those reference in the kdoc of this component.

> I wonder whether these can be named in terms of the form

Could you please give examples? Like `smallAvatarContent` and `largeAvatarContent`?

yschimke: Yeah, just generally if I'm adopting this auth library, and I want to integrate with my account avatars, is there some set of components I need to implement. Maybe smallAvatar and largeAvatar?

PR #1032, opened 2023-02-22
Rubberband

garanj:
#### WHAT

Provides basic DataLayer-based helpers for aiding phone-watch-based discovery, installation and launch

#### WHY

A challenge where an app has both phone and Wear versions is promoting easy ways to bring the two together.

#### HOW

Provides WearDataLayerAppHelper and PhoneDataLayerAppHelper, which have methods for:

- Discovering connected devices, with the app installation status on each, and status of Tile installation
- Launching installation of the app on the other device
- Launching the app remotely on the other device
- Monitoring the connected status of the device with the app installed
- Launching the Wear Companion, to help guide towards Tile installation

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


datalayer/src/main/java/com/google/android/horologist/data/DataLayerAppHelperService.kt
```diff
@@ -0,0 +1,86 @@
...
+import android.util.Log
+import com.google.android.gms.tasks.Task
+import com.google.android.gms.tasks.Tasks
+import com.google.android.gms.wearable.WearableListenerService
+import kotlinx.coroutines.runBlocking
+
+/**
+ * Base service to respond to incoming requests from the partnering app on the connected device.
+ */
+public abstract class DataLayerAppHelperService : WearableListenerService() {
```
yschimke, line 30: I might suggest we refactor this later, I was hoping that apps could extend `com.google.android.horologist.data.WearDataService` and those methods would call the right things on WearDataLayerRegistry, otherwise I fear apps will trend towards 4 or 5 implementations of `com.google.android.gms.wearable.WearableListenerService`

datalayer-phone/src/main/java/com/google/android/horologist/datalayer/phone/PhoneDataLayerAppHelper.kt
```diff
@@ -0,0 +1,71 @@
...
+@ExperimentalHorologistDataLayerApi
+public class PhoneDataLayerAppHelper(context: Context) :
+    DataLayerAppHelper(context) {
+    /**
+     * Some devices report back a different packageName from getCompanionPackageForNode() than is
+     * the actual package of the Companion app. Where this is the case, this lookup ensures the
+     * correct companion app can be launched. (pursuing whether this is a bug or not).
+     */
+    private val companionLookup = mapOf(
+        "com.samsung.android.waterplugin" to "com.samsung.android.app.watchmanager",
```
yschimke, line 41: bit nervous about this as a public pattern, we should raise a FR for it, or find a better way longer term.  Maybe worth a "TODO temporary while correct solution is discovered"

datalayer-phone/src/main/java/com/google/android/horologist/datalayer/phone/PhoneDataLayerAppHelper.kt
```diff
@@ -0,0 +1,71 @@
...
+        "com.samsung.android.heartplugin" to "com.samsung.android.app.watchmanager"
+    )
+
+    override suspend fun installOnNode(node: String) {
+        val intent = Intent(Intent.ACTION_VIEW)
+            .addCategory(Intent.CATEGORY_BROWSABLE)
+            .setData(Uri.parse(playStoreUri))
+        remoteActivityHelper.startRemoteActivity(intent, node).await()
```
yschimke, line 49: Any error cases we need to document?

datalayer/src/main/java/com/google/android/horologist/data/DataLayerAppHelperService.kt
yschimke: Basically if PhoneDataLayerAppHelper is a cheap singleton and created early in app startup, then it can register against the DataLayer and receive the same events without a specific listener.

datalayer-watch/quality/lint/lint-baseline.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<issues format="6" by="lint 7.4.1" type="baseline" client="gradle" dependencies="false" name="AGP (7.4.1)" variant="all" version="7.4.1">
```
yschimke, line 2: Why this entry?

datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
```diff
@@ -0,0 +1,96 @@
...
+public class WearDataLayerAppHelper(context: Context, private val appStoreUri: String? = null) :
+    DataLayerAppHelper(context) {
+    override suspend fun installOnNode(node: String) {
+        if (appStoreUri != null &&
+            PhoneTypeHelper.getPhoneDeviceType(context) == PhoneTypeHelper.DEVICE_TYPE_IOS
+        ) {
+            val intent = Intent(Intent.ACTION_VIEW)
+                .addCategory(Intent.CATEGORY_BROWSABLE)
+                .setData(Uri.parse(appStoreUri))
+            remoteActivityHelper.startRemoteActivity(intent, node).await()
```
yschimke, line 47: Error handling? or documentation?

datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
```diff
@@ -0,0 +1,96 @@
...
+import kotlinx.coroutines.tasks.await
+
+/**
+ * Subclass of [DataLayerAppHelper] for use on Wear devices.
+ */
+@ExperimentalHorologistDataLayerApi
+public class WearDataLayerAppHelper(context: Context, private val appStoreUri: String? = null) :
+    DataLayerAppHelper(context) {
+    override suspend fun installOnNode(node: String) {
+        if (appStoreUri != null &&
```
yschimke, line 41: Seems to silently do nothing in these cases, should we make that obvious via a return value or error?

datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Marks a tile as removed. Call this in [TileService#onTileRemoveEvent]. Supplying a name is
+     * mandatory to disambiguate from the installation or removal of other tiles your app may have.
+     *
+     * @param tileName The name of the tile.
+     */
+    public suspend fun markTileAsRemoved(tileName: String) {
+        require(tileName.isNotEmpty())
+        capabilityClient.removeLocalCapability("$tilePrefix$tileName").await()
```
yschimke, line 94: Can this fail with UNKNOWN_CAPABILITY?

datalayer/build.gradle.kts
```diff
@@ -16,17 +16,20 @@
...
+    id("com.google.protobuf")
     kotlin("android")
 }
 
 android {
     compileSdk = 33
 
     defaultConfig {
-        minSdk = 21
+        minSdk = 25
```
yschimke, line 32: Not sure we want to bump this.  Can we leave it? We don't want all phone apps to be minSdk 25

datalayer/src/main/java/com/google/android/horologist/data/DataLayerAppHelper.kt
```diff
@@ -0,0 +1,170 @@
...
+ * apps as part of the user's journey.
+ */
+abstract class DataLayerAppHelper(protected val context: Context) {
+    private val installedDeviceCapabilityUri = "wear://*/$CAPABILITY_DEVICE_PREFIX"
+
+    // The installation of individual Tiles is tracked via Local Capabilities with this prefix.
+    protected val tilePrefix = "${DATA_LAYER_APP_HELPER_CAPABILITY}_tile_"
+    protected val playStoreUri = "market://details?id=${context.packageName}"
+
+    protected val capabilityClient by lazy { Wearable.getCapabilityClient(context) }
```
yschimke, line 45: We should swap these 4 out for WearDataLayerRegistry

datalayer/src/main/java/com/google/android/horologist/data/DataLayerAppHelper.kt
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /**
+     * Creates a flow to keep the client updated with the set of connected devices with the app
+     * installed.
+     */
+    public val connectedAndInstalledNodes = callbackFlow<Set<Node>> {
+        val listener = object : CapabilityClient.OnCapabilityChangedListener {
+            override fun onCapabilityChanged(capability: CapabilityInfo) {
+                trySend(capability.nodes.filter { it.isNearby }.toSet())
```
yschimke, line 85: why filter on isNearby here? Is that what we mean by connected here?

datalayer/src/main/java/com/google/android/horologist/data/DataLayerAppHelper.kt
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    /**
+     * Launches to the appropriate store on the specified node to allow installation of the app.
+     *
+     * @param node The node to launch on.
+     */
+
+    abstract suspend fun installOnNode(node: String)
```
yschimke, line 108:
```suggestion
     */
    abstract suspend fun installOnNode(node: String)
```

datalayer/src/main/java/com/google/android/horologist/data/DataLayerAppHelper.kt
```diff
@@ -0,0 +1,170 @@
...
+        val response = messageClient.sendRequest(node, LAUNCH_APP, request.toByteArray()).await()
+        return AppHelperResult.parseFrom(response).code
+    }
+
+    /**
+     * Launch own app on the specified node.
+     */
+    public suspend fun startRemoteOwnApp(node: String): AppHelperResultCode {
+        val request = launchRequest { ownApp = ownAppConfig { } }
+        val response = messageClient.sendRequest(node, LAUNCH_APP, request.toByteArray()).await()
```
yschimke, line 136: I wonder about using message client for launching own app.  Should be use RemoteActivity and expect a uri, which could be a deeplink?

yschimke (approved):
Looks great, I'd suggest following up in separate PRs for any review comments. It's quite a big PR to review. I'd prefer smaller ones next.

datalayer/src/main/java/com/google/android/horologist/data/DataLayerAppHelperService.kt
garanj: Sounds good. I'll take a look

datalayer-phone/src/main/java/com/google/android/horologist/datalayer/phone/PhoneDataLayerAppHelper.kt
garanj: Yeah, it's very much not ideal. Will look for a better solution

datalayer-phone/src/main/java/com/google/android/horologist/datalayer/phone/PhoneDataLayerAppHelper.kt
garanj: Great point! Will get on it!

datalayer-watch/quality/lint/lint-baseline.xml
garanj: 😳 Don't know. Will remove!

datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
garanj: Noted, on the case.

datalayer/build.gradle.kts
garanj: I think that was for RemoteActivityHelper. I will check.

datalayer/src/main/java/com/google/android/horologist/data/DataLayerAppHelper.kt
garanj: +1

datalayer/src/main/java/com/google/android/horologist/data/DataLayerAppHelper.kt
garanj: I was presuming we wanted to limit to when the devices are connected by Bluetooth... is that not the case?

datalayer/src/main/java/com/google/android/horologist/data/DataLayerAppHelper.kt
garanj: Yeah, I agree really. There was a scenario from where the app in question wasn't able to use a URI in some scenarios... I think I should go back to that and better understand if and why that is the case.

(I was trying to support that use case because of that app, so I think there's an action for me to validate that use case a bit further)

datalayer/build.gradle.kts
yschimke: We should probably suppress, and make the code internally check before using.

PR #1034, opened 2023-02-23
Add loading state to TrackPositionUiModel for the progress ring to animate

laiyichin:
#### WHAT
Add loading animation to progress ring

#### WHY
Instead of only show Play / Pause, we should show when the PlayerState is loading.

#### HOW
- Add a loading state to TrackPositionUiModel and let the Mapper map to it when PlayerState is Loading. In PlayPauseButton, does not assign "progress" to the CircularProgressIndicator to allow it animate loading state.
- Add Preview for the animated loading state
- Add test for mapper

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/components/PlayPauseButton.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/components/PlayPauseButton.kt
@@ -107,32 +107,29 @@ public fun PlayPauseProgressButton(
         onPlayClick = onPlayClick,
         onPauseClick = onPauseClick,
         enabled = enabled,
-        playing = if (trackPositionUiModel.isLoading) true else playing,
+        playing = playing,
         modifier = modifier,
         colors = colors,
         iconSize = iconSize,
         tapTargetSize = tapTargetSize,
         backgroundColor = backgroundColor
     ) {
-        if (trackPositionUiModel.showProgress) {
-            if (trackPositionUiModel.isLoading) {
-                CircularProgressIndicator(
-                    modifier = Modifier.fillMaxSize(),
-                    indicatorColor = progressColour,
-                    trackColor = trackColor,
-                    strokeWidth = progressStrokeWidth
-                )
-            } else {
-                val progress by ProgressStateHolder.fromTrackPositionUiModel(trackPositionUiModel)
-
-                CircularProgressIndicator(
-                    modifier = Modifier.fillMaxSize(),
-                    progress = progress,
-                    indicatorColor = progressColour,
-                    trackColor = trackColor,
-                    strokeWidth = progressStrokeWidth
-                )
-            }
+        val progress by ProgressStateHolder.fromTrackPositionUiModel(trackPositionUiModel)
+        if (trackPositionUiModel.showProgress && trackPositionUiModel.isLoading) {
+            CircularProgressIndicator(
+                modifier = Modifier.fillMaxSize(),
+                indicatorColor = progressColour,
+                trackColor = trackColor,
+                strokeWidth = progressStrokeWidth
+            )
+        } else if (trackPositionUiModel.showProgress) {
+            CircularProgressIndicator(
+                modifier = Modifier.fillMaxSize(),
+                progress = progress,
+                indicatorColor = progressColour,
+                trackColor = trackColor,
+                strokeWidth = progressStrokeWidth
+            )
         }
     }
 }
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/components/PlayPauseButton.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/components/PlayPauseButton.kt
@@ -115,7 +115,7 @@ public fun PlayPauseProgressButton(
         backgroundColor = backgroundColor
     ) {
         val progress by ProgressStateHolder.fromTrackPositionUiModel(trackPositionUiModel)
-        if (trackPositionUiModel.showProgress && trackPositionUiModel.isLoading) {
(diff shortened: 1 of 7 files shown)


media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/TrackPositionUiModel.kt
```diff
@@ -36,17 +38,21 @@ public sealed class TrackPositionUiModel {
...
         public val position: Duration,
-        public override val shouldAnimate: Boolean = false
+        public override val shouldAnimate: Boolean = false,
+        public override val isLoading: Boolean = false
     ) : TrackPositionUiModel() {
         override val showProgress: Boolean get() = true
 
         public companion object {
-            public val ZERO: Actual = Actual(0f, Duration.ZERO, Duration.ZERO)
+            public val ZRO: Actual = Actual(0f, Duration.ZERO, Duration.ZERO)
```
yschimke, line 47:
```suggestion
            public val ZERO: Actual = Actual(0f, Duration.ZERO, Duration.ZERO)
```

yschimke:
I'll leave the actual review to @fstanis but LGTM. I do wonder whether the animation design should just jump between these states, if that is glitchy. But seems an improvement on what we have.

media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/TrackPositionUiModel.kt
laiyichin: Fixed.

media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/TrackPositionUiModel.kt
```diff
@@ -36,17 +38,21 @@ public sealed class TrackPositionUiModel {
...
         public val position: Duration,
-        public override val shouldAnimate: Boolean = false
+        public override val shouldAnimate: Boolean = false,
+        public override val isLoading: Boolean = false
     ) : TrackPositionUiModel() {
         override val showProgress: Boolean get() = true
 
         public companion object {
             public val ZERO: Actual = Actual(0f, Duration.ZERO, Duration.ZERO)
+            public val LOADING: Actual =
```
fstanis, line 48: I think this should be a top-level instance like `Hidden`, i.e. `public object Loading : TrackPositionUiModel() { override val isLoading get() = true ... }` and not a special case of `Actual`.

media-ui/src/main/java/com/google/android/horologist/media/ui/components/PlayPauseButton.kt
```diff
@@ -106,21 +107,32 @@ public fun PlayPauseProgressButton(
...
         enabled = enabled,
-        playing = playing,
+        playing = if (trackPositionUiModel.isLoading) true else playing,
         modifier = modifier,
         colors = colors,
         iconSize = iconSize,
         tapTargetSize = tapTargetSize,
         backgroundColor = backgroundColor
     ) {
-        val progress by ProgressStateHolder.fromTrackPositionUiModel(trackPositionUiModel)
```
fstanis, line 116: I believe this needs to stay here, otherwise a switch from not loading to loading won't call `fromTrackPositionUiModel` which (among other things) cancels the progress predictive coroutine and any animation in progress.

media-ui/src/main/java/com/google/android/horologist/media/ui/components/PlayPauseButton.kt
```diff
@@ -106,21 +107,32 @@ public fun PlayPauseProgressButton(
         onPlayClick = onPlayClick,
         onPauseClick = onPauseClick,
         enabled = enabled,
-        playing = playing,
+        playing = if (trackPositionUiModel.isLoading) true else playing,
```
fstanis, line 110: This isn't the correct place to change the value of `playing`, i.e. it's changing an explicit parameter based on another parameter. We should instead potentially do it on `PlayerUiState` level, if possible.

fstanis: Thinking about it further, shouldn't `playing` be independent from `loading`, i.e. the user should still be able to play/pause if something is loading?

media-ui/src/main/java/com/google/android/horologist/media/ui/components/PlayPauseButton.kt
laiyichin: Done. Moved it back.

media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/TrackPositionUiModel.kt
laiyichin: Added new Loading type under TrackPositionUiModel.

media-ui/src/main/java/com/google/android/horologist/media/ui/state/model/TrackPositionUiModel.kt
```diff
@@ -45,8 +48,18 @@ public sealed class TrackPositionUiModel {
         }
     }
 
+    public object Loading : TrackPositionUiModel() {
+        override val showProgress: Boolean
+            get() = true
```
fstanis, line 53: If you apply the above and make `isLoading` not require `showProgress`, both `showProgress` and `shouldAnimate` should be `false`.

fstanis (approved):

PR #1042, opened 2023-02-24
Volume - Wrap VolumeControls in a composed modifier

laiyichin:
#### WHAT
- Wrap volume controls in a composed modifier
- Fix RotaryInputAccumulator values when recomposed issue

#### WHY
Persist state for volume control

#### HOW
- Added rotaryVolumeControls as a Modifier
- Used composed and remember in RotaryInputAccumulator

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/RotaryVolumeControls.kt
```diff
@@ -0,0 +1,34 @@
...
+package com.google.android.horologist.compose.rotaryinput
+
+import androidx.compose.foundation.focusable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.composed
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.focus.focusRequester
+import androidx.wear.compose.foundation.RequestFocusWhenActive
+
+public fun Modifier.rotaryVolumeControls(
```
yschimke, line 26: Does this need the label volume in it? Or is it and always will be actually generic? Not sure of the right name though.

laiyichin: Updated from rotaryVolumeControls to rotaryControls.

audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
```diff
@@ -173,32 +173,37 @@ internal fun VolumeScreen(
     increaseIcon: @Composable () -> Unit = { VolumeScreenDefaults.IncreaseIcon() },
     decreaseIcon: @Composable () -> Unit = { VolumeScreenDefaults.DecreaseIcon() },
     showVolumeIndicator: Boolean = true,
-    onVolumeChangeByScroll: ((scrollPixels: Float) -> Unit)
+    onVolumeChangeByScroll: ((scrollPixels: Float) -> Unit)? = null
```
yschimke, line 176: This seems unfortunate that we are adding back the nullable, didn't we remove?

laiyichin: Yes, this reminds me of the change I actually wanted to do, was to remove this param and move it into the modifier like [PR1018](https://github.com/google/horologist/pull/1018). Thanks!

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/RotaryControls.kt
```diff
@@ -0,0 +1,34 @@
...
+package com.google.android.horologist.compose.rotaryinput
+
+import androidx.compose.foundation.focusable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.composed
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.focus.focusRequester
+import androidx.wear.compose.foundation.RequestFocusWhenActive
+
+public fun Modifier.rotaryControls(
```
fstanis, line 26: Nit: do we need this as a separate function? I wonder if we can just make it an overload of `onRotaryInputAccumulated` or a flag?

fstanis (approved):

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/RotaryControls.kt
laiyichin: Moved it to be with onRotaryInputAccumulated.

PR #1052, opened 2023-03-01
Add run configurations for the Auth Sample

NoahAndrews:
#### WHAT
Add JetBrains run configurations for the Auth Sample phone and Wear OS apps

#### WHY
The other samples have explicit run configurations saved to the repository

#### HOW
I told Android Studio to store the implicit run configurations it auto-created as project files, then compared the results to the existing run configurations, and manually added a couple of missing properties that were present on the others.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):
LGTM - thanks!

PR #1060, opened 2023-03-02
Update simple media app doc 

kul3r4:
#### WHAT
Currently the sample from the doc is outdated and does not compile

#### WHY
Improve docs

#### HOW
Fix the code and make it compiling

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


docs/simple-media-app-guide.md
```diff
@@ -121,23 +116,41 @@ init {
...
+        buttons = { }
 )
 ```
+Add `createVolumeViewModel` function to create a VolumeViewModel:
+
+```kotlin
+fun createVolumeViewModel(): VolumeViewModel {
+        val audioRepository = SystemAudioRepository.fromContext(application)
+        val vibrator: Vibrator = application.getSystemService(Vibrator::class.java)
+        //TODO Create a simpler volumeViewModel with just a context
```
luizgrp, line 148: nit: do you want to merge this comment?

docs/simple-media-app-guide.md
```diff
@@ -121,23 +116,41 @@ init {
...
-    },
-    controlButtons = { playerUiState ->
-        PodcastControlButtons(
-            playerViewModel = viewModel,
-            playerUiState = playerUiState,
-        )
-    },
-    buttons = { }
+        playerViewModel = viewModel,
+        volumeViewModel = createVolumeViewModel(),
```
luizgrp, line 128: nit: I'd suggest to create a variable for the volumeViewModel and add it under the comment on line 123 that says that ViewModels should not be created like this

luizgrp (approved):
Thanks!

docs/simple-media-app-guide.md
kul3r4: good point, done

docs/simple-media-app-guide.md
kul3r4: removed, thanks

PR #1061, opened 2023-03-02
Renames enums for datalayer error codes

garanj:
#### WHAT

Rename values in `AppHelperResultCode` to conform with naming convention.

#### WHY

The previous naming resulted in a warning because it did not conform with https://protobuf.dev/programming-guides/style/#enums

#### Checklist :clipboard:
- [ n/a] Add explicit visibility modifier and explicit return types for public declarations
- [ x] Run spotless check
- [ x] Run tests
- [ n/a] Update metalava's signature text files


luizgrp (approved):
LGTM

PR #1063, opened 2023-03-03
Refactor rememberActiveFocusRequester

laiyichin:
#### WHAT
Set mark focusRequester not required and set it to rememberActiveFocusRequester if not set

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/AccumulatedRotaryInputModifier.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/AccumulatedRotaryInputModifier.kt
@@ -24,7 +24,7 @@ import androidx.compose.ui.focus.FocusRequester
 import androidx.compose.ui.focus.focusRequester
 import androidx.compose.ui.input.rotary.RotaryScrollEvent
 import androidx.compose.ui.input.rotary.onRotaryScrollEvent
-import androidx.wear.compose.foundation.RequestFocusWhenActive
+import androidx.wear.compose.foundation.rememberActiveFocusRequester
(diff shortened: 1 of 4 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/AccumulatedRotaryInputModifier.kt
```diff
@@ -36,12 +36,12 @@ import com.google.android.horologist.compose.navscaffold.ExperimentalHorologistC
  */
 @ExperimentalHorologistComposeLayoutApi
 public fun Modifier.onRotaryInputAccumulatedWithFocus(
-    focusRequester: FocusRequester,
+    focusRequester: FocusRequester? = null,
     onValueChange: (Float) -> Unit
 ): Modifier = composed {
-    RequestFocusWhenActive(focusRequester)
+    val localFocusRequester = focusRequester ?: rememberActiveFocusRequester()
```
yschimke, line 42: Why not inline as the default value, making it non nullable.

laiyichin: `rememberActiveFocusRequester()` is a composable but fun Modifier.onRotaryInputAccumulatedWithFocus(...) is not. I got the "Functions which invoke @Composable functions must be marked with the @Composable annotation".

I was going to mark the extension composable but it seems like the Compose team is not recommending it. (https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#composed-modifiers "SHOULD NOT declare modifier extension factory functions as @Composable functions themselves")

yschimke (approved):

PR #1067, opened 2023-03-06
Fix Data Layer capabilities

yschimke:
#### WHAT

Fix Data Layer capabilities

#### WHY

There was a bug in naming. horologist_wear vs horologist_watch.

Now prefer horologist_phone and horologist_watch.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


datalayer-phone/src/res/values/wear.xml
```diff
@@ -3,5 +3,7 @@
     <string-array name="android_wear_capabilities">
         <!-- Used to indicate that the app is installed on this device -->
         <item>data_layer_app_helper_device_phone</item>
+        <!-- Used to indicate the device is a phone -->
```
luizgrp, line 6: is the one above in use? same for `data_layer_app_helper_device_watch`

auth-sample-wear/src/main/res/values/wear.xml
```diff
@@ -1,25 +0,0 @@
...
-     limitations under the License.
--->
-
-<resources xmlns:tools="http://schemas.android.com/tools">
-    <string-array
-        name="android_wear_capabilities"
-        translatable="false"
-        tools:ignore="UnusedResources">
-        <!-- declaring the provided capabilities -->
-        <item>horologist_wear</item>
```
luizgrp, line 23: this capability is also in `sample/src/main/res/value/wear.xml`

luizgrp (approved):

datalayer-phone/src/res/values/wear.xml
yschimke: Around here https://github.com/google/horologist/blob/4ba21e728db803308ce4967b0048ba99e3ba6153/datalayer/src/main/java/com/google/android/horologist/data/DataLayerAppHelper.kt#L157

auth-data-phone/src/main/AndroidManifest.xml
```diff
@@ -19,7 +19,5 @@
     xmlns:tools="http://schemas.android.com/tools"
     package="com.google.android.horologist.auth.data.phone">
 
-    <uses-feature android:name="android.hardware.type.watch" />
```
yschimke, line 22: @luizgrp this ok?

luizgrp: yes, thanks for that!

PR #1073, opened 2023-03-08
Datalayer complications

garanj:
#### WHAT

Adds Complication activation/deactivation status to the App Helper

#### WHY

Aids apps in determining whether their Wear app has one of its Complications added to the current watch face.

#### HOW

Provides a set of `installedComplications` in `AppHelperNodeStatus` and guidance on adding necessary hooks to `ComplicationDataSourceService`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
+++ b/datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
@@ -83,22 +83,22 @@ public class WearDataLayerAppHelper(context: Context, private val appStoreUri: S
     public suspend fun markTileAsRemoved(tileName: String): Unit = markSurfaceAsRemoved(tilePrefix, tileName)
 
     /**
-     * Marks a complication as installed. Call this in
+     * Marks a complication as active on the current watch face. Call this in
      * [ComplicationDataSourceService#onComplicationActivated]. Supplying a name is mandatory to
      * disambiguate from the installation or removal of other complications your app may have.
      *
      * @param complicationName The name of the complication.
      */
-    public suspend fun markComplicationAsInstalled(complicationName: String): Unit = markSurfaceAsInstalled(complicationPrefix, complicationName)
+    public suspend fun markComplicationAsActivated(complicationName: String): Unit = markSurfaceAsInstalled(complicationPrefix, complicationName)
 
     /**
-     * Marks a complication as removed. Call this in
+     * Marks a complication as deactivated. Call this in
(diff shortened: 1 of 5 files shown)


docs/datalayer-helpers-guide.md
```diff
@@ -125,4 +126,20 @@ phone.
 
     ```kotlin
     wearAppHelper.markTileAsRemoved("SummaryTile")
+    ```
+
+1.  **Tracking Complication installation** (Wear-only)
```
yschimke, line 131: Could we add this to media-sample complication also? Another sample. I realise it doesn't currently have a phone companion, but could in future. And it would help to have a real code sample.

datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
```diff
@@ -72,25 +74,54 @@ public class WearDataLayerAppHelper(context: Context, private val appStoreUri: S
...
+    public suspend fun markTileAsRemoved(tileName: String): Unit = markSurfaceAsRemoved(tilePrefix, tileName)
+
+    /**
+     * Marks a complication as active on the current watch face. Call this in
+     * [ComplicationDataSourceService#onComplicationActivated]. Supplying a name is mandatory to
+     * disambiguate from the installation or removal of other complications your app may have.
+     *
+     * @param complicationName The name of the complication.
+     */
+    public suspend fun markComplicationAsActivated(complicationName: String): Unit = markSurfaceAsInstalled(complicationPrefix, complicationName)
```
yschimke, line 94: IT seems a shame to drop type, should we include that also?

```
    public open fun onComplicationActivated(complicationInstanceId: Int, type: ComplicationType) {}
```

Also would complicate the account for it?

```
    public open fun onComplicationActivated(complicationInstanceId: Int, type: ComplicationType) {}
```

So maybe not possible?

yschimke (approved):

docs/datalayer-helpers-guide.md
yschimke: Also Tiles?

datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
yschimke: You are not considering multiple also.

garanj: Great spot added #1074

PR #1075, opened 2023-03-08
Add documentation for auth token sharing

luizgrp:
#### WHAT

Add documentation for token sharing solution using Horologist Auth libraries.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/docs/auth-tokenshare-guide.md
+++ b/docs/auth-tokenshare-guide.md
@@ -50,15 +50,20 @@ your phone and watch apps must:
    )
    ```
 
+   This class should be created as a singleton in your app.
+
 3. Define the data to be transferred
 
    Define which authentication data that should be transferred from the phone to the watch. It can
-   be a data class with many properties. For this guide, we will pass a simple `String` instance.
+   be a data class with many properties, it can also be a [protocol buffer](https://protobuf.dev/).
+   For this guide, we will pass a simple `String` instance.
 
 4. Create a `Serializer` for the data
 
-   Create a `Serializer` class for the class defined to be transferred from the phone to
-   the watch (`String` for this guide):
+   Create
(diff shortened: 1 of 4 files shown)


docs/auth-tokenshare-guide.md
```diff
@@ -0,0 +1,124 @@
...
+   ```groovy
+   dependencies {
+       implementation "com.google.android.horologist:horologist-auth-data:<version>"
+       implementation "com.google.android.horologist:horologist-datalayer:<version>"
+   }
+   ```
+
+2. Create a `WearDataLayerRegistry`
+
+   In both projects, create an instance
```
yschimke, line 42: Should we suggest in a singleton instance?

docs/auth-tokenshare-guide.md
```diff
@@ -0,0 +1,124 @@
...
+   ```
+
+3. Define the data to be transferred
+
+   Define which authentication data that should be transferred from the phone to the watch. It can
+   be a data class with many properties. For this guide, we will pass a simple `String` instance.
+
+4. Create a `Serializer` for the data
+
+   Create a `Serializer` class for the class defined to be transferred from the phone to
```
yschimke, line 60: Maybe say androidx DataStore `Serializer`

docs/auth-tokenshare-guide.md
```diff
@@ -0,0 +1,124 @@
...
+   from the [datalayer](datalayer.md):
+
+   ```kotlin
+   val registry = WearDataLayerRegistry.fromContext(
+       application = // application context,
+       coroutineScope = // a coroutine scope
+   )
+   ```
+
+3. Define the data to be transferred
```
yschimke, line 53: Do we want to mention protobuf as an option?

PR #1082, opened 2023-03-09
Remove unused dependency from datalayer

luizgrp:
#### WHAT

1) Remove `wear-phone-interactions` dependency from `datalayer` module.

2) Decrease minSdk to 23 

#### WHY

1) Doesn't seem to be in use.

2) It is the minSdk of `androidx.wear:wear-remote-interactions:1.0.0`.

Partially addresses https://github.com/google/horologist/pull/1032#discussion_r1115715339

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/build.gradle.kts
+++ b/datalayer/build.gradle.kts
@@ -29,7 +29,7 @@ android {
     compileSdkPreview = "UpsideDownCake"
 
     defaultConfig {
-        minSdk = 25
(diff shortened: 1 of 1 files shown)


yschimke (approved):
Nice!  We should still reduce minSdk to 21 or lower, I think.

Issue #1074, opened 2023-03-08
DataLayer App Helper should correctly handle complication installed more than once on the watch face


garanj:

yschimke:
Closing as low priority, reopen if keen to work on it.

PR #1085, opened 2023-03-09
Move DataLayer app helpers to using WearDataLayerRepository

garanj:
#### WHAT

- Move `WearDataLayerAppHelper` and `PhoneDataLayerAppHelper` to using`WearDataLayerRepository`.
- Addresses issue #1074 - Complication activation status now stores also the type and instance id.

#### WHY

- Move to using a centralized access to Data Layer

#### HOW

- Make use of `WearDataLayerRepository`.
- This includes restructuring `AppHelperNodeStatus` to not have separate tiles and complications properties, backed by capabilites, but instead a `surfacesInfo`, defined via proto (so extensible to watch faces and other surfaces).

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


datalayer-phone/src/main/java/com/google/android/horologist/datalayer/phone/PhoneDataLayerListenerService.kt
```diff
@@ -16,9 +16,18 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 
 public class PhoneDataLayerListenerService : DataLayerAppHelperService() {
-    public override val appHelper: DataLayerAppHelper by lazy { PhoneDataLayerAppHelper(this) }
+    private val serviceScope = CoroutineScope(Dispatchers.IO + Job())
+
+    public override val appHelper: DataLayerAppHelper by lazy {
+        val registry = WearDataLayerRegistry.fromContext(this, serviceScope)
```
yschimke, line 30: This should ideally come from a singleton.  Assuming we want this to just work, we might have to come up with some pattern for this. I'm imagining like Coil ImageLoaderFactory does.

To follow up.

Fine for now.

datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
```diff
@@ -35,30 +43,23 @@ const val TAG = "DataLayerAppHelper"
  * Provides utility functions for determining installation status and means to install and launch
  * apps as part of the user's journey.
  */
-abstract class DataLayerAppHelper(protected val context: Context) {
+abstract class DataLayerAppHelper(
+    protected val context: Context,
```
yschimke, line 47: Let's piggy back the context in the WearDataLayerRegistry to avoid this additional param.

datalayer/src/main/proto/app_helper_pb.proto
```diff
@@ -36,4 +36,15 @@ message CompanionConfig {
...
+}
+
+message ComplicationInfo {
+  string name = 1;
+  int32 instanceId = 2;
+  string type = 3;
+}
+
+message SurfacesInfo {
+  repeated string tileNames = 1;
```
yschimke, line 48: I'd suggest putting this as tiles, and having a TileInfo also, even just with a single field. You can evolve from primitive to a message. And you always want to eventually.

datalayer/src/main/proto/app_helper_pb.proto
```diff
@@ -36,4 +36,15 @@ message CompanionConfig {
 message ActivityConfig {
   string packageName = 1;
   string classFullName = 2;
+}
+
+message ComplicationInfo {
```
yschimke, line 41: Much better design. This is actually really useful generally.

datalayer/src/main/proto/app_helper_pb.proto
```diff
@@ -36,4 +36,15 @@ message CompanionConfig {
...
+  string name = 1;
+  int32 instanceId = 2;
+  string type = 3;
+}
```
yschimke, line 45: What about a timestamp as well?

datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
```diff
@@ -68,8 +69,7 @@ abstract class DataLayerAppHelper(protected val context: Context) {
                 id = it.id,
                 displayName = it.displayName,
                 isAppInstalled = allInstalledNodes.contains(it.id),
-                installedTiles = nodesToTiles[it.id] ?: setOf(),
-                installedComplications = nodesToComplications[it.id] ?: setOf(),
+                surfacesInfo = getSurfaceStatus(it.id),
```
yschimke, line 72: Fine for now, but it's feeling like connectedNodes is doing a little bit too much.

datalayer/src/main/proto/app_helper_pb.proto
yschimke: We should build this into the tiles and complication support automatically.

datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
```diff
@@ -74,54 +89,87 @@ public class WearDataLayerAppHelper(context: Context, private val appStoreUri: S
...
-
-    private suspend fun markSurfaceAsRemoved(surfacePrefix: String, name: String) {
-        require(name.isNotEmpty())
-        try {
-            capabilityClient.removeLocalCapability("$surfacePrefix$name").await()
-        } catch (e: ApiException) {
-            if (e.statusCode != WearableStatusCodes.UNKNOWN_CAPABILITY) {
-                throw e
+            info.copy {
+                if (complications.contains(complication)) {
```
yschimke, line 167: Feels like this could be cleaner. But not sure yet, what. Land as I might try to work out nicest pattern.  No issue really, just style.

yschimke (approved):

datalayer-phone/src/main/java/com/google/android/horologist/datalayer/phone/PhoneDataLayerListenerService.kt
garanj: Yeah, agree. I did see this and thought ergh. If there's a way to do this and also avoid the developer needing to subclass the service (and add XML to the manifest) etc, that'd be ideal. That's the nice thing about right now.

datalayer/src/main/proto/app_helper_pb.proto
garanj: Would that be timestamp of activation?

datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
garanj: Could switch to a map? I don't think proto3 has sets? If that was the cleanup you were referring to?

datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
garanj: Cool.

garanj: Looking at it, WearDataLayerRegistry doesn't keep hold of the context passed in via `fromContext()`. Are you suggesting changing WearDataLayerRegistry to have a context property?

datalayer/src/main/proto/app_helper_pb.proto
garanj: Nice

datalayer-phone/src/main/java/com/google/android/horologist/datalayer/phone/PhoneDataLayerListenerService.kt
```diff
@@ -16,9 +16,18 @@
...
+import com.google.android.horologist.data.WearDataLayerRegistry
+import com.google.android.horologist.data.apphelper.DataLayerAppHelper
+import com.google.android.horologist.data.apphelper.DataLayerAppHelperService
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 
 public class PhoneDataLayerListenerService : DataLayerAppHelperService() {
-    public override val appHelper: DataLayerAppHelper by lazy { PhoneDataLayerAppHelper(this) }
+    private val serviceScope = CoroutineScope(Dispatchers.IO + Job())
```
luizgrp, line 27: if this is meant to have the scope of this service, then it should be "cancelled" on the onDestroy event of this service

if the task should execute even after the destruction of this service, then I think having the coroutine scope hanging to this service will prevent the service being destroyed while the task is being executed and cause memory leak

datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
```diff
@@ -137,32 +148,18 @@ abstract class DataLayerAppHelper(protected val context: Context) {
...
-        return idToSurfaceSet
-    }
-
     public companion object {
         public const val DATA_LAYER_APP_HELPER_CAPABILITY: String = "data_layer_app_helper"
         public const val CAPABILITY_DEVICE_PREFIX = "${DATA_LAYER_APP_HELPER_CAPABILITY}_device"
         public const val PHONE_CAPABILITY = "${CAPABILITY_DEVICE_PREFIX}_phone"
         public const val WATCH_CAPABILITY = "${CAPABILITY_DEVICE_PREFIX}_watch"
         public const val LAUNCH_APP: String = "/launch_app"
+        public const val SURFACE_INFO_PATH: String = "/surface_info"
```
luizgrp, line 162: question: should we prefix these paths (with "horologist" or the package name of the lib, etc) to prevent unintentional clashes with the paths used by the app using this lib?

I did not check the code fully so apologies if this is already being prefixed with the DATA_LAYER_APP_HELPER_CAPABILITY above

luizgrp (commented):
hey just commenting on two things, I did not fully review the PR

PR #1087, opened 2023-03-10
Update TargetNodeId.kt

yschimke:
#### WHAT

Update TargetNodeId.kt docs

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):
thanks!

PR #1089, opened 2023-03-10
Improve TokenBundleRepositoryImpl to check if API is available

luizgrp:
#### WHAT

Improve `TokenBundleRepositoryImpl` to check if DataClient API is available.

#### WHY

In order to handle the following exception when device doesn't have the API available:
```
com.google.android.gms.common.api.ApiException: 17: API: Wearable.API is not available on this device.
```

#### HOW

- Delay creation of `DataStore` in `TokenBundleRepositoryImpl` until `update` function is called;
- Change `update` to check for api availability and fail silently when not available;
- Add `isAvailable` function to `TokenBundleRepository` so developers can check if the repository can be used on the device running;
- Improve sample app to show usage of new function;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenBundleRepositoryImpl.kt
+++ b/auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenBundleRepositoryImpl.kt
@@ -16,56 +16,75 @@
 
 package com.google.android.horologist.auth.data.phone.tokenshare.impl
 
+import android.util.Log
 import androidx.datastore.core.DataStore
 import androidx.datastore.core.Serializer
+import com.google.android.gms.common.GoogleApiAvailability
+import com.google.android.gms.common.api.AvailabilityException
 import com.google.android.horologist.auth.data.phone.ExperimentalHorologistAuthDataPhoneApi
+import com.google.android.horologist.auth.data.phone.common.TAG
 import com.google.android.horologist.auth.data.phone.tokenshare.TokenBundleRepository
+import com.google.android.horologist.auth.data.phone.tokenshare.impl.TokenBundleRepositoryImpl.Companion.DEFAULT_TOKEN_BUNDLE_KEY
 import com.google.android.horologist.data.WearDataLayerRegistry
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.tasks.await
 
 /**
  * Default implementation for [TokenBundleRepository].
+ *
+ * If multiple [token bundles][T] are required to be shared, specify a [key] in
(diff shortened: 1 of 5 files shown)


auth-sample-phone/src/main/java/com/google/android/horologist/auth/sample/MainActivity.kt
```diff
@@ -75,7 +84,16 @@ class MainActivity : ComponentActivity() {
                     modifier = Modifier.fillMaxSize(),
                     color = MaterialTheme.colorScheme.background
                 ) {
+                    val coroutineScope = rememberCoroutineScope()
+                    var apiAvailable by remember { mutableStateOf(false) }
```
yschimke, line 88: If it's safe to do the operations, then I think defaulting to true.  Or making this true/false/null would be better here.  Avoid glitching the screen in the hopefully typical case where it is available.

auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenBundleRepositoryImpl.kt
```diff
@@ -16,56 +16,75 @@
...
     override suspend fun update(tokenBundle: T) {
-        dataStore.updateData { tokenBundle }
+        getDataStore()?.updateData { tokenBundle }
     }
 
-    public companion object {
+    override suspend fun isAvailable(): Boolean {
+        return try {
+            GoogleApiAvailability.getInstance()
+                .checkApiAvailability(registry.dataClient)
```
yschimke, line 56: Not sure you like the God object of the registry, but if we push this up there, then it can be done once, at startup meaning it's immediately available so screen can draw correctly immediately.

yschimke (approved):

auth-sample-phone/src/main/java/com/google/android/horologist/auth/sample/MainActivity.kt
luizgrp: it makes sense, and it would be better to have this in a VM, and have proper states

this sample app is very simple, in comparison to the wear counterpart, it's in my plans to improve it further and will address it together

auth-data-phone/src/main/java/com/google/android/horologist/auth/data/phone/tokenshare/impl/TokenBundleRepositoryImpl.kt
luizgrp: it would be ideal to have these checks in WearDataLayerRegistry, not sure about having it immediately available, it would be an assumption about the time taken by play service lib to check it

raised https://github.com/google/horologist/issues/1091

PR #1096, opened 2023-03-13
Fix auth samples

luizgrp:
#### WHAT

Fix auth samples.

#### WHY

They are broken since https://github.com/google/horologist/pull/1067

#### HOW

- Add `wear.xml` file with `horologist_phone` capability to `auth-sample-phone`.
- Remove dependencies that were supposed to bring the `wear.xml` contents but don't seem to be;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
This seems worrying. I'll take a look, perhaps wear.xml aren't merged, or need config to be.

luizgrp:
merging as I need to use the samples, please feel free to revert it

PR #1098, opened 2023-03-14
Used PickerGroup in TimePicker and DatePicker

rajat4064g:
#### WHAT
Used PickerGroup in TimePicker and DatePicker

#### WHY
PickerGroup was created to handle multiple pickers in one component.

### DIFF
--- a/composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
@@ -16,6 +16,7 @@
 
 package com.google.android.horologist.composables
 
+import androidx.compose.foundation.focusable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.BoxWithConstraints
 import androidx.compose.foundation.layout.Column
@@ -35,9 +36,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.focus.FocusRequester
@@ -51,8 +50,11 @@ import androidx.compose.ui.unit.dp
 import androidx.wear.compose.material.Button
 import androidx.wear.compose.material.Icon
 import androidx.wear.compose.material.MaterialTheme
+import androidx.wear.compose.material.PickerGroup
+import androidx.wear.compose.material.PickerGroupState
 import androidx.wear.compose.material.PickerState
 import androidx.wear.compose.material.Text
+import androidx.wear.compose.material.rememberPickerGroupState
 import java.time.LocalDate
 import java.time.format.DateTimeFormatter
 import java.time.temporal.TemporalAdjusters
@@ -99,20 +101,24 @@ public fun DatePicker(
             fontSize = with(LocalDensity.current) { 34.dp.toSp() }
         )
     )
-    val talkbackEnabled by rememberTalkBackState()
+    val touchExplorationStateProvider = DefaultTouchExplorationStateProvider()
+    val touchExplorationServicesEnabled by touchExplorationStateProvider
+        .touchExplorationState()
 
     MaterialTheme(typography = typography) {
-        var focusedElement by remember {
-            mutableStateOf(
-                if (talkbackEnabled) {
-                    FocusableElementDatePicker.NONE
-                } else FocusableElementDatePicker.DAY
-            )
+        val pickerGroupState = if (touchExplorationServicesEnabled) {
+            rememberPickerGroupState(FocusableElementDatePicker.NONE.index)
+        } else {
+            rememberPickerGroupState(FocusableElementDatePicker.DAY.index)
         }
-        val focusRequesterDay = remember { FocusRequester() }
-        val focusRequesterMonth = remember { FocusRequester() }
-        val focusRequesterYear = remember { FocusRequester() }
+        val textStyle = MaterialTheme.typography.display3
+        val optionColor = MaterialTheme.colors.secondary
         val focusRequesterConfirmButton = remember { FocusRequester() }
+
+        val yearString = stringResource(R.string.horologist_picker_year)
+        val monthString = stringResource(R.string.horologist_picker_month)
+        val dayString = stringResource(R.string.horologist_picker_day)
+        val zeroString = stringResource(R.string.horologist_picker_zero_content_description)
 
         LaunchedEffect(
             datePickerState.yearState.selectedOption,
@@ -127,56 +133,46 @@ public fun DatePicker(
         }
         val shortMonthNames = remember { getMonthNames("MMM") }
         val fullMonthNames = remember { getMonthNames("MMMM") }
-        val yearContentDescription by remember(focusedElement, datePickerState.currentYear()) {
-            derivedStateOf {
-                createDescriptionDatePicker(
-                    focusedElement,
-                    datePickerState.currentYear(),
-                    "${datePickerState.currentYear()}"
-                )
-            }
-        }
-        val monthContentDescription by remember(focusedElement, datePickerState.currentMonth()) {
-            derivedStateOf {
-                createDescriptionDatePicker(
-                    focusedElement,
-                    datePickerState.currentMonth(),
-                    fullMonthNames[(datePickerState.currentMonth() - 1) % 12]
-                )
-            }
-        }
-        val dayContentDescription by remember(focusedElement, datePickerState.currentDay()) {
-            derivedStateOf {
-                createDescriptionDatePicker(
-                    focusedElement,
-                    datePickerState.currentDay(),
-                    "${datePickerState.currentDay()}"
-                )
-            }
-        }
-
-        BoxWithConstraints(
-            modifier = modifier
-                .fillMaxSize()
-                .then(
-                    if (talkbackEnabled) {
-                        when (focusedElement) {
-                            FocusableElementDatePicker.DAY ->
-                                Modifier.scrollablePicker(datePickerState.dayState)
-
-                            FocusableElementDatePicker.MONTH ->
-                                Modifier.scrollablePicker(datePickerState.monthState)
-
-                            FocusableElementDatePicker.YEAR ->
-                                Modifier.scrollablePicker(datePickerState.yearState)
-
-                            else -> Modifier
-                        }
-                    } else {
-                        Modifier
-                    }
-                )
+        val yearContentDescription by remember(
+            pickerGroupState.selectedIndex,
+            datePickerState.currentYear()
         ) {
+            derivedStateOf {
+                createDescriptionDatePicker(
+                    pickerGroupState,
+                    datePickerState.currentYear(),
+                    yearString,
+                    zeroString
+                )
+            }
+        }
+        val monthContentDescription by remember(
+            pickerGroupState.selectedIndex,
+            datePickerState.currentMonth()
+        ) {
+            derivedStateOf {
+                if (pickerGroupState.selectedIndex == FocusableElementDatePicker.NONE.index) {
+                    monthString
+                } else {
+                    fullMonthNames[(datePickerState.currentMonth() - 1) % 12]
+                }
+            }
+        }
+        val dayContentDescription by remember(
+            pickerGroupState.selectedIndex,
+            datePickerState.currentDay()
+        ) {
+            derivedStateOf {
+                createDescriptionDatePicker(
+                    pickerGroupState,
+                    datePickerState.currentDay(),
+                    dayString,
+                    zeroString
+                )
+            }
+        }
+
+        BoxWithConstraints(modifier = modifier.fillMaxSize()) {
             val boxConstraints = this
             Column(
                 verticalArrangement = Arrangement.Center,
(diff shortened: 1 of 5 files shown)


composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -99,21 +101,25 @@ public fun DatePicker(
             fontSize = with(LocalDensity.current) { 34.dp.toSp() }
         )
     )
-    val talkbackEnabled by rememberTalkBackState()
+    val touchExplorationStateProvider = DefaultTouchExplorationStateProvider()
```
yschimke, line 104: should this be remembered?

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -99,21 +101,25 @@ public fun DatePicker(
...
+        .touchExplorationState()
 
     MaterialTheme(typography = typography) {
-        var focusedElement by remember {
-            mutableStateOf(
-                if (talkbackEnabled) {
-                    FocusableElementDatePicker.NONE
-                } else FocusableElementDatePicker.DAY
-            )
+        val pickerGroupState = if (touchExplorationServicesEnabled) {
```
yschimke, line 109: Can we explain this with a comment, what does Touch Exploration mean here?

yschimke: Should more of this logic be moving inside of Wear Material PickerGroup? Seems strange that we require so much extra logic outside of the PickerGroup.

yschimke: OR maybe we need a PickerDialog?

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -663,158 +552,192 @@ private fun Separator(width: Dp, textStyle: TextStyle) {
...
+            modifier = Modifier
+                .align(Alignment.Center)
+                .wrapContentSize()
+        )
     }
-    var talkbackEnabled = remember {
-        mutableStateOf(accessibilityManager.isEnabled && touchExplorationEnabled)
+}
+
+internal class DefaultTouchExplorationStateProvider : TouchExplorationStateProvider {
```
yschimke, line 601: Can we delete this from here and use from Wear Material? 
If not, do we have tracking bugs to surface that?

yschimke:
Since you are having some bugs with interactions, I'd suggest including those in the androidTest.

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -203,85 +199,86 @@ public fun DatePicker(
...
                 ) {
-                    if (focusedElement.index < 2) {
-                        DatePickerImpl(
-                            state = datePickerState.dayState,
-                            readOnly = focusedElement != FocusableElementDatePicker.DAY,
+                    PickerGroup(
+                        PickerGroupItemWithRSB(
+                            pickerState = datePickerState.dayState,
+                            modifier = Modifier.size(dayWidth, 100.dp),
+                            focusRequester = remember { FocusRequester() },
```
yschimke, line 235: It looks like if you leave this out, it should try do the right thing internally. So maybe omit this param.

Or do we need to call this elsewhere?

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -376,69 +345,71 @@ public fun TimePickerWith12HourClock(
...
+        val pmString = stringResource(R.string.horologist_time_picker_pm)
+        val periodContentDescription by remember(
+            pickerGroupState.selectedIndex,
+            periodState.selectedOption
+        ) {
             derivedStateOf {
-                if (focusedElement == FocusableElement12Hour.NONE) {
-                    createDescription12Hour(focusedElement, periodState.selectedOption, periodString)
+                if (pickerGroupState.selectedIndex == FocusableElement12Hour.NONE.index) {
+                    periodString
```
BowerSteve, line 405: I hear this being announced as 'full stop' rather than period (presumbably a english translation of the american 'period'!)

rajat4064g: This needs to be fixed by translation.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -126,70 +116,92 @@ public fun TimePicker(
...
+        val pickerOption = pickerTextOption(textStyle) { "%02d".format(it) }
         val focusRequesterConfirmButton = remember { FocusRequester() }
 
         val hourString = stringResource(R.string.horologist_time_picker_hour)
         val minuteString = stringResource(R.string.horologist_time_picker_minute)
         val secondString = stringResource(R.string.horologist_time_picker_second)
+        val zeroString = stringResource(R.string.horologist_picker_zero_content_description)
 
-        val hourContentDescription by remember {
+        val hourContentDescription by remember(
```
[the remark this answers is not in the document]
rajat4064g: Will add plural string resource

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -126,70 +116,92 @@ public fun TimePicker(
...
+                createDescription(
+                    pickerGroupState,
+                    hourState.selectedOption,
+                    hourString,
+                    zeroString
+                )
             }
         }
-        val minuteContentDescription by remember {
+        val minuteContentDescription by remember(
```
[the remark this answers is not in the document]
rajat4064g: Will add plural string resource

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
rajat4064g: focusRequester is given as a param to PickerGroupItem()

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -99,21 +101,25 @@ public fun DatePicker(
...
         val focusRequesterConfirmButton = remember { FocusRequester() }
 
+        val yearString = stringResource(R.string.horologist_picker_year)
+        val monthString = stringResource(R.string.horologist_picker_month)
+        val dayString = stringResource(R.string.horologist_picker_day)
+        val zeroString = stringResource(R.string.horologist_picker_zero_content_description)
+
         LaunchedEffect(
             datePickerState.yearState.selectedOption,
             datePickerState.monthState.selectedOption
```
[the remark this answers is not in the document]
rajat4064g: We're just changing the number of options in Picker. Scrolling to the new option is done by Picker internally.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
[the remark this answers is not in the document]
rajat4064g: Added plural string resource

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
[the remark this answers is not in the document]
rajat4064g: Added

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
rajat4064g: Done

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
yschimke: OK, can we internalise that functionality in PickerGroup then and not have it here.

We can say "We don't want to surface this from Wear Material since it is not recommended to have logic based on whether the accessibility services are enabled or not."

But the reality seems to be if you want to build a Date or Time Picker or similar you need to, this will get copy and pasted into other pickers, if it's not solved lower.

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
yschimke: Yep, but if null, it handles it internally, possibly more correctly.

rajat4064g: Removed focusRequester

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
rajat4064g: We can't generalize that the other use cases of PickerGroup will be using talkback mode. Hence, we're keeping the logic of identifying if the device is in talkback mode(or not) here.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
AnkitAti: I think, the exact requirement of `PickerGroup` may depend on the context of the usage. And maybe we can push that burden out to caller and they can determine how they want to find the accessibility service status, if at all they need it. I mean, even in this case, may we can add a new parameter, if that is possible, for talkback. I know that maybe a breaking change so may not be that easy to go through that route.

We do have internal implementation of the accessibility service listener in PickerGroup but it was specifically recommended by the compose folks not to make the API public as that opens the door to customisations. So, I feel even if we try to make it public, there may be strong push back on that.

composables/src/main/res/values/strings.xml
```diff
@@ -18,8 +18,22 @@
     <string description="Lets the user know that the time unit being changed is hour. Appears on the top of the TimePicker component. [CHAR_LIMIT=8]" name="horologist_time_picker_hour">Hour</string>
     <string description="Lets the user know that the time unit being changed is minute. Appears on the top of the TimePicker component. [CHAR_LIMIT=8]" name="horologist_time_picker_minute">Minute</string>
     <string description="Lets the user know that the time unit being changed is second. Appears on the top of the TimePicker component. [CHAR_LIMIT=8]" name="horologist_time_picker_second">Second</string>
+    <plurals description="Content description of the current number of hours being selected in TimePicker component. [CHAR_LIMIT=NONE]" name="horologist_time_picker_hours_content_description">
```
BowerSteve, line 21: These plurals also need a declaration in the base values ifolder to avoid crashes

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -99,21 +101,26 @@ public fun DatePicker(
             fontSize = with(LocalDensity.current) { 34.dp.toSp() }
         )
     )
-    val talkbackEnabled by rememberTalkBackState()
+    val touchExplorationStateProvider = remember { DefaultTouchExplorationStateProvider() }
```
BowerSteve, line 104: Please also add test material into DatePickerTest.kt e.g. to verify each of the contentDescriptions. It should be possible exercise each of Day/Month/Year and check that the content description is correct, like 'Day, 3' or 'March'. Otherwise, we risk regressions.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -126,70 +118,71 @@ public fun TimePicker(
         initialNumberOfOptions = 60,
         initiallySelectedOption = time.second
     )
-    val talkbackEnabled by rememberTalkBackState()
+
+    val touchExplorationStateProvider = remember { DefaultTouchExplorationStateProvider() }
```
BowerSteve, line 122: Please add tests to TimePickerTest.kt e.g. to verify content descriptions for each of Hours/Minutes/Seconds in both singular and plural.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
BowerSteve: The TouchExplorationStateProvider is only used to decide whether to select the first Picker when the screen starts. So that logic belongs in PickerGroup - and then we don't need TouchExplorationStateProvider in Horologist (or other apps).

composables/src/main/res/values/strings.xml
rajat4064g: This is the base values folder

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
AnkitAti: But wouldn't that be an assumption that picker group would be the first thing that gets focused when it is present on the screen?

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
rajat4064g: Done

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
BowerSteve: Having looked at the usage in horologist and demos in androidx, all of them repeat this code - so once this iteration for accessibility is complete, we'll look at centralising this code in PickerGroup as it's reasonable to apply the same accessibility guidelines once in that component.

BowerSteve (approved):
Thanks Rajat, looks good.

composables/src/main/res/values/strings.xml
```diff
@@ -18,8 +18,22 @@
...
+        <item quantity="one">%d Hour</item>
+        <item quantity="other">%d Hours</item>
+    </plurals>
+    <plurals description="Content description of the current number of hours being selected in TimePicker component. [CHAR_LIMIT=NONE]" name="horologist_time_picker_minutes_content_description">
```
yschimke, line 25: Nice. TIL!

yschimke (approved):
LGTM

PR #1109, opened 2023-03-15
Persist seek button order in RTL layouts

fstanis:
Per [Bidirectionality - Mirroring elements](https://m2.material.io/design/usability/bidirectionality.html#mirroring-elements):

> Do not mirror media playback buttons and the media progress indicator as they refer to the direction of tape being played, not the direction of time.

This PR forces the `Row` in `ControlButtonLayout` to always treat elements as if the layout is LTR. In effect, this means "start" and "end" always map to "left" and "right" and the first element inside is always the leftmost one, matching the parameter names.

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/components/ControlButtonLayout.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/components/ControlButtonLayout.kt
@@ -23,8 +23,11 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalLayoutDirection
+import androidx.compose.ui.unit.LayoutDirection
 import androidx.compose.ui.unit.dp
 
 @Composable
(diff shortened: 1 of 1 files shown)


luizgrp (approved):
Nice!

media-ui/src/main/java/com/google/android/horologist/media/ui/components/ControlButtonLayout.kt
```diff
@@ -34,23 +37,25 @@ public fun ControlButtonLayout(
...
 ) {
-    Row(
-        modifier = modifier.fillMaxWidth(),
-        verticalAlignment = Alignment.CenterVertically,
-        horizontalArrangement = Arrangement.Center
-    ) {
-        Box(modifier = Modifier.padding(start = 17.dp)) {
-            leftButton()
-        }
+    CompositionLocalProvider(LocalLayoutDirection provides LayoutDirection.Ltr) {
```
luizgrp, line 40: nit: it would be nice to have screenshot tests checking the UI components with this configuration.

Not entirely sure but I wonder if this could help achieve it: https://github.com/google/accompanist/pull/1414

yschimke: +1 Happy to help you implement this, possibly in follow up.  I think we should take the opportunity to add it for PlayerScreen and VolumeScreen.

Also https://github.com/cashapp/paparazzi/blob/a76702744a7f380480f323ffda124e845f2733aa/paparazzi/paparazzi-gradle-plugin/src/test/projects/layout-direction/src/test/java/app/cash/paparazzi/plugin/test/LayoutDirectionTest.kt#L31

PR #1111, opened 2023-03-15
Add new logo

fstanis:


media-sample/src/debug/res/drawable-mdpi/ic_uamp.png
luizgrp, line 1: I think for media-sample is best to check with @yschimke as I think the initial idea was to have it matching with https://github.com/android/uamp

yschimke: I'll defer to @kul3r4

yschimke (approved):
LGTM - but let's wait for Chiara to respond.

yschimke:
I'd land, and we can fix later when we want to match Uamp branding.

PR #1112, opened 2023-03-15
Allow RSB scrolling with reversed direction

Kpeved:
#### WHAT
Add reverseDirection parameter into Rotary modifiers

#### WHY
Currently the scroll is reverses when `reverseLayout`  parameter is true for ScalingLazyColumn or other Scrollable

#### HOW
That new `reverseDirection` parameter will allow the reverse of the rotary scroll direction if necessary.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
```diff
@@ -751,7 +764,7 @@ public fun Modifier.rotaryHandler(
                 channel.trySend(
                     TimestampedDelta(
                         it.uptimeMillis,
-                        it.verticalScrollPixels
+                        it.verticalScrollPixels * sign(if (reverseDirection) -1f else 1f)
```
luizgrp, line 767: question: is the call to `sign` needed?

luizgrp (approved):
LGTM

yschimke (approved):
Agree about the sign operation, but landing for now. Can follow up with any improvements.

PR #1115, opened 2023-03-16
Bump memory and version

yschimke:
#### WHAT

Bump memory and version

#### WHY

Release still failing during javadoc generation.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -34,7 +34,7 @@ systemProp.org.gradle.internal.http.socketTimeout=120000
 
 GROUP=com.google.android.horologist
 # !! No longer need to update this manually when using a Compose SNAPSHOT
-VERSION_NAME=0.3.9
+VERSION_NAME=0.3.10
 
 POM_DESCRIPTION=Utilities for Wear OS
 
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -24,7 +24,7 @@ android.useAndroidX=true
 android.disableAutomaticComponentCreation=true
 
 # Increase memory
-org.gradle.jvmargs=-Xmx4096m -XX:MaxMetaspaceSize=2048m -XX:+HeapDumpOnOutOfMemoryError
+org.gradle.jvmargs=-Xmx4096m -XX:MaxMetaspaceSize=2560m -XX:+HeapDumpOnOutOfMemoryError
```
luizgrp, line 27: 😢 are you sure? this is green https://github.com/google/horologist/actions/runs/4435405853

luizgrp (approved):

gradle.properties
yschimke: It's in the release stage, it generates javadoc to ship to sonatype.

PR #1118, opened 2023-03-16
Apphelper improvements

garanj:
#### WHAT

Addressing feedback for AppHelper
- Moves Tile from a String to a Proto, for greater future options
- Adds timestamp for Tile and Complication representation

#### WHY

Makes the information more useful and future proof

#### HOW

Updates to proto representation

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/src/main/proto/app_helper_pb.proto
+++ b/datalayer/src/main/proto/app_helper_pb.proto
@@ -39,12 +39,18 @@ message ActivityConfig {
 }
 
 message ComplicationInfo {
-  string name = 1;
-  int32 instanceId = 2;
-  string type = 3;
+  int64 timestamp = 1;
+  string name = 2;
+  int32 instanceId = 3;
+  string type = 4;
+}
+
+message TileInfo {
+  int64 timestamp = 1;
+  string name = 2;
 }
 
 message SurfacesInfo {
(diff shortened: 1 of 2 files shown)


datalayer/src/main/proto/app_helper_pb.proto
```diff
@@ -39,12 +39,18 @@ message ActivityConfig {
 }
 
 message ComplicationInfo {
-  string name = 1;
-  int32 instanceId = 2;
-  string type = 3;
+  int64 timestamp = 1;
```
yschimke, line 42: Can you revert to the existing ids? No point breaking compat for this.

datalayer/src/main/proto/app_helper_pb.proto
```diff
@@ -39,12 +39,18 @@ message ActivityConfig {
...
+}
+
+message TileInfo {
+  int64 timestamp = 1;
+  string name = 2;
 }
 
 message SurfacesInfo {
-  repeated string tileNames = 1;
+  repeated TileInfo tiles = 1;
```
yschimke, line 54:
```suggestion
  reserved 1;
  repeated TileInfo tiles = 3;
```

yschimke (approved):

PR #1119, opened 2023-03-16
Volume indicator update

laiyichin:
#### WHAT
- Make volume position indicator visible even when user tries to go beyond min or max in player screen
- When volume is already at min/max, avoid providing further haptic feedback, have the position indicator shown up instead (last point).

#### WHY
UX improvement - without showing the volume progress indicator, there is no feedback on if the RSB event is trigger and could have the RSB volume change not working impression, when actually it's just already at max or min.

#### HOW
- Modify `VolumeState` to have an extra property of `isMin` when current volume is 0
- Add a `VolumeUiState` in the `VolumeViewModel` to wrap the original `VolumeState` data class along with a timestamp. The timestamp is updated every time we attempt to update the volume (even though sometimes nothing is updated), so that compose can pick up the change in the `VolumePositionIndicator`
- Add `VolumeUiStateMapper` for converting `VolumeState` to `VolumeUiState`

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeUiState.kt
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.audio.ui
+
+import com.google.android.horologist.audio.VolumeState
+
+/*
+* A UI state for volume which contains the same information as VolumeState but with an additional
```
luizgrp, line 22: TIOLI: could make this object not coupled to VolumeState, removing mentions to it from the kdoc, and creating a [mapper](https://github.com/google/horologist/blob/1eacf11151de4fdf36715c6638c606707033ef1a/media-data/src/main/java/com/google/android/horologist/media/data/database/mapper/PlaylistEntityMapper.kt) class instead of having constructors receiving a VolumeState

luizgrp (approved):
LGTM

audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeUiState.kt
```diff
@@ -0,0 +1,45 @@
...
+    val isMax: Boolean = false
+) {
+    public constructor(volumeState: VolumeState) : this(
+        timestamp = System.currentTimeMillis(),
+        current = volumeState.current,
+        max = volumeState.max,
+        isMax = volumeState.isMax
+    )
+
+    public constructor(timestamp: Long, volumeState: VolumeState) : this(
```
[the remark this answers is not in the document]
laiyichin: Thanks! Added mapper and reduced to a single fun.

audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeViewModel.kt
```diff
@@ -52,23 +58,39 @@ public open class VolumeViewModel(
...
-        performHaptics()
+        if (!volumeState.value.isMax) {
+            performHaptics()
+        }
     }
 
     public fun decreaseVolumeWithHaptics() {
         decreaseVolume()
-        performHaptics()
+        if (volumeState.value.current != 0) {
```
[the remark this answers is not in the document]
laiyichin: Thanks! Added `isMin` to `VolumeState` and modified the usages of `current == 0` to `isMin`.

audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeViewModel.kt
```diff
@@ -52,23 +59,39 @@ public open class VolumeViewModel(
 ) : ViewModel() {
     public val volumeState: StateFlow<VolumeState> = volumeRepository.volumeState
 
+    // Keep a timestamp that always updates when we attempt ot update volume (even when it's already
```
yschimke, line 62:
```suggestion
    // Keep a timestamp that always updates when we attempt to update volume (even when it's already
```

audio/src/main/java/com/google/android/horologist/audio/VolumeState.kt
```diff
@@ -25,4 +25,7 @@ public data class VolumeState(
 ) {
     public val isMax: Boolean
         get() = current >= max
+
+    public val isMin: Boolean
```
yschimke, line 29: There is a small assumption here

https://github.com/androidx/media/blob/e78332f9f47838417d17e75208c1f06042a9e711/libraries/common/src/main/java/androidx/media3/common/DeviceInfo.java#L53-L54

```
  /** The minimum volume that the device supports. */
  public final int minVolume;
```

But I think it's ok?

audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumePositionIndicator.kt
```diff
@@ -40,16 +39,17 @@ import kotlinx.coroutines.delay
...
     autoHide: Boolean = true
 ) {
     var actuallyVisible by remember { mutableStateOf(!autoHide) }
     var isInitial by remember { mutableStateOf(true) }
 
     if (autoHide) {
-        val current = volumeState().current
-        LaunchedEffect(current) {
+        val current = volumeUiState().current
+        val timestamp = volumeUiState().timestamp
```
yschimke, line 51: nit: What actually ensures that the volumeUiState() are the same event? Will they have different timestamps if time ticks by chance in between these two statements?

The UI has a stateflow which looks good, but unclear from this code.

audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeViewModel.kt
```diff
@@ -52,23 +59,39 @@ public open class VolumeViewModel(
...
+    // at min/max)
+    private var timestamp = MutableStateFlow(System.currentTimeMillis())
+    public val volumeUiState: StateFlow<VolumeUiState> =
```
yschimke, line 65: Nice solution!

yschimke (approved):

PR #1121, opened 2023-03-20
Add auth-data-watch-oauth module

luizgrp:
#### WHAT

Add `auth-data-watch-oauth` module.

#### WHY

In order to reduce amount of dependencies in `auth-data`, that might not be required to every project that uses the lib.

#### HOW

- Move all oauth implementation code to `auth-data-watch-oauth` and leave the repositories definition in `auth-data`;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/screens/oauth/pkce/signin/PKCEViewModelFactory.kt
```diff
@@ -20,10 +20,10 @@ import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.ViewModelProvider.AndroidViewModelFactory.Companion.APPLICATION_KEY
 import androidx.lifecycle.viewmodel.initializer
 import androidx.lifecycle.viewmodel.viewModelFactory
-import com.google.android.horologist.auth.data.oauth.common.impl.google.api.GoogleOAuthServiceFactory
-import com.google.android.horologist.auth.data.oauth.pkce.impl.PKCEOAuthCodeRepositoryImpl
-import com.google.android.horologist.auth.data.oauth.pkce.impl.google.PKCEConfigRepositoryGoogleImpl
-import com.google.android.horologist.auth.data.oauth.pkce.impl.google.PKCETokenRepositoryGoogleImpl
+import com.google.android.horologist.auth.data.watch.oauth.common.impl.google.api.GoogleOAuthServiceFactory
```
yschimke, line 23: That is an impressively long package name :)

auth-data-watch-oauth/src/main/java/com/google/android/horologist/auth/data/watch/oauth/common/impl/google/api/DeviceCodeResponse.kt
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.google.android.horologist.auth.data.oauth.common.impl.google.api
+package com.google.android.horologist.auth.data.watch.oauth.common.impl.google.api
```
yschimke, line 17: What about

`com.google.android.horologist.auth.oauth.google.api`

Feels longer than I've seen before.

auth-data-watch-oauth/src/main/java/com/google/android/horologist/auth/data/watch/oauth/ExperimentalHorologistAuthDataWatchOAuthApi.kt
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.auth.data.watch.oauth
+
+@RequiresOptIn(
+    message = "Horologist Auth Data Watch OAuth is experimental. The API may be changed in the future."
+)
+@Retention(AnnotationRetention.BINARY)
+public annotation class ExperimentalHorologistAuthDataWatchOAuthApi
```
yschimke, line 23: This feels unfortunate, if we want to avoid so many annotations.  Is there a general auth one we could use instead.

#tioli

yschimke (approved):

PR #1123, opened 2023-03-21
Use a single experimental annotation

yschimke:
#### WHAT

Use a single experimental annotation

#### WHY

Simplify library usage

#### HOW

Add new annotation, use it everywhere.  Leave existing annotations but unused to minimise upgrade pain.

#### Checklist :clipboard:
- [x] Add explicit visibility modifier and explicit return types for public declarations
- [x] Run spotless check
- [x] Run tests
- [x] Update metalava's signature text files

PR #1125, opened 2023-03-22
Move opt-ins for experimental annotation to build.gradle file

luizgrp:
#### WHAT

Move opt-ins for experimental annotation to build.gradle file

#### WHY

https://github.com/google/horologist/pull/1123#discussion_r1143312738

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/audio-ui/build.gradle.kts
+++ b/audio-ui/build.gradle.kts
@@ -44,6 +44,7 @@ android {
 
     kotlinOptions {
         jvmTarget = "11"
+        freeCompilerArgs = freeCompilerArgs + "-opt-in=com.google.android.horologist.annotations.ExperimentalHorologistApi"
     }
 
     composeOptions {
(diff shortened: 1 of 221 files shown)


yschimke (approved):
This seems nicer to me, Filip should have a say as well.

What is our rule after this?
- horologist annotations in gradle
- other experimental annotations in build.gradle.kts? Source files?

luizgrp:
As far as I understand, the other ones can't just be in build.gradle, as the project is used with another build system. It only works with the horologist annotation because they are removed before importing.

yschimke:
@luizgrp so should we move them all into code? Not necessary in this PR, just curious what the exact rules are when adding more.

yschimke:
@luizgrp are we intending to land this, do you want it in the next release?  Just noticed the conflicts.

luizgrp:
yes, intend to land, not necessary in next release

waiting on review from @fstanis as per your request, not going to sort conflicts until then

PR #1126, opened 2023-03-22
Add placeholder when we lottie composition is still loading in

laiyichin:
#### WHAT
- Update PlayPause.json lottie asset 
- Add a placeholder image vector for lottie animated playpause / previous / next buttons

#### WHY
- Original PlayPause.json Play triangle was a little shifted to the left
- Lottie resources take a while to load in initial load, and UI would be missing buttons

#### HOW
- Added `ImageVector` for each corresponding lottie resource
- Modified the `AnimatedMediaButton` to take in extra optional parameter for placeholder

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
If it fixes a real issue, then absolutely go for it.

As a pattern it suggests an issue with lottie-compose. You shouldn't have to do this to get it glitch free. So we should follow up with the Lottie folk, via a github issue or a conversation in slack to find out the best options for this without extra code.

yschimke:
Other ideas to avoid defining a static image for each animation. Ensure the cache for all assets is populated before showing the screen? Delay the splashscreen slightly until the assets are loaded.

Once loaded they should be cache for the process lifetime.

media-ui/src/main/java/com/google/android/horologist/media/ui/components/animated/AnimatedPlayPauseButton.kt
```diff
@@ -128,14 +125,24 @@ public fun AnimatedPlayPauseButton(
...
-                        .graphicsLayer(alpha = LocalContentAlpha.current),
-                    composition = composition,
-                    progress = { lottieProgress.value }
-                )
+                val contentModifier = Modifier
+                    .size(iconSize)
+                    .align(Alignment.Center)
+                    .graphicsLayer(alpha = LocalContentAlpha.current)
+
+                if (isCompositionReady) {
```
fstanis, line 133: I wonder if we can move this bit (lines 133-145) into a component, something like `LottieAnimationWithPlaceholder` which accepts a `composition` and `ImageVector`. That could contain the `isCompositionReady` logic too and then we can even get rid of `LottieButtonPlaceholder`.

Reasoning being, we're now passing a lot of things to these components that are derived from each other while in reality they only need the animation and its placeholder.

fstanis (approved):

media-ui/src/main/java/com/google/android/horologist/media/ui/components/animated/AnimatedPlayPauseButton.kt
laiyichin: Done. Extracted the `LottieAnimationWithPlaceholder`.

laiyichin:
> Other ideas to avoid defining a static image for each animation. Ensure the cache for all assets is populated before showing the screen? Delay the splashscreen slightly until the assets are loaded.
> 
> Once loaded they should be cache for the process lifetime.

iirc, I am hoping by calling the `rememberLottieComposition` we will be using the [default caching mechanism](https://github.com/airbnb/lottie-android/blob/7f1e643bd9b573369f3f99332f445b1359378632/lottie-compose/src/main/java/com/airbnb/lottie/compose/rememberLottieComposition.kt#L79) although it'd still be worth checking that the key is set.

media-ui/src/main/java/com/google/android/horologist/media/ui/components/animated/LottieAnimationWithPlaceholder.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.airbnb.lottie.compose.LottieCompositionResult
+import com.airbnb.lottie.compose.LottieDynamicProperties
+
+/**
+ * A lottie animation with placeholder.
+ */
+@Composable
+public fun LottieAnimationWithPlaceholder(
+    lottieCompositionResult: LottieCompositionResult,
+    lottieAnimatable: LottieAnimatable,
```
fstanis, line 37: So, I don't think this composable needs so many lottie parameters:

- `LottieComposition?` is just calling `lottieCompositionResult.value`, so we can remove that
- `LottieAnimatable` is technically a state, which shouldn't be passed directly in compose - I'd instead have `progress: () -> Float` as a parameter (same as what the `LottieAnimation` composable accepts as a parameter).

PR #1057, opened 2023-03-01
Changes volume to support relative and absolute

fstanis:
#### WHAT
Abstracts volume control to allow implementations that support relative and absolute adjustment and both and treats volume as % in the UI.

#### WHY
To simplify UI domain and make it easier to have other volume implementations (e.g. ExoPlayer). Also fixes #739.

#### Checklist :clipboard:
- [X] Add explicit visibility modifier and explicit return types for public declarations
- [X] Run spotless check
- [ ] Run tests
- [X] Update metalava's signature text files

PR #1129, opened 2023-03-23
Removes timestamp from volume state

fstanis:
#### WHAT
Removes timestamp from `VolumeUiState` and decouples "show volume indicator" from the state.

#### WHY
To not pollute state with events and make it easier to merge #1057.

#### HOW
Adds a private flow `userActionEvents`into `VolumeViewModel` which can be emitted to to force display the indicator. Then merges this flow with `volumeUiState`, giving a unified flow of `displayIndicatorEvents`. `VolumePositionIndicator` listens to this combined flow and show the indicator for two seconds on every emit.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests

### DIFF
--- a/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumePositionIndicator.kt
+++ b/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumePositionIndicator.kt
@@ -20,47 +20,39 @@ import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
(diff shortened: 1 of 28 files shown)


audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumePositionIndicator.kt
```diff
@@ -20,47 +20,39 @@ import androidx.compose.animation.AnimatedVisibility
...
+ * Shows on the left side to match Bezel behaviour. If [displayIndicatorEvents] is non-null, the
+ * indicator is initially hidden and only displays when that flow emits, then auto-hides after two
+ * seconds.
  */
 @Composable
 public fun VolumePositionIndicator(
     volumeUiState: () -> VolumeUiState,
     modifier: Modifier = Modifier,
-    autoHide: Boolean = true
+    displayIndicatorEvents: Flow<Unit>? = null
```
yschimke, line 43: I wonder what the typical pattern is for this? Combining

a) Hide by default
b) Show on volume changes from any source
c) Show when user tries to change (may not success at max or min), from UI.

cc @BowerSteve

PR #1130, opened 2023-03-23
Mirrors volume icons in RTL

fstanis:


audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
```diff
@@ -199,7 +202,12 @@ public object VolumeScreenDefaults {
     public fun IncreaseIcon() {
         Icon(
```
yschimke, line 203: What about a StandardIcon in base-ui that has this as a Flag?

yschimke: How many icons does this likely affect, why just volume ones?

fstanis: Done. Good shout, it did increase min SDK to 26 though.

audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/actions/SettingsButton.kt
```diff
@@ -59,13 +61,14 @@ public fun SettingsButton(
...
     ) {
-        Icon(
+        StandardIcon(
             imageVector = imageVector,
             contentDescription = contentDescription,
             modifier = Modifier
                 .size(iconSize)
                 .align(Alignment.Center)
-                .semantics { iconImageVector = imageVector }
+                .semantics { iconImageVector = imageVector },
```
yschimke, line 70: nit: put into StandardIcon?

yschimke (approved):

PR #1133, opened 2023-03-24
Fix talkback highlight for song title.

RavishankarreddyB:
#### WHAT
If the title of the song is long enough that it cannot fit in the screen, then only partial title is shown and then the title scrolls. Talkback highlight also scrolls along with title and finally disappears once end of the title is reached.

#### WHY
clearAndSetSemantics will remove the already set properties of how the talkback highlights and will highlight the text space allocated to the title. This highlight will not scroll along with the title.

### DIFF
--- a/media-ui/src/main/java/com/google/android/horologist/media/ui/components/animated/MarqueeTextMediaDisplay.kt
+++ b/media-ui/src/main/java/com/google/android/horologist/media/ui/components/animated/MarqueeTextMediaDisplay.kt
@@ -21,6 +21,8 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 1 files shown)


media-ui/src/main/java/com/google/android/horologist/media/ui/components/animated/MarqueeTextMediaDisplay.kt
```diff
@@ -41,7 +43,9 @@ public fun MarqueeTextMediaDisplay(
     Column(modifier = modifier, horizontalAlignment = Alignment.CenterHorizontally) {
         MarqueeText(
             text = title.orEmpty(),
-            modifier = Modifier.fillMaxWidth(0.7f),
+            modifier = Modifier.fillMaxWidth(0.7f).clearAndSetSemantics {
```
yschimke, line 46: OK, to land.  

I would suggest that we move this into the MarqueeText, but actually I think we should aim to replace that with basicMarquee modifier.

yschimke (approved):
LGTM

PR #1139, opened 2023-03-27
Fix AnimatedPlayPauseButton Wrong State Issue

laiyichin:
#### WHAT
Fix AnimatedPlayPauseButton Wrong State Issues

#### WHY
currently,
1) On startup, when nothing is playing, AnimatedPlayPause button goes to "playing" state (which is the double pause bar)
2) On resume, the AnimatedPlayPauseButton always goes to "playing" state (which is the double pause bar)

#### HOW
There were a few reasons of these behaviours,
1) Lottie json after the update in #1126, animates in three stages: doube_bar -> triangle -> double_bar for the whole iteration, and even though we used the `LottieClipSpec` the progress was still using the whole composition to calculate (which means 0f is double bar at the beginning). I modified the json to clip the frame from 20-40 instead of 0-40 which means it'd now would only have triangle to double_bar animated.
2) When we use `animateLottieCompositionAsState` it is default to speed = 1f which means it'd have one iteration to play to the end. So setting it to 0 would stop it from playing to the end.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):

luizgrp:
nice investigation there 👍

yschimke:
@laiyichin We should consider (in a few weeks) putting together a robolectric test that verifies each of these states. Let's find some free time to add this in April.

PR #1142, opened 2023-03-28
Adding animation for header of PlayerScreen

hardikagr:
Adding transition animation for song and artist name on clicking seek next and seek previous button

Wrapped the title and artist name with "AnimatedContent" so any change in the respective state triggers the required entry and exit animation.


laiyichin (approved):
LGTM. Thank you!

PR #1143, opened 2023-03-28
Fix protobuf usage of time

yschimke:
#### WHAT

Fix protobuf usage of time

#### WHY

An internal warning of bad proto usage.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


garanj (approved):
This is great, I didn't know about this! Thanks!

PR #1144, opened 2023-03-28
Fix case for Composable function with return value

yschimke:
#### WHAT

Fix case for Composable function with return value

#### WHY

https://googlesamples.github.io/android-custom-lint-rules/checks/ComposableNaming.md.html

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


rajat4064g (approved):
LGTM

luizgrp (approved):

PR #1150, opened 2023-03-30
Makes modifier recompoose when onValueChange changes

fstanis:


compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/AccumulatedRotaryInputModifier.kt
```diff
@@ -61,7 +61,7 @@ public fun Modifier.onRotaryInputAccumulated(
     rateLimitCoolDownMs: Long = RotaryInputConfigDefaults.DEFAULT_RATE_LIMIT_COOL_DOWN_MS,
     onValueChange: (change: Float) -> Unit
 ): Modifier = composed {
-    val rotaryInputAccumulator = remember {
+    val rotaryInputAccumulator = remember(onValueChange) {
```
yschimke, line 64: you can also use `rememberUpdatedState(onValueChange)` and avoid recomposing because of this.

yschimke: onValueChange might recompose almost every frame, if you are not remembering the lambda in the caller.

yschimke:
Fixed in https://github.com/google/horologist/pull/1155/files

PR #1154, opened 2023-04-03
Cleanup robo PR from review comments

yschimke:
#### WHAT

Cleanup robo PR from review comments

#### WHY

1 bug found, and some cleanup.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/roboscreenshots/src/main/java/com/google/android/horologist/screenshots/Diffing.kt
+++ b/roboscreenshots/src/main/java/com/google/android/horologist/screenshots/Diffing.kt
@@ -31,6 +31,7 @@ import com.quickbird.snapshot.bitmap
 import okio.Buffer
 import okio.ByteString
 import java.io.File
+import kotlin.math.ceil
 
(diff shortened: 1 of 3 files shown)


roboscreenshots/src/main/java/com/google/android/horologist/screenshots/Diffing.kt
```diff
@@ -76,10 +77,12 @@ internal fun Diffing.Companion.bitmapWithTolerance(tolerance: Float, colorDiffin
                     setPixel(x, y, diffColor.toArgb())
                 }
             }
-            val diffPercent = differentCount.toDouble() / originalBytes.size
-            if (diffPercent < tolerance) {
+            val pixelCount = originalBitmap.width * originalBitmap.height
```
yschimke, line 80: original bytes was wrong because 1 pixel > 1 byte.

luizgrp (approved):

PR #1156, opened 2023-04-03
Migrate base-ui screenshot tests

yschimke:
#### WHAT

Migrate base-ui screenshot tests

#### WHY

Remove paparazzi

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


base-ui/src/test/snapshots/images/com.google.android.horologist.base.ui.components_PrimaryChipTest_disabled.png
luizgrp, line 1: it looks like there is more transparency applied to the chip background

luizgrp (approved):

base-ui/src/test/snapshots/images/com.google.android.horologist.base.ui.components_PrimaryChipTest_disabled.png
yschimke: Agreed with Filip on Friday, but this one surprised me.  We wanted the transparency to be visible, but not distracting, so went with black, 50% alpha.  Are we expecting these to have transparent chip background?

Or do you think it's an RNG bug?

cc @fstanis

luizgrp: it looks like wear compose material applies alpha to the contents of the chip when in disabled state

will check it with some sample code at some point

yschimke: So this caught something! nice.

PR #1162, opened 2023-04-03
Fix formatting

yschimke:
#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (commented):
LGTM

PR #1164, opened 2023-04-03
Adds basic example use of App Helper to the existing Auth Sample.

garanj:
#### WHAT

Adds datalayer app helper example usage to Auth Sample

#### WHY

To demonstrate basic use of the app helper library

#### HOW

Adds example of `PhoneDataLayerAppHelper` to main screen of auth sample, to demonstrate listing nodes, tiles, complications, app installation, launching of app and companion

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


auth-sample-phone/src/main/res/values/strings.xml
```diff
@@ -19,4 +19,16 @@
     <string name="token_share_button_update_token_default">Update token - default key</string>
     <string name="token_share_button_update_token_custom">Update token - custom key</string>
     <string name="token_share_message_api_unavailable">This device does not have capability to communicate to Wearable Data Layer API</string>
+
+    <string name="node_name_label">Node: %1$s</string>
```
luizgrp, line 23: nit: strings could be prefixed by the screen or feature, otherwise it gets out of control quickly as we add more features

auth-sample-phone/src/main/java/com/google/android/horologist/auth/sample/MainActivity.kt
```diff
@@ -165,11 +223,40 @@ fun MainScreen(
 @Preview(showBackground = true)
 @Composable
 fun GreetingPreview() {
+    val nodeList = listOf(
+        AppHelperNodeStatus(
+            id = "a1b2c3d4",
+            displayName = "Pixel Watch",
```
luizgrp, line 229: nit: is the preview meant to be a "watch"? just asking given it's in the "phone" sample. same for preview in AppHelperNodeStatusCard

auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/MainActivity.kt
```diff
@@ -19,10 +19,33 @@ package com.google.android.horologist.auth.sample
...
+
+        lifecycleScope.launch {
+            wearDataLayerAppHelper.markTileAsInstalled("MyTile")
+            delay(1000)
+            wearDataLayerAppHelper.markComplicationAsActivated(
+                complicationName = "MyComplication",
+                complicationInstanceId = 1234,
+                complicationType = ComplicationType.SHORT_TEXT
+            )
+        }
```
luizgrp, line 47: question (not a request): do you want this to happen on app launch, or would you rather having a separate screen that triggers them?

luizgrp (approved):
LGTM!

screenshots of the screens in the PR description would be helpful to review!

auth-sample-phone/src/main/res/values/strings.xml
garanj: Done!

auth-sample-phone/src/main/java/com/google/android/horologist/auth/sample/MainActivity.kt
garanj: Yes, it's meant to be the watch, as it lists the connected devices, so in the phone sample it will list connected watches.

auth-sample-wear/src/main/java/com/google/android/horologist/auth/sample/MainActivity.kt
garanj: I intended to have a separate screen, but thought that in the first instance it would be best to go with a simpler and smaller PR, and then add more features to the watch side (such as adding / removing complications) separately, if that works for you?

luizgrp: sounds good!

PR #1167, opened 2023-04-03
Add a saveSnapshot API method for overriding

yschimke:
#### WHAT

Add a saveSnapshot API method for overriding

#### WHY

ScreenshotTest is opinionated and default to Paparazzi naming, but allow this to be overridden. 

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):
Any further plans to evolve this API? Just wondering if it could be a rule instead of a class that needs to be inherited

PR #1174, opened 2023-04-04
Volume button

laiyichin:
#### WHAT
Default volume button to volume up icon instead of volume down when no volume ui state is passed in.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


yschimke:
Is there a broken test in SetVolumeButtonTest?

If not, can we add one?

yschimke (approved):

PR #1176, opened 2023-04-04
Fix Volume Position Indicator

laiyichin:
#### WHAT
Fix VolumePositionIndicator

#### WHY
So compose knows to swap out the `displayIndicatorEvents` for a different session.

#### HOW
Make `visible` in `VolumePositionIndicator` observe with key `displayIndicatorEvents` in `produceState`

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests

### DIFF
--- a/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumePositionIndicator.kt
+++ b/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumePositionIndicator.kt
@@ -42,7 +42,7 @@ public fun VolumePositionIndicator(
     modifier: Modifier = Modifier,
     displayIndicatorEvents: Flow<Unit>? = null
 ) {
-    val visible by produceState(false) {
+    val visible by produceState(displayIndicatorEvents == null, displayIndicatorEvents) {
         displayIndicatorEvents?.collectLatest {
             value = true
             delay(2000)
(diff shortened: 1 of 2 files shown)


audio-ui/src/test/kotlin/com/google/android/horologist/audio/ui/components/actions/SetVolumeButtonTest.kt
```diff
@@ -76,4 +76,11 @@ class SetVolumeButtonTest : ScreenshotTest() {
             )
         }
     }
+
+    @Test
+    fun givenNoVolumeUiState_thenIconIsVolumeUp() {
```
yschimke, line 81: Is this from other PR?

audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumePositionIndicator.kt
```diff
@@ -42,7 +42,7 @@ public fun VolumePositionIndicator(
     modifier: Modifier = Modifier,
     displayIndicatorEvents: Flow<Unit>? = null
 ) {
-    val visible by produceState(false) {
+    val visible by produceState(displayIndicatorEvents == null, displayIndicatorEvents) {
```
yschimke, line 45: I'm ok with this as a specific fix.  But also think it's leaking your implementation details for a Screen outside of Horologist into here. 

We should maybe consider how to represent this cleanly, and move this produceState into your app.

Maybe 
`displayIndicatorEvents: Flow<Unit>`
or
`displayIndicatorEvents: Channel<Instant>`?

fstanis (approved):

PR #1178, opened 2023-04-05
Add ScreenshotTestRule to roboscreenshots

luizgrp:
#### WHAT

Add `ScreenshotTestRule` to `roboscreenshots`.

#### WHY

In order to not force usage to have to extend a specific class.

Also improve encapsulation, as params have to be passed via constructor of the rule instead of being changed during init (or at any time).

#### HOW

- Add `ScreenshotTestRule`;
- Add convenience class `ScreenShotBaseTest` that can be used the same way that `ScreenshotTest` was being used;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp:
@yschimke TIOLI, proposing it for review before removing `ScreenshotTest` and updating all tests; changed few classes to test it, haven't checked the record functionality yet.

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenShotBaseTest.kt
```diff
@@ -0,0 +1,43 @@
...
+ * A test class that can be used as base class for tests that require a [ScreenshotTestRule].
+ */
+@RunWith(RobolectricTestRunner::class)
+@Config(
+    sdk = [30],
+    qualifiers = "w227dp-h227dp-small-notlong-round-watch-xhdpi-keyshidden-nonav"
+)
+@GraphicsMode(GraphicsMode.Mode.NATIVE)
+@ExperimentalHorologistApi
+public abstract class ScreenShotBaseTest(
```
yschimke, line 37: nit: we should be consistent on capitalisation of screenShot.

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotTestRule.kt
```diff
@@ -0,0 +1,236 @@
...
+import org.junit.rules.TestName
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+import org.robolectric.RobolectricTestRunner
+
+/**
+ * A [TestRule] that allows you to run screenshot tests of your composable components.
+ *
+ * This rule requires robolectric, so the test suite should run with [RobolectricTestRunner]. See
```
yschimke, line 63: I wonder if we can test this via the Rule api?

auth-composables/src/test/java/com/google/android/horologist/auth/composables/dialogs/SignedInConfirmationDialogTest.kt
```diff
@@ -22,19 +22,16 @@ import androidx.compose.ui.Alignment
...
-    }
+class SignedInConfirmationDialogTest : ScreenShotBaseTest() {
 
     @Test
     fun signedInConfirmationDialog() {
-        this.fakeImageLoader = FakeImageLoader.Resources
-
-        takeScreenshot {
+        screenshotTestRule.takeScreenshot(
+            fakeImageLoader = FakeImageLoader.Resources
```
yschimke, line 33: nice!

yschimke (approved):
Looks much better, thanks for tackling.

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenShotBaseTest.kt
luizgrp: thanks!

auth-composables/src/test/java/com/google/android/horologist/auth/composables/dialogs/SignedInConfirmationDialogTest.kt
luizgrp: this change could be implemented in the code as it is, hope I'm not missing anything else that I will find once I start porting the rest of the functions

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotTestRule.kt
luizgrp: you mean add the RunWith, Config and GraphicsMode via Rule? that would be the ideal! will explore it in further iteration

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotTestRule.kt
```diff
@@ -0,0 +1,236 @@
...
+    private val resources: Resources =
+        ApplicationProvider.getApplicationContext<Application>().resources
+
+    override fun apply(base: Statement, description: Description): Statement {
+        return RuleChain.outerRule(testName)
+            .around(composeContentTestRule)
+            .apply(base, description)
+    }
+
+    public fun takeScreenshot(
```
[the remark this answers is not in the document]
luizgrp: It makes sense, will explore that too. Were the tests in DatePickerA11yTest supposed to be a single test?

auth-composables/src/test/java/com/google/android/horologist/auth/composables/screens/CheckYourPhoneScreenTest.kt
```diff
@@ -16,24 +16,21 @@
 
 package com.google.android.horologist.auth.composables.screens
 
-import com.google.android.horologist.screenshots.ScreenshotTest
+import com.google.android.horologist.screenshots.ScreenShotBaseTest
 import org.junit.Test
 
-class CheckYourPhoneScreenTest : ScreenshotTest() {
-    init {
-        screenTimeText = {}
```
luizgrp, line 24: in a first version of the port, I forgot to override this param, and the test still passed, sounds like the tolerance is too high or there is something off

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotTestRule.kt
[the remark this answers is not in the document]
luizgrp: done in e7109c508b78de13a728ad90c6d33b5ba573994e, please review it and let me know 

caveat: had to stop throwing exceptions when recording, otherwise the same test would have to run as many times as it takes screenshots

audio-ui/src/test/kotlin/com/google/android/horologist/audio/ui/VolumeScreenIndividualTest.kt
```diff
@@ -37,13 +38,15 @@ class VolumeScreenIndividualTest : ScreenshotTest() {
         )
         val audioOutput = AudioOutput.BluetoothHeadset("id", "Pixelbuds")
 
-        takeScreenshot {
+        screenshotTestRule.setContent {
```
luizgrp, line 41: could potentially add a flag to also take a screenshot after setting the content, to avoid the extra call

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotTestRule.kt
yschimke: I meant something different.  I guess first question, is whether it really does require it? https://robolectric.org/androidx_test/

> It is now possible to use the AndroidX test runner in Robolectric tests. If you require a custom test runner currently

Is this because we request customisation?

But I do think I've had failures because I wasn't using the robolectric runner, and they were not obvious.  I wonder whether the rule can fail fast by just checking which runner is active via the junit APIs?

yschimke: On the second one, you can certainly change the qualifiers during the test.  Not sure about GraphicsMode, it seems a very fundamental early split, based on https://github.com/robolectric/robolectric/issues/8073

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotTestRule.kt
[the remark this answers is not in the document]
yschimke: Maybe the rule can fail right at the end, if recording mode is on?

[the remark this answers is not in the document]
yschimke: That diff looks good, as I was thinking but unable to implement.  Do we need `screenshotTestRule.interact` ?

audio-ui/src/test/kotlin/com/google/android/horologist/audio/ui/VolumeScreenIndividualTest.kt
yschimke: Your call.

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotTestRule.kt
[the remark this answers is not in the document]
luizgrp: it's needed as we don't expose the private ComposeContentTestRule, unless you think we should make it public?

auth-composables/src/test/java/com/google/android/horologist/auth/composables/screens/CheckYourPhoneScreenTest.kt
yschimke: Yep, we should work out some process that let's us 

a) discover the right values for typical screens not failing across mac and linux? maybe it's two values? screenshot tolerance, and platform tolerance?
b) Some way to get these correct, like failing if the tolerance is too high?

composables/src/test/kotlin/com/google/android/horologist/composables/DatePickerA11yTest.kt
```diff
@@ -21,52 +21,47 @@ import androidx.compose.ui.test.assertHasClickAction
...
+class DatePickerA11yTest : ScreenshotBaseTest(
+    screenshotTestRuleParams {
         screenTimeText = {}
-        enableA11yTest()
+        enableA11y = true
     }
-
+) {
     @Test
     fun initial() {
```
yschimke, line 36: let's rename initial to something like interactionTest?

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotBaseTest.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.robolectric.annotation.Config
+import org.robolectric.annotation.GraphicsMode
+
+/**
+ * A test class that can be used as base class for tests that require a [ScreenshotTestRule].
+ */
+@RunWith(RobolectricTestRunner::class)
+@Config(
+    sdk = [30],
+    qualifiers = "w227dp-h227dp-small-notlong-round-watch-xhdpi-keyshidden-nonav"
```
yschimke, line 33: Should we move these to some `const val` for devices, like WearSquare, WearLargeRound, WearSmallRound?

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotTestRule.kt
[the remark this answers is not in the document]
yschimke: Maybe we should, it would let you do other things you do with the ComposeTestRule here also.

yschimke:
LGTM

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotTestRule.kt
```diff
@@ -0,0 +1,409 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalView
+import androidx.compose.ui.test.SemanticsNodeInteraction
+import androidx.compose.ui.test.junit4.ComposeContentTestRule
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.compose.ui.test.onRoot
+import androidx.test.core.app.ApplicationProvider
+import androidx.wear.compose.material.PositionIndicator
```
yschimke, line 44: Should we split this in some way, such that this rule isn't tied to Wear? Basically thinking about an app that uses this for Wear screenshots? Can it use this also for the modules that are shared, like individual components?

Some wear specific builder?

yschimke: Just a thought, not blocking.

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotTestRule.kt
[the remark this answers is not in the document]
luizgrp: The problem is that if the user call setContent from ComposeTestRule instead of from ScreenshotTestRule then things won't work properly. That was the main reason from encapsulating ComposeTestRule and exposing only what it's needed.

The ideal would be if ScreenshotTestRule could inherit from AndroidComposeTestRule, so we could just delegate all the calls to AndroidComposeTestRule and enhance the setContent call. But AndroidComposeTestRule is not open.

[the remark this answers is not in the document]
luizgrp: Changed interact to expose ComposeTestRule instead of Compose**Content**TestRule, so the setContent is not exposed.
We could remove the interact function and create a getter that returns a ComposeTestRule, that would have the same effect, but I will leave for a future PR.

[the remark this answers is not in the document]
luizgrp: > Maybe the rule can fail right at the end, if recording mode is on?

Leaving this to be done in a another pr, could be done when adding the gradle task to do the recordings. 
Where we currently have the println we could add the path to the generated images into a list, then in the end of the tests we could print them all to std or to a file.

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotTestRule.kt
luizgrp: sounds good! given this PR is already too big, let's explore it in another one

audio-ui/src/test/kotlin/com/google/android/horologist/audio/ui/VolumeScreenIndividualTest.kt
luizgrp: seconds thoughts after merging, maybe the flag's default value should be true

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/ScreenshotBaseTest.kt
luizgrp: https://github.com/google/horologist/issues/1184

PR #1187, opened 2023-04-06
Test alternative builds

yschimke:
#### WHAT

Change to use simple build commands instead of the more complicated ones from accompanist.

#### WHY

They are too slow, suggesting something else going on.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


scripts/run-unit-tests.sh
```diff
@@ -1,64 +0,0 @@
...
-
-FILTER_OPTS=""
-# Filter out flaky tests if we're not set to run them
-if [[ -z "$RUN_FLAKY_TESTS" ]]; then
-  FILTER_OPTS="$FILTER_OPTS -Pandroid.testInstrumentationRunnerArguments.notAnnotation=androidx.test.filters.FlakyTest"
-fi
-
-# If we're set to only run affected test, update the Gradle task
-if [[ ! -z "$RUN_AFFECTED" ]]; then
-  TASK="runAffectedUnitTests"
```
luizgrp, line 50: if proceeding with not running this task anymore, this [plugin](https://github.com/google/horologist/blob/main/gradle/libs.versions.toml#L58) can be removed, which is supposed to speed up the builds

luizgrp (approved):

PR #1192, opened 2023-04-11
Ignore WearDataLayerRegistryTest because it's flaky

laiyichin:
#### WHAT
Ignore flaking test

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
Let's merge after we see it one more time.

yschimke (approved):

PR #1202, opened 2023-04-13
Adds documentation for Download

kul3r4:
#### WHAT
Adds documentation for downloading media.

#### WHY
Current documentation does not explain how progress is implemented and which type of UI screen we provide.

#### HOW
Provide some code snippets to give context.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


docs/media-data.md
```diff
@@ -28,3 +28,176 @@ internal database.
 ```
 
 [media3]: https://developer.android.com/jetpack/androidx/releases/media3
+
+### Media Toolkit implementation
+
+#### DownloadManagerListener
+The `DownloadManagerListener` is an implementation of listener for [DownloadManager][downloadmanager] events which can also get notified of
```
yschimke, line 35: Maybe worth calling out it's Media3 DownloadManager and DownloadService.

docs/media-data.md
```diff
@@ -28,3 +28,176 @@ internal database.
...
+override fun onDownloadRemoved(downloadManager: DownloadManager, download: Download) {
+        coroutineScope.launch {
+            val mediaId = download.request.id
+            mediaDownloadLocalDataSource.delete(mediaId)
+        }
+    }
+```
+
+#### DownloadProgressMonitor
+The `DownloadProgressMonitor` monitors the status of the download by polling the `DownloadManager`
```
yschimke, line 67: We should probably highlight that we duplicate the status from Media3's StandaloneDatabaseProvider, but that isn't available for our UI in an easy to query or reactive way.

Unrelated to this PR, we should probably suggest a modern Room implementation of the  StandaloneDatabaseProvider to the media team.

yschimke (approved):

docs/media-data.md
kul3r4: If my syntax is working well, this is linked to the Media 3 resources, see line 202 of this file

yschimke: Yep, the links work, but are only visible after clicking. Your call.

kul3r4: done, thanks

PR #1205, opened 2023-04-17
Improve marquee

yschimke:
#### WHAT

Improve marquee, use the new basicMarquee modifier.

#### WHY

The current implementation of marquee composes the text twice, although so does this one :(
But this should be a more optimal implementation.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


laiyichin (approved):
LGTM

PR #1206, opened 2023-04-17
Bump to AGP 8.0

yschimke:
#### WHAT

Bump to AGP 8.0

#### WHY

Android Studio Flamingo is stable, so AGP 8.0 is now usable without non final AS.

https://android-developers.googleblog.com/2023/04/android-studio-flamingo-is-stable.html

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


sample/src/main/java/com/google/android/horologist/tile/SampleTileRenderer.kt
```diff
@@ -115,8 +115,8 @@ class SampleTileRenderer(context: Context) :
     companion object {
         const val Image1 = "image1"
         const val Icon1 = "icon1"
-        const val TileIcon = R.drawable.ic_android
-        const val TileImage = R.drawable.ic_tileicon
+        val TileIcon = R.drawable.ic_android
```
fstanis, line 118: Why not `const`?

media-ui/src/test/java/com/google/android/horologist/media/ui/components/actions/ShowPlaylistChipA11yTest.kt
```diff
@@ -21,7 +21,7 @@ import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import com.google.android.horologist.compose.tools.coil.FakeImageLoader
-import com.google.android.horologist.media.ui.R
+import com.google.android.horologist.logo.R
```
fstanis, line 24: It feels a bit inconsistent that we import the logo's R in some places and use a fully qualified name in others. I wonder if we can avoid this if we alias the logo resources in the module itself?

fstanis (approved):

sample/src/main/java/com/google/android/horologist/tile/SampleTileRenderer.kt
yschimke: https://android.googlesource.com/platform/tools/base/+/54fe2d86%5E%21/

https://b.corp.google.com/issues/260752253

media-ui/src/test/java/com/google/android/horologist/media/ui/components/actions/ShowPlaylistChipA11yTest.kt
yschimke: Not sure of the right fix. R is a terrible name for something of which there are many.

PR #1207, opened 2023-04-17
Add theme tests for pickers

yschimke:
#### WHAT

Add theme tests for pickers

#### WHY

Validate pickers when themes are used.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


laiyichin (approved):
LGTM

PR #1208, opened 2023-04-17
Fix release builds on AGP 8.0

yschimke:
#### WHAT

Fix release builds on AGP 8.0

#### WHY

AGP 8.0 is stricter and needs these fixes of OkHttp 4.11 (does not exist) or 5.0-alpha

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
@fstanis will follow up with review comments, but landing for now.

PR #1209, opened 2023-04-17
Add the start of a Marquee Benchmark

yschimke:
#### WHAT

Add the start of a Marquee Benchmark

#### WHY

Get some tracing of marquee before and after changes.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
@laiyichin will follow up with review comments, but want to land for now.

PR #1210, opened 2023-04-17
Remove SubcomposeLayout from MarqueeText.

zach-klippenstein:
#### WHAT

Remove SubcomposeLayout from MarqueeText and simplify the code a bunch. I also added some more samples for alignment and RTL.

#### WHY

SubcomposeLayout is expensive and unnecessary for this use case.

#### HOW

The functionality can be replaced with intrinsics.

#### Checklist :clipboard:
- [x] Run spotless check
- [ ] Run tests


yschimke (approved):

yschimke:
Going to land even though Marquee test is hanging in robolectric, will disable that test and raise a bug with compose/robolectric to investigate.

PR #1214, opened 2023-04-18
Fix Marquee

yschimke:
#### WHAT

Fix Marquee Test

#### WHY

Was measuring the item repeatedly.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
Landing to fix the app functionality.

cc @zach-klippenstein

PR #1215, opened 2023-04-18
Avoid automatic navigation on Benchmark tests

yschimke:
#### WHAT

Avoid automatic navigation on Benchmark tests

#### WHY

Want to be able to launch to a particular screen and stay there

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
@laiyichin will address any review comments, but landing to continue testing.

Issue #1200, opened 2023-04-12
Can uses-feature watch be removed from network-awareness library?


mchowning:
It would be nice to be able to add the network-awareness module to modules that are shared by a phone and watch app. Because the network-awareness library's [manifest file specifies that it uses watch features](https://github.com/google/horologist/blob/main/network-awareness/src/main/AndroidManifest.xml#L22), building an app that transitively depends on network-awareness for a phone results in Android Studio presenting a "missing feature: WATCH" pop-up warning.

Does the network-awareness library only work with a watch or could the `uses-feature`-watch line be removed from the network-awareness library's manifest file (which would make it easier to use this library in shared modules)?

mchowning:
If it is possible to use depend on this module within a phone app, would it also make sense to drop the minSdk to from 26 to 23, like [was recently done on some of the other modules that can be used on phones](https://github.com/google/horologist/commit/b6227070ffe0dd30c3f4a2539e267cf29e0b10e8)?

My app's min SDK is currently 23, so I'm not exactly unbiased in making this suggestion 😉. (we are due to increase our min sdk, but I don't think we'll be raising it to 26).

yschimke:
I'll need to split out the UI components.  But it feels like a good time to do that.

yschimke:
Should be fixed in 0.4.3, but wouldn't be surprised if I missed something.

https://github.com/google/horologist/releases

It's now 4 modules to avoid a wear dependency 

network-awareness
network-awareness-db
network-awareness-okhttp
network-awareness-ui (with wear dependencies)

PR #1217, opened 2023-04-19
Split up network-awareness

yschimke:
#### WHAT

Split up network-awareness

#### WHY

https://github.com/google/horologist/issues/1200

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


fstanis:
Three thoughts:

- I'd prefer not to have `network-awareness` under `network-awareness`. Could we rename it to e.g. `network-awareness-core`?
- Similarly, do we need to repeat the parent folder name, i.e. would it work to use `network-awareness/okhttp` instead of `network-awareness/network-awareness-okhttp`?
- I'd argue we should try to keep the top folders one word, i.e. `networkawareness` instead of `network-awareness` to make it clearer what's the "main" part and what's the "subcomponent" (e.g. `okhttp`, `db`).

PR #1218, opened 2023-04-19
Auth Module Refactor

yschimke:
#### WHAT

Auth Module Refactor

#### WHY

Avoid a confusing explosion of top level modules

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


fstanis:
In addition to comments in https://github.com/google/horologist/pull/1217#issuecomment-1514578496, could we also move the three `*sample` folders into a `sample` subfolder, i.e. `auth/auth-sample-phone` -> `auth/sample/phone`?

PR #1219, opened 2023-04-19
Media Module Refactor

yschimke:
#### WHAT

Media Module Refactor.

#### WHY

Avoid a confusing explosion of modules at top level.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


fstanis:
This is the trickiest one...

- What happens with `media-benchmark` and `media-sample-benchmark`? Should they be moved into `media` too?
- What happens with `audio`? While it's not part of media per se, I fear it may cause confusion if it remains in the top level but is very tied to media in general.
- This may be a good opportunity to rename `media3-backend` into `media-backend-media3` which I think is more in line with the rest of the naming convention. Or just `backend-media3` if we go with stripping the parent folder prefix per https://github.com/google/horologist/pull/1217#issuecomment-1514578496.

fstanis (approved):

PR #1220, opened 2023-04-20
Dependency Updates

yschimke:
#### WHAT

Dependency Updates

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/MarqueeTextMediaDisplay.kt
```diff
@@ -24,7 +24,7 @@ import androidx.compose.animation.fadeIn
 import androidx.compose.animation.fadeOut
 import androidx.compose.animation.slideInHorizontally
 import androidx.compose.animation.slideOutHorizontally
-import androidx.compose.animation.with
```
yschimke, line 27: This is deprecated, plus minor cleanup.

PR #1226, opened 2023-04-21
Add smooth transition for progress ring color change

laiyichin:
#### WHAT
Add smooth transition for progress ring color change

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


.idea/kotlinc.xml
yschimke, line 1: probably not needed?

yschimke (approved):

PR #1231, opened 2023-04-25
Version updates and new Pager API in compose

yschimke:
#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media/ui/src/main/java/com/google/android/horologist/media/ui/screens/playerlibrarypager/PlayerLibraryPagerScreen.kt
```diff
@@ -57,6 +57,8 @@ public fun PlayerLibraryPagerScreen(
         if (pageParam != null) {
             try {
                 pagerState.animateScrollToPage(pageParam)
+            } catch (nsee: NoSuchElementException) {
+                // ignore for now
```
luizgrp, line 61: would like to merge with this? what about logging an error to logcat?

luizgrp (approved):

PR #1233, opened 2023-04-26
Align LoadingMediaDisplay positino to TextMediaDisplay

laiyichin:
#### WHAT
Align LoadingMediaDisplay position to TextMediaDisplay and MarqueeMediaDisplay text position

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests
text_alignment

### DIFF
--- a/media/ui/src/main/java/com/google/android/horologist/media/ui/components/display/LoadingMediaDisplay.kt
+++ b/media/ui/src/main/java/com/google/android/horologist/media/ui/components/display/LoadingMediaDisplay.kt
@@ -21,7 +21,6 @@ import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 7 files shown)


media/ui/src/main/java/com/google/android/horologist/media/ui/components/display/LoadingMediaDisplay.kt
```diff
@@ -33,28 +32,30 @@ import androidx.wear.compose.material.MaterialTheme
 import com.google.android.horologist.annotations.ExperimentalHorologistApi
 
 /**
- * A loading state display.
+ * A loading state display. This style is matched to the TextViews of the and [TextMediaDisplay]
```
yschimke, line 35: Should we mention MarqueeTextMediaDisplay as well?

yschimke, line 35:
```suggestion
 * A loading state display. This style is matched to the Text of [TextMediaDisplay] as [MarqueeTextMediaDisplay]
```

yschimke, line 35: If we actually care about this, we should probably consider how we prove this.

Either 

1) The components use the same layout structure, so it's clear they match.  But I don't think this is easy.

2) Maybe a player screen shot with both layouts, with placeholders on top at 50% alpha?

Can be for later.

yschimke (approved):

media/ui/src/main/java/com/google/android/horologist/media/ui/components/display/LoadingMediaDisplay.kt
laiyichin: Great suggestion! Added screenshots overlaying them. Thank you!

PR #1242, opened 2023-04-28
Speed up the playback progress animation

schmiphi:
#### WHAT
Define a new animation spec to be used for the two playback progress animations in Horologist.

#### WHY
The current next-track rotation animation is too slow.

#### HOW
Create a new animation spec, which uses a higher stiffness value (50 before; 100 after)

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
LGTM

PR #1245, opened 2023-05-03
Semantics fix - Remove clearing semantics

laiyichin:
#### WHAT
Remove clearing semantics 

#### WHY
So we can test with onNoteWithText

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/RotaryVolumeControls.kt
yschimke, line 1: accidental?

media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/MarqueeTextMediaDisplay.kt
```diff
@@ -71,7 +72,7 @@ public fun MarqueeTextMediaDisplay(
                 currentTitle ->
             MarqueeText(
                 text = currentTitle.orEmpty(),
-                modifier = Modifier.fillMaxWidth(0.7f).clearAndSetSemantics {
+                modifier = Modifier.fillMaxWidth(0.7f).semantics {
```
yschimke, line 75: I think just remove this, text is probably more correct than contentDescription.

media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/RotaryVolumeControls.kt
laiyichin: My git commit mess.... Should be good now. Only one file changed.

PR #1247, opened 2023-05-04
Update localization.bzl

fstanis:
Follow up to #1219.


yschimke (approved):

yschimke:
@fstanis thanks for fixing after my change broke it

PR #1249, opened 2023-05-04
Add isAvailable API

yschimke:
#### WHAT

Add isAvailable API.

#### WHY

Need to check API availability before using any Wear Data Layer APIs.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


auth/sample/phone/src/main/java/com/google/android/horologist/auth/sample/MainActivity.kt
```diff
@@ -103,7 +103,7 @@ class MainActivity : ComponentActivity() {
                     LaunchedEffect(Unit) {
                         coroutineScope.launch {
                             apiAvailable = tokenBundleRepositoryDefaultKey.isAvailable()
-                            nodeList = phoneDataLayerAppHelper.connectedNodes()
+                            nodeList = if (apiAvailable) phoneDataLayerAppHelper.connectedNodes() else listOf()
```
yschimke, line 106: This was causing the crash.  It didn't actually need public API.

garanj (approved):

PR #1254, opened 2023-05-05
Rotary volume logic for both high res and low res devices

laiyichin:
#### WHAT
Add support for high / low resolution rotary devices to make volume changes.

#### WHY
So that sensitivity for different types of devices can tune volume with their rotary with good user experiences.

#### HOW
- Modify the rotary accumulator to count low res devices positive change as 1 and negative change as -1
- Add modifier for RotaryVolumeControls, and take the changes as is for low res devices and do a 1 pixel = 0.1% volume change for high res devices.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/RotaryInputAccumulator.kt
```diff
@@ -38,16 +39,27 @@ internal class RotaryInputAccumulator(
     public fun onRotaryScroll(scrollPixels: Float, eventTimeMillis: Long) {
         val timeSinceLastAccumulatedMs = eventTimeMillis - lastAccumulatedEventTimeMs
         lastAccumulatedEventTimeMs = eventTimeMillis
+
+        val change = if (isLowRes && scrollPixels > 0f) { // For positive tick in low res devices
```
yschimke, line 43: with these first three branches, and then the different handling in onEventAccumulated, it feels like you might really want to simplfy this into two major cases

if (lowRes) {
 doImmediately
} else {
  current logic
}

yschimke:
@Kpeved want to review?

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/RotaryInputAccumulator.kt
laiyichin: I have considered this, and tested it feeling like if we don't accumulate lowRes the volume change is still quite slow. I can check with UX again on this to verify what is a better experience.

yschimke:
I'll do a fuller review on Monday, but I feel out of my depth.

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/RotaryInputAccumulator.kt
laiyichin: This `onRotaryScroll` method is accumulating the changes when `timeSinceLastAccumulatedMs > eventAccumulationThresholdMs`; for high res devices, it'd be accumulating the pixels,  and for low res devices, it'd be accumulating by notches.

These changes then are passed into the `RotaryVolumeControls`'s `onRotaryInputAccumulated`, high res accumulated pixels are converted into volume (by the 1 pixel == 0.1% volume ratio), and low res notches are converted directly with the ratio of 1 notch : 1 volume change.

media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
```diff
@@ -223,3 +228,8 @@ private fun volumeDescription(volumeUiState: VolumeUiState, isAudioOutputConnect
         stringResource(id = R.string.horologist_volume_screen_not_connected_state)
     }
 }
+
+@ExperimentalHorologistApi
+@Composable
+private fun isLowResInput(): Boolean = LocalContext.current.packageManager
```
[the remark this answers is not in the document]
laiyichin: Yeah I took this from https://github.com/google/horologist/blob/49f98802cbcff8ad2cbab595d6d0c8317ce16b11/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt#L328-L331

Still thinking where could be a good place for this. Suggestion?

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/AccumulatedRotaryInputModifier.kt
```diff
@@ -42,10 +42,11 @@ import com.google.android.horologist.annotations.ExperimentalHorologistApi
 @ExperimentalHorologistApi
 public fun Modifier.onRotaryInputAccumulatedWithFocus(
     focusRequester: FocusRequester? = null,
+    isLowRes: Boolean = false,
```
yschimke, line 45: Can we default this to the right value?

yschimke (approved):

PR #1265, opened 2023-05-11
Ensures watch screen is woken up when remote launching apps

garanj:
Ensures watch screen is woken up when remote launching apps

#### WHAT

Adds screen wakeup when launching apps / activities via AppHelper

#### WHY

When the screen is off, for example on the watch, then launching an app via `startActivity`, does not wake the screen by default.

#### HOW

Use a wake lock with flags to wake the screen, which is released immediately on being obtained, prior to launching the activity.

#### Checklist :clipboard:
- [ x] Run spotless check
- [ x] Run tests

### DIFF
--- a/datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelperService.kt
+++ b/datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelperService.kt
@@ -18,6 +18,7 @@ package com.google.android.horologist.data.apphelper
 
 import android.content.ActivityNotFoundException
 import android.content.Intent
+import android.os.PowerManager
 import android.util.Log
 import com.google.android.gms.tasks.Task
 import com.google.android.gms.tasks.Tasks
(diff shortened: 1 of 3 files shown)


yschimke (approved):

datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelperService.kt
```diff
@@ -88,4 +89,41 @@ public abstract class DataLayerAppHelperService : WearableListenerService() {
...
+
+        // Wakelock timeout should not be required as it is being immediately released but
+        // linting guidance recommends one so setting it nonetheless.
+        wakeLock.acquire(wakeLockTimeoutMs)
+        wakeLock.release()
+    }
+
+    companion object {
+        // Tag format as per recommendations: https://developer.android.com/reference/android/os/PowerManager#newWakeLock(int,%20java.lang.String)
+        private val wakeLockTag = "horologist:apphelper"
```
yschimke, line 126: TIL!

Issue #1106, opened 2023-03-15
Review api vs implementation dependencies


yschimke:
Using the auth code, I found I needed to pull in a bunch of dependencies that should have been brought in by horologist maven artifacts.

I think due to `api` vs `implementation` in gradle, we aren't using api enough.

We should review across the project, potentially using https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin

PR #1266, opened 2023-05-12
Add dependency analysis plugin

luizgrp:
#### WHAT

Add [dependency analysis](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin) plugin to the project.

#### WHY

https://github.com/google/horologist/issues/1106

#### HOW

- Follow project setup from the documentation;

- Run `projectHealth` task for builds in CI;

#### Checklist :clipboard:
- [N/A] Run spotless check
- [N/A] Run tests

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -50,3 +50,8 @@ POM_DEVELOPER_ID=google
 POM_DEVELOPER_NAME=Google
 
 RELEASE_SIGNING_ENABLED=true
+
+# Plugin should be applied individually to modules that are reviewed, in order to avoid having
+# unnecessary performance hit in CI. Once all modules are reviewed, this line can be removed, and
+# the "apply" in each module should also be removed.
+dependency.analysis.autoapply=false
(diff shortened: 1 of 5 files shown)


yschimke (approved):

luizgrp:
Caveat: 
```
The Dependency Analysis plugin is only known to work with versions of AGP between 4.2.2 and 7.4.2. You are using 8.0.1. Proceed at your own risk.
```

gradle.properties
```diff
@@ -50,3 +50,8 @@ POM_DEVELOPER_ID=google
 POM_DEVELOPER_NAME=Google
 
 RELEASE_SIGNING_ENABLED=true
+
+# Plugin should be applied individually to modules that are reviewed, in order to avoid having
+# unnecessary performance hit in CI. Once all modules are reviewed, this line can be removed, and
+# the "apply" in each module should also be removed.
+dependency.analysis.autoapply=false
```
luizgrp, line 57: this is saving ~12mins of CI time, it will decrease as we enable more modules, but worth adding gradually

PR #1267, opened 2023-05-12
Fix dependencies of auth-composables

luizgrp:
#### WHAT

Fix dependencies of `auth-composables` module.

#### WHY

https://github.com/google/horologist/issues/1106

#### HOW

- Run task from [dependency analysis](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin) plugin and address issues.

- Change configuration of the sub project to fail in case of any issues

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth/composables/build.gradle.kts
+++ b/auth/composables/build.gradle.kts
@@ -20,6 +20,7 @@ plugins {
     id("com.android.library")
     id("org.jetbrains.dokka")
     id("me.tylerbwong.gradle.metalava")
+    alias(libs.plugins.dependencyAnalysis)
     kotlin("android")
 }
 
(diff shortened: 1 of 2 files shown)


auth/composables/build.gradle.kts
```diff
@@ -95,29 +95,41 @@ metalava {
 }
 
 dependencies {
-    api(projects.annotations)
```
luizgrp, line 98: This one was marked as unused but it is [used](https://github.com/google/horologist/blob/f5531b434709b21e958635d6279ef853455fb8a1/auth/composables/src/main/java/com/google/android/horologist/auth/composables/screens/SelectAccountScreen.kt#L28):

```
Unused dependencies which should be removed:
  api(project(":annotations"))
```

Maybe the reason is being flagged is because it gets imported from `base-ui` or `compose-layout` intransitively. If that's the case, I'd still think that it should be listed in the dependencies here. I will keep an eye on it as the other modules get fixed.

yschimke (approved):

PR #1274, opened 2023-05-15
Test with emulator 33

yschimke:
#### WHAT

Test with emulator 33

#### WHY

Confirm API 33 emulators are working.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
Failing with the same cause as API 30, but installing fine.

luizgrp (approved):

PR #1276, opened 2023-05-15
Disable macrobenchmark in CI

yschimke:
#### WHAT

Disable macrobenchmark in CI

#### WHY

Used for performance testing, not CI with emulators.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/.github/workflows/android_test.yml
+++ b/.github/workflows/android_test.yml
@@ -7,6 +7,7 @@ on:
     paths-ignore:
       - '**.md'
   workflow_dispatch:
+  pull_request:
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/android_test.yml
```diff
@@ -7,6 +7,8 @@ on:
     paths-ignore:
       - '**.md'
   workflow_dispatch:
+  # TODO remove
+  pull_request:
```
luizgrp, line 11: intend to merge?

yschimke: Nope.

PR #1284, opened 2023-05-16
Reduce frequency of progress animations

yschimke:
#### WHAT

Reduce frequency of progress animations

#### WHY

Want to avoid causing a recomposition on every frame.  

image

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/AnimatedPlayPauseButton.kt
```diff
@@ -216,14 +229,17 @@ public fun AnimatedPlayPauseProgressButton(
                 strokeWidth = progressStrokeWidth
             )
         } else if (trackPositionUiModel.showProgress) {
-            CircularProgressIndicator(
+            val progress = ProgressStateHolder.fromTrackPositionUiModel(trackPositionUiModel)
```
fstanis, line 232: You'll probably want to keep this in the block above. Otherwise, I suspect a loading state won't cancel the previously started coroutine.

yschimke: It will go out of scope, so I think it must.

yschimke: Discussed offline, the LaunchedEffect should be cancelled.

fstanis (approved):
The calculations are a bit hard to read, would be good to add comments in follow up.

PR #1286, opened 2023-05-16
Improve StandardToggleChip to handle largest font scale

luizgrp:
#### WHAT

Improve `StandardToggleChip` to handle largest font scale.

Screenshot 2023-05-16 at 20 23 34

#### WHY

So it won't clip the text when user is using largest font scale.

#### HOW

- Increase the size of the chip when the font scale is bigger than the threshold;
- Add tests and previews with largest font scale;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/base-ui/src/main/java/com/google/android/horologist/base/ui/components/StandardToggleChip.kt
+++ b/base-ui/src/main/java/com/google/android/horologist/base/ui/components/StandardToggleChip.kt
@@ -28,6 +28,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.style.TextAlign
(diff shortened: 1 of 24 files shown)


base-ui/src/main/java/com/google/android/horologist/base/ui/components/StandardToggleChip.kt
```diff
@@ -49,7 +51,7 @@ import com.google.android.horologist.base.ui.util.DECORATIVE_ELEMENT_CONTENT_DES
 @Composable
 public fun StandardToggleChip(
     checked: Boolean,
-    onCheckedChange: (Boolean) -> Unit,
+    onCheckedChanged: (Boolean) -> Unit,
```
yschimke, line 54: Why past tense? Shouldn't we match the ToggleChip?

yschimke (approved):

base-ui/src/main/java/com/google/android/horologist/base/ui/components/StandardToggleChip.kt
luizgrp: My interpretation is that the callback is called after the event has performed. Can revert in the future once figure out that's not the case!

PR #1293, opened 2023-05-19
Fix concurrent writes to WearLocalDataStore

yschimke:
#### WHAT

Fix concurrent writes to WearLocalDataStore

#### WHY

2 issues.

1) The updates are received asynchronously via a listener, so may have a stale value for sequential writes.
2) We need a mutex to make sure we strictly sequence writes in some order. 

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


garanj (approved):
Nice find.

kul3r4 (approved):

PR #1294, opened 2023-05-19
Add a method to mark the app as opened

kul3r4:
#### WHAT
Add a method to mark the app has been opened

#### WHY
Apps can track if there is a friction with an initial screen such as the login

#### HOW
track the first time the app is opened

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
+++ b/datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
@@ -117,7 +117,9 @@ public class WearDataLayerAppHelper(
     public suspend fun markAppHasBeenOpened() {
         surfaceInfoDataStore.updateData { info ->
             info.copy {
-                appHasBeenOpened = true
+                if (!appHasBeenOpened) {
+                    appHasBeenOpened = true
+                }
             }
         }
     }
--- a/datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
+++ b/datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
@@ -112,6 +112,17 @@ public class WearDataLayerAppHelper(
     }
 
     /**
+     * Marks that the app has been opened at least once.
+     */
+    public suspend fun markAppHasBeenOpened() {
+        surfaceInfoDataStore.updateData { info ->
+            info.copy {
+                appHasBeenOpened = true
+            }
+        }
+    }
+
+    /**
      * Marks a tile as removed. Call this in [TileService#onTileRemoveEvent]. Supplying a name is
      * mandatory to disambiguate from the installation or removal of other tiles your app may have.
      *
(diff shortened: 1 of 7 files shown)


datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
```diff
@@ -111,6 +111,19 @@ public class WearDataLayerAppHelper(
         }
     }
 
+    /**
+     * Marks that the app has been opened at least once.
+     */
+    public suspend fun markAppHasBeenOpened() {
```
yschimke, line 117: In terms of naming, what are we trying to convey?

app vs activity?
open vs launch?

I'm ok with it as is, but poking a discussion.

yschimke (approved):
LGTM

datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
kul3r4: Yep launched makes more sense, will fix

datalayer/src/main/proto/app_helper_pb.proto
```diff
@@ -57,4 +57,5 @@ message SurfacesInfo {
   reserved 1;
   repeated ComplicationInfo complications = 2;
   repeated TileInfo tiles = 3;
+  bool appHasBeenOpened = 4;
```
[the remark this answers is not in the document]
kul3r4: done

datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
kul3r4: done

datalayer/src/main/proto/app_helper_pb.proto
```diff
@@ -53,9 +53,15 @@ message TileInfo {
   .google.protobuf.Timestamp timestamp = 3;
 }
 
+message ActivityLaunched {
+  reserved 1;
```
yschimke, line 57: nit: you shouldn't need reserved as part of the initial checkin, really only if we released it during this time.

datalayer-watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
```diff
@@ -112,13 +113,13 @@ public class WearDataLayerAppHelper(
...
+     * Marks that the main activity has been launched at least once.
      */
-    public suspend fun markAppHasBeenOpened() {
+    public suspend fun markActivityLaunchedOnce() {
         surfaceInfoDataStore.updateData { info ->
             info.copy {
-                if (!appHasBeenOpened) {
-                    appHasBeenOpened = true
+                if (!activityLaunched.activityLaunchedOnce) {
+                    activityLaunched = ActivityLaunched.newBuilder().setActivityLaunchedOnce(true).build()
```
yschimke, line 122: Not setting the timestamp here.

yschimke: Should we update timestamp on each launch? Or do we want it to be time of the first launch?

PR #1302, opened 2023-05-22
Add timestamp writing when logging that activity is launched

kul3r4:
#### WHAT
Add writing timestamp when activity is launched

#### WHY
to keep track on when activity was launched

#### HOW
Address feedback from #1294

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/datalayer/src/main/proto/app_helper_pb.proto
+++ b/datalayer/src/main/proto/app_helper_pb.proto
@@ -54,7 +54,6 @@ message TileInfo {
 }
 
 message ActivityLaunched {
-  reserved 1;
   bool activityLaunchedOnce = 2;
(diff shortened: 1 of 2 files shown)


yschimke (approved):

datalayer/src/main/proto/app_helper_pb.proto
```diff
@@ -54,7 +54,6 @@ message TileInfo {
 }
 
 message ActivityLaunched {
-  reserved 1;
```
yschimke, line 57: I'd only remove the reserved value, if you renumber 2 & 3.

kul3r4: done, thank you!

PR #1317, opened 2023-05-26
Avoid rounding the progress to 0 so that the progress does not go away

schmiphi:
#### WHAT
Make sure the drawn progress is at least 0.1% if it was rounded from a non-zero value.

#### WHY
This ensures that the progress bar does not go away if the progress is non-zero bug very small.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/AnimatedPlayPauseButton.kt
+++ b/media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/AnimatedPlayPauseButton.kt
@@ -320,13 +320,14 @@ private fun CircularProgressIndicatorFast(
     }
 }
 
-/** Rounds the progress to avoid producing too many UI updates. */
 private fun roundProgress(progress: Float, progressSteps: Int) = if (progress == 0f) {
     0f
 } else {
-    (floor(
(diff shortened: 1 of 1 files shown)


media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/AnimatedPlayPauseButton.kt
```diff
@@ -286,7 +286,9 @@ private fun CircularProgressIndicatorFast(
             .focusable()
     ) {
         val backgroundSweep = 360f - ((startAngle - endAngle) % 360 + 360) % 360
-        val progressSweep = backgroundSweep * truncatedProgress.coerceIn(0f..1f)
+        // Coerce the progress, avoiding rounding to 0 so that the progress bar does not disappear.
+        val progressSweep =
+            backgroundSweep * (if (progress.value == 0f) 0f else truncatedProgress.coerceIn(0.001f..1f))
```
yschimke, line 291: This has to be read inside the derivedStateOf, or you'll cause us to redraw on every progress change, essentially each frame.

yschimke (approved):

PR #1319, opened 2023-05-26
Add sample of tests checking compose navigation

luizgrp:
#### WHAT

Add sample of tests checking compose navigation.

#### WHY

In order to have it as reference.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Thanks for adding this.

Issue #1314, opened 2023-05-26
Reenable project health after solution for -android modules


yschimke:
See https://github.com/google/horologist/pull/1312#issuecomment-1563948230

yschimke:
****

PR #1321, opened 2023-05-29
Fix dependencies of compose-layout

luizgrp:
#### WHAT

Fix dependencies of `compose-layout` module.

> **Warning**
> Projects that were using `compose-layout` for intransitive dependencies, like `collectAsStateWithLifecycle`, will need to add `androidx.lifecycle:lifecycle-runtime-compose` as dependency.

#### WHY

Follow up on https://github.com/google/horologist/issues/1106

#### HOW

- Run task from [dependency analysis](https://github.com/autonomousapps/dependency-analysis-android-gradle-plugin) plugin and address issues;

- Change configuration of the sub project to fail in case of any issues;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/compose-layout/build.gradle.kts
+++ b/compose-layout/build.gradle.kts
@@ -21,6 +21,7 @@ plugins {
     id("org.jetbrains.dokka")
     id("org.jetbrains.kotlin.kapt")
     id("me.tylerbwong.gradle.metalava")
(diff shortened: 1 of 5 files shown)


luizgrp:
Closing as it is risky while https://github.com/google/horologist/issues/1314 is still an issue. Will re-open once it is sorted.

PR #1337, opened 2023-05-31
Archive snapshot delta images

luizgrp:
#### WHAT

Archive snapshot diff output folder.

#### WHY

So we can download and see the differences.

Last attempt did not work: https://github.com/google/horologist/pull/1334

#### Checklist :clipboard:
- [N/A] Run spotless check
- [N/A] Run tests


luizgrp:
previous https://github.com/google/horologist/pull/1334 did work

PR #1350, opened 2023-06-01
Fix ToggleChip for RTL

luizgrp:
#### WHAT

Fix `ToggleChip` and `SplitToggleChip` for RTL.

#### HOW

- Replace `TextAlign` from `Left` to `Start`;
- Use `compose-material`'s `Icon` with `IconRtlMode.Mirrored` for the toggle control;
- Add `iconRtlMode` param to `ToggleChip` to allow icon to also be mirroed;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/compose-material/api/current.api
+++ b/compose-material/api/current.api
@@ -22,7 +22,7 @@ package com.google.android.horologist.compose.material {
   }
 
   public final class ToggleChipKt {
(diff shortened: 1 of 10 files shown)


yschimke (approved):
I'll take your word for it on the mirror options. But with tests in place it's deliberate even if we change in future.

Issue #1324, opened 2023-05-29
Add Compose Material library


luizgrp:
### WHAT

Add Horologist Compose Material library.

### WHY

In order to provide components built on the core AndroidX library, to remove boilerplate and to add opinionated a11y support.

### HOW
- https://github.com/google/horologist/pull/1327;
- [x] Copy components from `base-ui` to `compose-material` module, changing the package name;
   - [x] Change component names to match the ones in wear-compose;
   - [x] Keep components in `base-ui` for few releases, marked as deprecated, suggesting the new library;
   - [x] Change the code of the components in `base-ui` to use the ones from `compose-material`;
   - [x] Remove previews from `base-ui`, keep the unit tests to check if the usage of `compose-material` is correct;
- [ ] Add new components TBD;
   - https://github.com/google/horologist/issues/1318
   - https://github.com/google/horologist/issues/1346
   - https://github.com/google/horologist/issues/1396
   - https://github.com/google/horologist/issues/1397
   - https://github.com/google/horologist/issues/1398
   - https://github.com/google/horologist/issues/1411
- https://github.com/google/horologist/issues/1347;
- https://github.com/google/horologist/pull/1392;
- https://github.com/google/horologist/pull/1492;

yschimke:
Closing as completed in favour of individual issues.

PR #1353, opened 2023-06-02
Add rememberVectorPainter to compose-material

luizgrp:
#### WHAT

Add `rememberVectorPainter` to `compose-material`.

#### WHY

https://github.com/google/horologist/issues/1324

#### HOW

- Keep components in `base-ui` for few releases, marked as deprecated, suggesting the new library;
- Change the code of the components in `base-ui` and samples to use the ones from `compose-material`;
- Remove previews from `base-ui`, keep the unit tests to check if the usage of `compose-material` is correct;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/base-ui/src/main/java/com/google/android/horologist/base/ui/components/StandardChip.kt
+++ b/base-ui/src/main/java/com/google/android/horologist/base/ui/components/StandardChip.kt
@@ -44,8 +44,8 @@ import androidx.wear.compose.material.MaterialTheme
 import androidx.wear.compose.material.Text
 import coil.compose.rememberAsyncImagePainter
 import com.google.android.horologist.annotations.ExperimentalHorologistApi
-import com.google.android.horologist.base.ui.util.adjustChipHeightToFontScale
 import com.google.android.horologist.compose.material.util.DECORATIVE_ELEMENT_CONTENT_DESCRIPTION
+import com.google.android.horologist.compose.material.util.adjustChipHeightToFontScale
 
 /**
  * This composable fulfils the redlines of the following components:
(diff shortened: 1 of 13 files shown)


base-ui/src/test/java/com/google/android/horologist/base/ui/components/StandardChipPrimaryTest.kt
```diff
@@ -234,8 +234,9 @@ class StandardChipPrimaryTest : ScreenshotBaseTest() {
                     secondaryLabel = "Secondary label",
                     icon = "iconUri",
                     placeholder = rememberVectorPainter(
-                        image = Icons.Default.Image,
-                        tintColor = Color.Black
+                        Icons.Default.Image,
+                        Color.Black,
+                        Icons.Default.Image.tintBlendMode
```
luizgrp, line 239: remove this param, bring back the param names

base-ui/src/main/java/com/google/android/horologist/base/ui/components/StandardChip.kt
```diff
@@ -186,7 +186,7 @@ public fun StandardChip(
         label = labelParam,
         onClick = onClick,
         modifier = modifier
-            .adjustChipHeightToFontScale(LocalConfiguration.current.fontScale)
+            .adjustChipHeightToFontScale(LocalConfiguration.current.fontScale, 0.dp)
```
luizgrp, line 189: remove the default value

yschimke (approved):

PR #1354, opened 2023-06-02
Add Chip to compose-material

luizgrp:
#### WHAT

Add `Chip` to `compose-material`.

#### WHY

https://github.com/google/horologist/issues/1324

#### HOW

- Keep components in `base-ui` for few releases, marked as deprecated, suggesting the new library;
- Change the code of the components in `base-ui` and samples to use the ones from `compose-material`;
- Remove previews from `base-ui`, keep the unit tests to check if the usage of `compose-material` is correct;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Mostly skimmed

Issue #1346, opened 2023-06-01
Improve Title component


luizgrp:
Improve `Title` component to have the variants listed in the [Material Design kit](https://developer.android.com/design/ui/wear/guides/foundations/download).

<details><summary>Redlines</summary>
<p>

![Screenshot 2023-06-01 at 10 26 56](https://github.com/google/horologist/assets/878134/9385ff99-9a91-4dda-8edf-ddf495da21eb)

![Screenshot 2023-06-01 at 10 31 57](https://github.com/google/horologist/assets/878134/e70c7867-0c89-4559-a483-7a12e05fe4b2)

</p>
</details> 

Expected output:
- Implementation in `compose-material` module, matching the redlines;
- Previews in the `debug` folder for each variant;
- Snapshot tests for:
   - each variant;
   - text overflow;

DenoRS:
@luizgrp happy to pick this up :)

Issue #1335, opened 2023-05-30
Improve ScreenshotBaseTest


luizgrp:
### WHAT

Improve `ScreenshotBaseTest` to allow it to set the parameters in a more convenient way, without requiring to call and import `screenshotTestRuleParams`. For example:

```kotlin
class MyTest : ScreenshotBaseTest({
    screenTimeText = {}
    enableA11y = true
}) {
```

### WHY

Feedback on usage https://github.com/google/horologist/pull/1312#discussion_r1205741849

### HOW

_Suggestion_ of implementation with [type safe builders](https://kotlinlang.org/docs/type-safe-builders.html).

yschimke:
We should probably also pick up the record default value from a system property so we can run tests in a record mode, and not edit the files.

PR #1368, opened 2023-06-07
Improve Title Component

DenoRS:
#### WHAT

- Improve Title component

#### WHY

- resolves https://github.com/google/horologist/issues/1346

#### HOW

- Improved Title component based on redlines.
- Added previews to debug folder
- Added snapshot tests

#### Screenshots

**Primary Title preview**:

image

**Secondary Title preview**:

image

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/compose-material/src/main/java/com/google/android/horologist/compose/material/Title.kt
+++ b/compose-material/src/main/java/com/google/android/horologist/compose/material/Title.kt
@@ -17,16 +17,30 @@
 package com.google.android.horologist.compose.material
 
 import androidx.annotation.StringRes
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.heading
 import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
 import androidx.wear.compose.material.MaterialTheme
 import androidx.wear.compose.material.Text
 import com.google.android.horologist.annotations.ExperimentalHorologistApi
+import com.google.android.horologist.compose.material.util.DECORATIVE_ELEMENT_CONTENT_DESCRIPTION
 
 /**
  * A title heading to group and identify items.
@@ -35,11 +49,17 @@ import com.google.android.horologist.annotations.ExperimentalHorologistApi
 @Composable
 public fun Title(
     @StringRes textId: Int,
-    modifier: Modifier = Modifier
+    modifier: Modifier = Modifier,
+    textType: TextType = TextType.Primary,
+    icon: ImageVector? = null,
+    iconSize: Dp = 24.dp
 ) {
     Title(
         text = stringResource(id = textId),
-        modifier = modifier
+        modifier = modifier,
+        textType = textType,
+        icon = icon,
+        iconSize = iconSize
     )
 }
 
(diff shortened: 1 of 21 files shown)


compose-material/src/main/java/com/google/android/horologist/compose/material/Title.kt
```diff
@@ -50,14 +70,53 @@ public fun Title(
...
+            .padding(horizontal = 14.dp),
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        icon?.let {
+            Icon(
+                imageVector = icon,
+                contentDescription = DECORATIVE_ELEMENT_CONTENT_DESCRIPTION,
+                modifier = Modifier
+                    .size(iconSize)
+                    .clip(CircleShape),
```
luizgrp, line 90: It doesn't need to be clipped to circle shape according to spec

compose-material/src/main/java/com/google/android/horologist/compose/material/Title.kt
```diff
@@ -50,14 +70,53 @@ public fun Title(
...
-        overflow = TextOverflow.Ellipsis,
-        maxLines = 3,
-        style = MaterialTheme.typography.title3
-    )
+    Row(
+        modifier = Modifier
+            .padding(horizontal = 14.dp),
+        verticalAlignment = Alignment.CenterVertically
+    ) {
+        icon?.let {
```
luizgrp, line 84: Icon should only be displayed for secondary title, so it should add a check here to ignore the param or throw an exception to fail fast and let the developer know the proper usage. 

That might be a confusing API, so I'm leaning towards having a separate `Title` function for the secondary title. But if we use the same name, the signatures might be the same and it will be difficult to use the secondary title:

```kotlin
fun Title(text: String, modifier: Modifier, ...) {} // primary
fun Title(text: String, modifier: Modifier, icon: ImageVector? = null, ...) {} // secondary

// this might call the primary title:
Title(text = "I want a secondary title with no icon")
```

So maybe this extra function should be called `SecondaryTitle`. Thoughts?

compose-material/src/main/java/com/google/android/horologist/compose/material/Title.kt
```diff
@@ -50,14 +70,53 @@ public fun Title(
...
+            style = when (textType) {
+                TextType.Primary -> MaterialTheme.typography.button
+                TextType.Secondary -> MaterialTheme.typography.caption1
+            }
+        )
+    }
+}
+
+@ExperimentalHorologistApi
+public enum class TextType {
```
luizgrp, line 119: nit: rename it to **Title**Type if we keep it?

compose-material/src/main/java/com/google/android/horologist/compose/material/Title.kt
```diff
@@ -50,14 +70,53 @@ public fun Title(
...
+            modifier = modifier
+                .semantics { heading() }
+                .fillMaxWidth(),
+            color = when (textType) {
+                TextType.Primary -> MaterialTheme.colors.onSurfaceVariant
+                TextType.Secondary -> MaterialTheme.colors.onBackground
+            },
+            textAlign = when (textType) {
+                TextType.Primary -> TextAlign.Center
+                TextType.Secondary -> TextAlign.Left
```
luizgrp, line 106: Should use TextAlign.Start instead of Left in order to work properly with RTL locales.
Could you please also add a test checking RTL layout? Similar to [this](https://github.com/google/horologist/blob/d94130ee2a991bd785c6764a9812843c05c82d75/compose-material/src/test/java/com/google/android/horologist/compose/material/ToggleChipTest.kt#L303)

compose-material/src/test/java/com/google/android/horologist/compose/material/TitleTest.kt
```diff
@@ -16,22 +16,76 @@
...
 
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.outlined.MusicNote
+import androidx.compose.ui.graphics.Color
 import com.google.android.horologist.screenshots.ScreenshotBaseTest
+import com.google.android.horologist.screenshots.ScreenshotTestRule
 import org.junit.Test
 
-class TitleTest : ScreenshotBaseTest() {
+class TitleTest : ScreenshotBaseTest(ScreenshotTestRule.screenshotTestRuleParams { record = true }) {
```
luizgrp, line 26: remove the `screenshotTestRuleParams { record = true }`

luizgrp (commented):
Thanks! Looking good, left a comment to consider regarding the API.

Apologies for CI failing due to the images, will regenerate them once it's merged.

compose-material/src/main/java/com/google/android/horologist/compose/material/Title.kt
DenoRS: Thanks, that's a good observation.
Agreed and added the 'SecondaryTitle' function for simplicity :)

compose-material/src/main/java/com/google/android/horologist/compose/material/Title.kt
DenoRS: Thanks, removed as no longer used.

compose-material/src/main/java/com/google/android/horologist/compose/material/Title.kt
DenoRS: Ahh yes, I wasn't sure about adding rtl. Added now :)

compose-material/src/test/java/com/google/android/horologist/compose/material/TitleTest.kt
DenoRS: 🤦🏼‍♂️ removed :)

luizgrp: thanks! we should really have a better solution for this as suggested in https://github.com/google/horologist/issues/1335#issuecomment-1582302637

compose-material/src/main/java/com/google/android/horologist/compose/material/Title.kt
```diff
@@ -44,20 +56,93 @@ public fun Title(
...
+        text = stringResource(id = textId),
+        modifier = modifier,
+        icon = icon,
+        iconSize = iconSize,
+        iconRtlMode = iconRtlMode
     )
 }
+
+/**
+ * A primary title heading to group and identify items with optional icon.
```
luizgrp, line 108: nit: secondary

luizgrp (commented):
LGTM - Thank you!

PR #1369, opened 2023-06-07
Adds timeout handling to MessageClient use in App Helper

garanj:
#### WHAT

Adds timeout handling to use of MessageClient in App Helper

#### WHY

When using the App Helper to launch app, activity or companion on the remote device, Horologist should handle timeouts.

#### HOW

Introduce a new value for `AppHelperResultCode` for timeouts. Handle both timeouts from GMS itself as well as limiting the `MessageClient` requests to 15s with `withTimeout { } `.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
```diff
@@ -141,18 +145,36 @@ abstract class DataLayerAppHelper(
...
+    }
+
+    /**
+     * Attempts a [MessageClient#sendRequest] with a timeout, covering both the GMS Timeout that may
+     * occur (after 1 minute?) but also a timeout specified by Horologist, as 1 minute is often too
+     * long.
+     */
+    protected suspend fun sendRequestWithTimeout(node: String, path: String, data: ByteArray): AppHelperResultCode {
+        val response = try {
+            withTimeout(MESSAGE_REQUEST_TIMEOUT_MS) {
```
yschimke, line 166: Maybe worth a comment as @fstanis pointed out, cancellation doesn't cancel the GM's task.

yschimke, line 166: Consider making the timeout a param with this default value.

datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
```diff
@@ -141,18 +145,36 @@ abstract class DataLayerAppHelper(
...
+                registry.messageClient.sendRequest(node, path, data).await()
+            }
+        } catch (timeoutException: TimeoutCancellationException) {
+            return AppHelperResultCode.APP_HELPER_RESULT_TIMEOUT
+        } catch (e: ApiException) {
+            if (e.statusCode == CommonStatusCodes.TIMEOUT) {
+                return AppHelperResultCode.APP_HELPER_RESULT_TIMEOUT
```
yschimke, line 173: This worries me that it will be silently ignored. Should we either throw a well known timeout exception here, or enforce checking result?

yschimke: Or turn other errors into a 7th value?

yschimke (approved):

datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
garanj: Yeah, I agree with the issue of being silently ignored. Maybe: (1) throw the well-known timeout exception, (2) don't catch ApiException at all, so the developer needs to handle the timeout exception?

datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
garanj: Oh interesting. I just looked around just now, there's no way to cancel it right?

datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
yschimke: From your own use, which ones would you prefer to deal with?

datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
garanj: Done

garanj: Added comment.

datalayer/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
garanj: Reading a bit more about CheckResult I've gone that route for now.

PR #1371, opened 2023-06-08
Fix Chip for RTL locale

luizgrp:
#### WHAT

Fix `Chip` for RTL locale.

#### HOW

- Replace `TextAlign` from `Left` to `Start`;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


compose-material/src/test/snapshots/images/com.google.android.horologist.compose.material_ChipPrimaryTest_withSecondaryLabelAndIconRtl.png
yschimke, line 1: Nerd sniping you, I wonder if it's worth testing these with the pseudo locale? https://developer.android.com/guide/topics/resources/pseudolocales

compose-material/src/test/java/com/google/android/horologist/compose/material/ChipPrimaryTest.kt
```diff
@@ -243,6 +244,20 @@ class ChipPrimaryTest : ScreenshotBaseTest() {
         }
     }
 
+    @Test
+    fun withSecondaryLabelAndIconRtl() {
+        screenshotTestRule.setContent(isComponent = true, takeScreenshot = true) {
+            TestHarness(layoutDirection = LayoutDirection.Rtl) {
```
yschimke, line 250: should we move this as an optional param on setContent?

yschimke (approved):

compose-material/src/test/snapshots/images/com.google.android.horologist.compose.material_ChipPrimaryTest_withSecondaryLabelAndIconRtl.png
luizgrp: Interesting! It looks like the pseudolocales is enabled in a build type, which is not great as we don't have a separate build type just for tests, so if we enable it, it would affect our debug builds. I will raise a ticket to explore it.

compose-material/src/test/java/com/google/android/horologist/compose/material/ChipPrimaryTest.kt
luizgrp: that's a good idea!

compose-material/src/test/snapshots/images/com.google.android.horologist.compose.material_ChipPrimaryTest_withSecondaryLabelAndIconRtl.png
yschimke: I guess two reasons here

1) english in RTL looks weird
2) in case we have some content that is not translated

PR #1375, opened 2023-06-08
Change Chips to rely on default colors provided by compose material

luizgrp:
#### WHAT

Change Chip components of `compose-material` to rely on default colors provided by wear compose material.

#### WHY

- ~They should be redundant~ Compose material has a more up to date color definition, taking contrast ratio into account. 
- Our current code to set the colors is impacting the disabled states.

#### HOW

Remove code that sets the color of Text and Icon components in the Chip components.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


compose-material/src/test/snapshots/images/com.google.android.horologist.compose.material_ChipSecondaryTest_withSecondaryLabel.png
luizgrp, line 1: it looks like the secondary label has the same color of the primary label, which shouldn't, will raise a bug with wear compose material to check

luizgrp: confirmed with @BowerSteve that is intended, for contrast ratio reasons

yschimke (approved):

PR #1379, opened 2023-06-08
Use the Tile's state in the TilePreview

zlandorf:
#### WHAT
Use the Tile's state in the TilePreview

#### WHY
There was a recent change to support state changes in the public API of the TileRenderer: [aosp/2607086](https://android-review.git.corp.google.com/c/platform/frameworks/support/+/2607086).

Tile previews should be as close to what would be run on the device as possible.

#### HOW
By using the updated TileRenderer API.

Checklist 📋
* [N/A] Add explicit visibility modifier and explicit return types for public declarations
* [x] Run spotless check
* [x] Run tests
* [N/A] Update metalava's signature text files

### DIFF
--- a/compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
+++ b/compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
@@ -42,6 +42,7 @@ import androidx.wear.protolayout.StateBuilders.State
 import androidx.wear.protolayout.TimelineBuilders
 import androidx.wear.tiles.RequestBuilders
 import androidx.wear.tiles.TileBuilders
+import androidx.wear.tiles.renderer.TileRenderer
 import com.google.android.horologist.annotations.ExperimentalHorologistApi
 import com.google.android.horologist.tiles.render.TileLayoutRenderer
 import kotlinx.coroutines.Dispatchers
@@ -85,11 +86,15 @@ public fun TilePreview(
             }
         },
         update = {
-            val tileRenderer = androidx.wear.tiles.renderer.TileRenderer(
+            val tileRenderer = TileRenderer(
                 /* uiContext = */ it.context,
                 /* loadActionExecutor = */ Dispatchers.IO.asExecutor(),
                 /* loadActionListener = */ {}
             )
+
+            tile.state?.let { state ->
+                tileRenderer.setState(state.keyToValueMapping)
+            }
 
             // Returning a future
             tileRenderer.inflateAsync(


compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -85,12 +86,19 @@ public fun TilePreview(
             }
         },
         update = {
-            val tileRenderer = androidx.wear.tiles.renderer.TileRenderer(
+            var tileRenderer: TileRenderer? = null
```
zlandorf, line 89: I'm declaring the variable here so that I can then call it in the `loadActionListener` lambda below. I was not able to use `this` or didn't find an elegant way of doing that. Please let me know if you know of a better way of doing that and I'd be happy to update it.

Not sure if there is a better way than doing it like this.

yschimke: No better suggestion.  Should we make this a `lateinit`?

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -85,12 +86,19 @@ public fun TilePreview(
...
+            tileRenderer = TileRenderer(
                 /* uiContext = */ it.context,
                 /* loadActionExecutor = */ Dispatchers.IO.asExecutor(),
```
yschimke, line 92: is this is now called do we want to to be synchronous in previews?

`{ runnable -> runnable.run() }`

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
yschimke: as far as calling newState, I guess I don't really understand why it's exposed to the caller? Is it expected that callers change the state in some way?

yschimke:
Run `./gradlew spotlessApply` to fix the formatting issues.

yschimke (approved):

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
zlandorf: My understanding is that this is just the public TileRenderer API which is meant to be used for previews and such. The real TileRenderer in SysUI is different. So I guess it's expected in case something needs to be done at the preview level. 

But my understanding might be wrong, I'll ping msab to see if I am using this in the wrong way.

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
zlandorf: Sorry, I'm not quite sure I follow the suggestion. Could you please clarify?

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
castedmo: Sorry about the confusing parameter name there. This parameter actually doesn't contain any key/value state items. It only contains the id of the clickable that has triggered this loadAction (https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/protolayout/protolayout/src/main/java/androidx/wear/protolayout/StateBuilders.java;l=77;drc=be147a1bdead346eaee13e69e025b0d854a608a4;bpv=0;bpt=0).

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -85,12 +86,19 @@ public fun TilePreview(
...
         },
         update = {
-            val tileRenderer = androidx.wear.tiles.renderer.TileRenderer(
+            lateinit var tileRenderer: TileRenderer
+            tileRenderer = TileRenderer(
                 /* uiContext = */ it.context,
                 /* loadActionExecutor = */ Dispatchers.IO.asExecutor(),
-                /* loadActionListener = */ {}
+                /* loadActionListener = */ { newState ->
+                tileRenderer.setState(newState.keyToValueMapping)
```
castedmo, line 94: This isn't correct. Inside this callback you have access to the clickable Id that has triggered this reload. newState is empty except for `getLastClickableId`.
This information is for the Tile service to figure out how the new layout should look like. Passing it again to the renderer is wrong.

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
yschimke: What thread do you want the loadActionExecutor to run jobs on? Should it be IO? Or if we start using click handling, do we want this to be kept on the main thread so that interactive preview or robolectric tests run predictably?

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
yschimke: I think for Previews and Tests, it might be good to allow callers to respond to the event with some new State?

So maybe a param on TilePreview `onClick: (lastClickableId: String, State) -> State = {it}` ?

yschimke: This can be a follow up, after more discussion, and empty in this PR.

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
zlandorf: Thanks for the clarification. You're absolutely right and thanks for catching that. It should be the main thread for handling click events.

As I've made the `loadActionListener` empty again, I'll leave it as is, but will keep it in mind for any future changes.

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -85,12 +86,17 @@ public fun TilePreview(
             }
         },
         update = {
-            val tileRenderer = androidx.wear.tiles.renderer.TileRenderer(
+            lateinit var tileRenderer: TileRenderer
```
yschimke, line 89: Can you inline this again and remove lateinit?

zlandorf: Ah yes, of course. Thanks for catching that and sorry for the brain lags 😅

PR #1381, opened 2023-06-09
Change Chip API to allow custom colors and background

luizgrp:
#### WHAT

Change Chip API to allow custom colors and background.

#### WHY

In order to allow fulfil the redlines for Chip component from [Wear Material Design Kit](https://developer.android.com/design/ui/wear/guides/foundations/download).

#### HOW

Remove `ChipType` class and param and add `color` param of `ChipColors` type.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
much nicer.

PR #1382, opened 2023-06-09
[on hold] Publish node info

yschimke:
#### WHAT

Publish node info

#### WHY

Build a nice overall set of device info

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


datalayer/src/main/proto/app_helper_pb.proto
```diff
@@ -25,9 +25,21 @@ message LaunchRequest {
...
 }
 
+message NodeInfo {
+  DeviceInfo device = 1;
+}
+
+message DeviceInfo {
+  uint32 api_version = 1;
+  string device = 2;
+  string model = 3;
```
garanj, line 39: Maybe also `MANUFACTURER` - I don't know how well these are populated? `SKU` might possibly be useful too?

datalayer/src/main/proto/app_helper_pb.proto
```diff
@@ -26,9 +26,31 @@ message LaunchRequest {
...
+  AppInfo app = 2;
+}
+
+message DeviceInfo {
+  uint32 api_version = 1;
+  string device = 2;
+  string model = 3;
+  string product = 4;
+  string manufacturer = 5;
+  repeated string feature = 6;
```
yschimke, line 44: This might be overkill.  there are 95 of these.

PR #1386, opened 2023-06-12
Change rateLimitCoolDown for rotary accumulator for TimePicker.

Kpeved:
This number (5ms) proofed to be better than default ( 30ms)

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke (approved):

Kpeved:
> Happy to land on trust, but generally it would be good to have eitehr
> 
> 1. video showing manual testing why it's better
>    or
> 2. Some test that changes, that shows the output is better.
> 
> I'm going to try to make a test for 2.

This is very subjective change, Ux team just thought that 5ms looks better :)

PR #1387, opened 2023-06-12
Don't write when data is unchanged

yschimke:
#### WHAT

Don't write when data is unchanged.

#### WHY

The pattern of DataLayer helper includes writing just in case, when data may not have changed. We should de-dupe these updates to minimise extra work.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
makes sense

PR #1389, opened 2023-06-12
Add haptics and reverse to ScalingLazyColumnState

yschimke:
#### WHAT

Add haptics and reverse to ScalingLazyColumn

#### WHY

Missing API and functionality.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
@@ -38,6 +38,8 @@ import androidx.wear.compose.foundation.lazy.ScalingParams
 import androidx.wear.compose.foundation.rememberActiveFocusRequester
 import com.google.android.horologist.annotations.ExperimentalHorologistApi
 import com.google.android.horologist.compose.layout.ScalingLazyColumnState.RotaryMode
+import com.google.android.horologist.compose.rotaryinput.rememberDisabledHaptic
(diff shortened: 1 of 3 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
```diff
@@ -67,7 +69,8 @@ public class ScalingLazyColumnState(
     public val horizontalAlignment: Alignment.Horizontal = Alignment.CenterHorizontally,
     public val flingBehavior: FlingBehavior? = null,
     public val userScrollEnabled: Boolean = true,
-    public val scalingParams: ScalingParams = WearScalingLazyColumnDefaults.scalingParams()
+    public val scalingParams: ScalingParams = WearScalingLazyColumnDefaults.scalingParams(),
+    public val hapticsEnabled: Boolean = true
```
luizgrp, line 73: nit: in `scalingLazyColumnDefaults` the params anchorType, hapticsEnabled and reverseLayout are close to each other, here they are separated - should the signature of these be re-reviewed?

luizgrp (approved):

compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
yschimke: Possibly, but the existing ones are not consistent.  I don't actually expect any one to call them in ordered form, so would suggest this is low priority.  arguments don't match 1:1.

PR #1399, opened 2023-06-15
Improve A11y for ToggleChip and SplitToggleChip

luizgrp:
#### WHAT

Improve A11y for `ToggleChip` and `SplitToggleChip`.

#### HOW

[Describe element state](https://developer.android.com/jetpack/compose/accessibility#anchor) with `stateDescription`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/compose-material/src/test/snapshots/images/com.google.android.horologist.compose.material_SplitToggleChipA11yTest_disabled.png
+++ b/compose-material/src/test/snapshots/images/com.google.android.horologist.compose.material_SplitToggleChipA11yTest_disabled.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:ef6180be59d7af0a223f52b4687313b85ea42e07c2ba1952bf4fa3ead110dfc5
-size 21138
+oid sha256:3fbb6a880fa82e7599979cb90b228fe22487dae74cd97fe616a212d60a01c930
+size 21143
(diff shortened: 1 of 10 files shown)


compose-material/src/test/snapshots/images/com.google.android.horologist.compose.material_SplitToggleChipA11yTest_disabled.png
yschimke, line 1: Should we be merging these? Or it's deliberate because of the "Split" nature?

Maybe state description needs to describe what is on?

yschimke (approved):

compose-material/src/test/snapshots/images/com.google.android.horologist.compose.material_SplitToggleChipA11yTest_disabled.png
luizgrp: My understanding is that it is deliberated, this is from the kdocs:

> The SplitToggleChip differs from the ToggleChip by having two "tappable" areas, one clickable and one toggleable.

So, tapping on the content area can have a different action than tapping on the toggle control.

yschimke: ok, so maybe it's just about explaining what is on/off.

PR #1403, opened 2023-06-16
Improve CreateAccountChip

luizgrp:
#### WHAT

Improve `CreateAccountChip`.

#### WHY

- Icon from icon library does look like what's seen on the [website](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:send_to_mobile:FILL@0;wght@400;GRAD@0;opsz@48&icon.query=send+to+mobile);
- Requirement to have small icon from this chip aligned with chips with large avatar - we should consider having this param in `Chip` itself;

#### HOW

- Add svg icon from website to the auth library itself;
- Add `largeIconSpace` param to place the icon in a box with the size of the large icon;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


auth/composables/src/main/java/com/google/android/horologist/auth/composables/chips/CreateAccountChip.kt
```diff
@@ -38,15 +42,39 @@ public fun CreateAccountChip(
...
 ) {
-    Chip(
-        label = label,
-        onClick = onClick,
-        modifier = modifier,
-        icon = Icons.Default.SendToMobile,
-        colors = colors,
-        enabled = enabled
-    )
+    if (largeIconSpace) {
```
yschimke, line 49: personal style: can the if be just on the icon param?

yschimke (approved):

auth/composables/src/main/java/com/google/android/horologist/auth/composables/chips/CreateAccountChip.kt
luizgrp: Not in this case, as it uses different Chip functions depending on the flag (one with signature `icon:Any?` and the other with signature `icon:(@Composable BoxScope.() -> Unit)`.  If the if statement is added just on the icon param, it thinks both cases should use the function with signature `icon:Any?` and eventually breaks when the composable value is passed. If you know a way, please let me know.

luizgrp:
cc @LorenzNickel

PR #1404, opened 2023-06-16
Fix end content padding of Chip

luizgrp:
#### WHAT

Fix end content padding of `Chip`.

#### WHY

In order to align with specs.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


compose-material/src/main/java/com/google/android/horologist/compose/material/Chip.kt
```diff
@@ -200,9 +200,14 @@ public fun Chip(
         }
 
     val contentPadding = if (largeIcon) {
-        val horizontalPadding = 10.dp
-        val verticalPadding = 6.dp // same as Chip.ChipVerticalPadding
-        PaddingValues(horizontal = horizontalPadding, vertical = verticalPadding)
+        val horizontalPadding = 14.dp // same as ChipDefaults.ChipHorizontalPadding
```
yschimke, line 203: Should these be public?

yschimke: Also, may as well inline if it's only the end padding?

yschimke (approved):

compose-material/src/main/java/com/google/android/horologist/compose/material/Chip.kt
luizgrp: I can raise a feature request to ask them to be public.

What would be the inline version?

luizgrp: b/287599459

compose-material/src/main/java/com/google/android/horologist/compose/material/Chip.kt
```diff
@@ -200,9 +200,14 @@ public fun Chip(
...
+        val verticalPadding = 6.dp // same as ChipDefaults.ChipVerticalPadding
+        PaddingValues(
+            start = 10.dp,
+            top = verticalPadding,
+            end = horizontalPadding,
+            bottom = verticalPadding
```
yschimke, line 209:
```suggestion
        val verticalPadding = 6.dp // same as ChipDefaults.ChipVerticalPadding
        PaddingValues(
            start = 10.dp,
            top = verticalPadding,
            end = 14.dp, // same as ChipDefaults.ChipHorizontalPadding
            bottom = verticalPadding
```

PR #1419, opened 2023-06-16
Avoid clipping when not paging

yschimke:
#### WHAT

Avoid clipping when not paging.

#### WHY

skia dump suggested that layers with clipping were expensive.  We only need the clipping while scrolling between pages.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
@BowerSteve I haven't benchmarked this yet, will try to find some time to test with and without.

luizgrp (approved):

PR #1420, opened 2023-06-16
Change TilePreview API to allow providing an onLoadAction callback

zlandorf:
This callback is used to provide a new Tile whenever there is a change of state

#### WHAT
Add a callback parameter to `TilePreview` that allows the user to provide an updated Tile representation in order to update the tile's preview.

#### WHY
At the moment, tile previews are not interactive as LoadActions are ignored. The addition of this parameter will allow a way for a user to test interactivity of their tiles.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests
- n/a Update metalava's signature text files

### DIFF
--- a/compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
+++ b/compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
@@ -19,6 +19,8 @@ package com.google.android.horologist.compose.tools
 import android.content.res.Resources
 import android.graphics.Color
 import android.util.DisplayMetrics
+import android.view.View
+import android.view.ViewGroup
 import android.widget.FrameLayout
 import androidx.annotation.ColorInt
 import androidx.compose.foundation.layout.fillMaxSize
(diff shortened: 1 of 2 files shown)


compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
```diff
@@ -85,27 +92,39 @@ public fun TilePreview(
...
+private fun TileRenderer.preview(
+    tile: TileBuilders.Tile,
+    tileResources: ResourceBuilders.Resources,
+    parent: ViewGroup
+): ListenableFuture<View> {
+    tile.state?.let { state ->
+        setState(state.keyToValueMapping)
+    }
+
+    // Returning a future
```
yschimke, line 120: Are we ok with ignoring the future here? It's only preview code.  I'm not really sure what else to do.

yschimke (approved):
Looks good to me. I know it's draft, but no issue landing it.

compose-tools/src/main/java/com/google/android/horologist/compose/tools/TilePreview.kt
zlandorf: Yeah, I think it's fine to ignore.

PR #1422, opened 2023-06-16
Implement OutlinedCompactChip component

DenoRS:
#### WHAT

Implement OutlinedCompactChip component 

#### WHY

resolves https://github.com/google/horologist/issues/1397

#### HOW

- Implemented OutlinedCompactChip component in compose-material based on redlines from https://github.com/google/horologist/issues/1397
- Added previews to debug folder
- Added snapshot tests

### Screenshots

image

#### Checklist :clipboard:
- [x] Add explicit visibility modifier and explicit return types for public declarations
- [x] Run spotless check
- [x] Run tests
- [x] Update metalava's signature text files

PR #1436, opened 2023-06-21
Update OutlinedCompactChip

luizgrp:
#### WHAT

Update `OutlinedCompactChip`.

#### WHY

Address comments from https://github.com/google/horologist/pull/1422 and regenerate snapshot images from a linux machine.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


compose-material/src/main/java/com/google/android/horologist/compose/material/OutlinedCompactChip.kt
```diff
@@ -139,21 +136,27 @@ public fun OutlinedCompactChip(
...
     onClick: () -> Unit,
+    modifier: Modifier = Modifier,
     icon: Any? = null,
+    iconRtlMode: IconRtlMode = IconRtlMode.Default,
     placeholder: Painter? = null,
     colors: ChipColors = ChipDefaults.outlinedChipColors(),
-    enabled: Boolean = true
+    enabled: Boolean = true,
+    interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    border: ChipBorder = ChipDefaults.outlinedChipBorder()
```
yschimke, line 148: How do you decide whether to make this a param? Why it's it hardcoded for this OutlinedCompactChip component?

yschimke (approved):

compose-material/src/main/java/com/google/android/horologist/compose/material/OutlinedCompactChip.kt
luizgrp: I thought this param would be relevant to this component in specific as it has a visible border. Whereas Chip doesn't have a border in the specs. Does it make sense? If we wanted we could also add this param to Chip to allow more customisation.

yschimke: OK, so it can be outlined with a different border, not just the boring one.  Makes sense.

I'm picturing 
image

luizgrp: do it!

PR #1447, opened 2023-06-22
AGP with compileSdk 34

yschimke:
#### WHAT

The latest androidx version bumps are using compileSdk 34.
But it requires disabling a check.

#### WHY

So we can track the newer releases.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/auth/composables/build.gradle.kts
+++ b/auth/composables/build.gradle.kts
@@ -25,7 +25,7 @@ plugins {
 }
 
 android {
-    compileSdk = 33
+    compileSdk = 34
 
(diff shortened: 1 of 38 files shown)


luizgrp (approved):

PaulWoitaschek:
This also forces all consumers of horologists to add that suppress.

yschimke:
Let me see if I can remove this on 0.4.x, publish a new version on there.  And leave 0.5.x for newer alphas etc.

PR #1453, opened 2023-06-26
Add separate constants for API33 haptics

Kpeved:
#### WHAT
Add constants for Pixel Watch API 33

#### WHY
Haptics were not working 

#### HOW
Added new constants

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
@@ -306,22 +306,35 @@ private class PixelWatchRotaryHapticFeedback(private val view: View) : RotaryHap
     override fun performHapticFeedback(
         type: RotaryHapticsType
     ) {
+        val api33orHigher = Build.VERSION.SDK_INT >= 33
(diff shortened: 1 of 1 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
```diff
@@ -306,22 +306,35 @@ private class PixelWatchRotaryHapticFeedback(private val view: View) : RotaryHap
     override fun performHapticFeedback(
         type: RotaryHapticsType
     ) {
+        val api33orHigher = Build.VERSION.SDK_INT >= 33
```
yschimke, line 309: nit: move into a field?

yschimke (approved):

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
Kpeved: removed and replaced with direct call

PR #1465, opened 2023-07-02
[MediaUI] Don't encode null into route when navigating to a media item

dabluck:
#### WHAT
When selecting a media item with a null collection, the route becomes
`mediaItem?id=<id>null`

I don't think this is the intended behavior. Instead we should skip the collection argument entirely in that case. Otherwise when parsing the argument you have to deal with the null at the end.

The sample app doesn't seem to call this function. I tried to add a simple unit test here but it doesn't look like there's any test coverage in the file so I think it'd be a lot of setup. Let me know if you want to me try though. 

Thanks

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
@dabluck thanks for the fix!

yschimke (approved):

PR #1468, opened 2023-07-04
Fix ChipIconWithProgress doc

dabluck:
#### WHAT

The view expects a percentage not a decimal. See Line 140. Also see the [tests](https://github.com/google/horologist/blob/1b6ea1102124664f17b00d18c31e3a2fcd204c00/base-ui/src/test/java/com/google/android/horologist/base/ui/components/StandardChipIconWithProgressTest.kt#L37)

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
Nice one, thanks!

PR #1483, opened 2023-07-13
Add task affinity and preview lint

yschimke:
#### WHAT

Address lint including Task Affinity

See https://developer.android.com/training/wearables/apps/launcher

#### WHY

Code Health

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/auth/sample/phone/src/main/AndroidManifest.xml
+++ b/auth/sample/phone/src/main/AndroidManifest.xml
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
   ~ Copyright 2023 The Android Open Source Project
   ~
   ~ Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


auth/sample/phone/src/main/AndroidManifest.xml
```diff
@@ -31,7 +30,7 @@
         <activity
             android:name=".MainActivity"
             android:exported="true"
-            android:label="@string/app_name"
+            android:taskAffinity=".main"
```
luizgrp, line 33: This [docs](https://developer.android.com/guide/topics/manifest/activity-element#aff) says:

> If this attribute isn't set, the activity inherits the affinity set for the application.

Would lint still complain if this is not set?

It looks like it's the [same](https://developer.android.com/training/wearables/apps/launcher#label) for label ("it defaults to the icon and label of your application tag"), so why explicitly setting `taskAffinity` but not `label`?

luizgrp (approved):

auth/sample/phone/src/main/AndroidManifest.xml
yschimke: The Wear guidance seems stronger. https://developer.android.com/training/wearables/apps/launcher#assign  Lint is complaining unless it's set. We added this recently.

For label activity>label defaults to application>label. 

But for taskAffinity when not specified, all activities default to the package name, which is likely wrong.

We could be smarter with our checks, but I think this is following the Wear docs.

PR #1486, opened 2023-07-14
Fix docs

yschimke:
#### WHAT

Fix docs.

#### WHY

image

#### HOW

Checked with https://www.drlinkcheck.com/

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


README.md
```diff
@@ -81,18 +81,18 @@ Libraries to help developers to build apps following
 the [Sign-In guidelines for Wear OS](https://developer.android.com/training/wearables/design/sign-in)
 .
 
-- [horologist-auth-composables](./auth/composables): composable screens for authentication use
+- [horologist-auth-composables](./auth-composables): composable screens for authentication use
```
luizgrp, line 84: This is tricky.

If someone is accessing https://github.com/google/horologist, the original link is the one that is correct: https://github.com/google/horologist/tree/main/auth/composables versus https://github.com/google/horologist/tree/main/auth-composables (returns 404).

However, if accessing https://google.github.io/horologist/ the link proposed by the changes is the correct: https://google.github.io/horologist/auth-composables versus https://google.github.io/horologist/auth/composables

shumelchyk (approved):

README.md
yschimke: Should we move it out of the README? have a separate index.md in docs, and trim most of README.md?

luizgrp: I think that's a good idea

yschimke: Or use absolute links? Maybe simpler.

luizgrp: They are pointing to different places at the moment:
- https://github.com/google/horologist/tree/main/auth/composables from README
- https://google.github.io/horologist/auth-composables/ from DOCS

We would have to choose one if using absolute links.

yschimke: OK, going with absolute.

luizgrp (approved):

PR #1494, opened 2023-07-17
Fix VersionFileTest

luizgrp:
#### WHAT

Fix VersionFileTest

#### WHY

Broken since project's folder was changed.

```
com.google.android.horologist.data.VersionFileTest > testVersionFileIsPresent[test(AVD) - 11] FAILED 
	expected not to be: null
	at com.google.android.horologist.data.VersionFileTest.testVersionFileIsPresent(VersionFileTest.kt:30)
```

#### HOW

Change expected version file name.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Approving to unblock, but I was trying to have file retain existing name.  Will follow up.

PR #1506, opened 2023-07-19
Improve DataLayer updates in sample

yschimke:
#### WHAT

Improve DataLayer updates in sample

#### WHY

Should use the result from the RPC call.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


auth/sample/wear/src/main/java/com/google/android/horologist/auth/sample/screens/datalayer/DataLayerViewModel.kt
```diff
@@ -49,7 +51,11 @@ class DataLayerViewModel(
     fun addDelta(i: Int) {
         viewModelScope.launch {
             try {
-                counterService.increment(counterDelta { delta = i.toLong() })
+                val newValue: CounterValue =
+                    counterService.increment(counterDelta { delta = i.toLong() })
+                uiState.update {
+                    updateIfNewer(it, newValue)
```
luizgrp, line 57: nit: could move the call to `uiState.update` to the `updateIfNewer` function (premature optimisation)

luizgrp (approved):

auth/sample/wear/src/main/java/com/google/android/horologist/auth/sample/screens/datalayer/DataLayerViewModel.kt
yschimke: Yep, I didn't want to hide the DataStore.update call, but I'm not strongly of one way or the other.  As a nit I'll leave as is, unless you have stronger opinions.

PR #1512, opened 2023-07-21
Use bigger Chips already for medium font scale.

LorenzNickel:
For some fonts, it's fine to only increase the size of the Chip for "large" (or bigger).

However, some fonts (including Google Sans) already require this chip size increase for "medium" (1.06) font scale.


luizgrp (approved):
thanks!

PR #1521, opened 2023-07-26
Move MediaPlayer background to Modifier

yschimke:
#### WHAT

Move MediaPlayer background to Modifier

#### WHY

Drawing the background via a Modifier allows avoiding Offscreen Composition for Marquee Text.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media/ui/src/main/java/com/google/android/horologist/media/ui/components/background/ArtworkColorBackground.kt
```diff
@@ -43,18 +44,63 @@ import coil.request.ImageRequest
  * Background using a radial gradient extracted from artwork.
  */
 @Composable
+@Deprecated("Prefer background modifier")
```
luizgrp, line 47: is the experimental annotation missing for all the components of this package?

media/ui/src/test/snapshots/images/com.google.android.horologist.media.ui_MediaPlayerA11yScreenshotTest_mediaPlayerLargeRound.png
luizgrp, line 1: screenshots seem to be identical, were they updated due to generating them in a non linux machine?

luizgrp (approved):

media/ui/src/main/java/com/google/android/horologist/media/ui/components/background/ArtworkColorBackground.kt
yschimke: I'll add.  I was going to suggest we remove a bunch of these from 0.5.x

luizgrp: Did you mean the ones from https://github.com/google/horologist/pull/1522 ?

I meant the single one that is not deprecated, ExperimentalHorologistApi

yschimke:
Example diffs 
image

image

PR #1523, opened 2023-07-26
Graduate auth experimental APIs

yschimke:
#### WHAT

Graduate auth experimental APIs.

#### WHY

Bump to a new major version, APIs seem relatively stable. Even with this we can change APIs, either deprecating, or just giving a headsup (we are not androidx).

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
Happy to close, or change some.  But wanted to spark a discussion, not force the result.

luizgrp (commented):
The ones from packages `devicegrant` and `pkce` should still be marked as experimental. 

We should have a way of listing all the components that were marked as "stable" so once we see a component without the annotation, we know that is not "missing" but it was removed on purpose. Maybe an annotation "HorologistApi"?

yschimke:
Will fix those packages before landing.

I agree, but not super keen on a permanent annotation on almost everything.

yschimke:
Assuming this is soft approved.

PR #1544, opened 2023-08-02
Add a private copy of a Picker and PickerGroup 

Kpeved:
#### WHAT
Add a private copy of a Picker and PickerGroup into repository

#### WHY
We need them for adding a rotary improvement for TimePicker and DatePicker

#### HOW
Classes were copied 1:1 except they don't have any public api

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):
Assuming classes were copied without changes - LGTM

PR #1551, opened 2023-08-04
Remove adjustChipHeightToFontScale function

luizgrp:
#### WHAT

Remove `adjustChipHeightToFontScale` function.

#### WHY

It should be redundant in [wear-compose 1.3.0-alpha02](https://developer.android.com/jetpack/androidx/releases/wear-compose#1.3.0-alpha02):

> Chip and ToggleChip will now adjust their height to accommodate content that has grown due to large fonts for accessibility, when required.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp:
cc @LorenzNickel

yschimke (approved):

PR #1561, opened 2023-08-08
Deprecate rotaryWithFling modifier 

Kpeved:
#### WHAT
rotaryWithFling was deprecated in favour of rotaryWithScroll, which now does the fling. 
RotaryMode.Fling was also deprecated

rememberActiveFocusRequester was also added as a default focusRequester for rotaryWithScroll and rotaryWithSnap modifiers
#### WHY
Those 2 modifiers were confusing - by default developers were using rotaryWithScroll, which didn't have a fling. Now it should be easier to understand .

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -549,12 +548,9 @@ internal fun pickerGroupItemWithRSB(
     readOnlyLabel: @Composable (BoxScope.() -> Unit)? = null,
     option: @Composable PickerScope.(optionIndex: Int, pickerSelected: Boolean) -> Unit
 ): PickerGroupItem {
-    val localFocusRequester = rememberActiveFocusRequester()
-
     return PickerGroupItem(
         pickerState = pickerState,
         modifier = modifier.rotaryWithSnap(
-            localFocusRequester,
```
BowerSteve, line 557: I see - this will be assigned in PickerGroup by default.

yschimke (approved):

PR #1562, opened 2023-08-08
Fix layout of SignedInConfirmationDialog

luizgrp:
#### WHAT

Fix layout of `SignedInConfirmationDialog`.

| Before | After |
|--------|--------|
| CurrentSpecs | Screenshot_20230808_145455 |

#### WHY

Avatar was too close to time text on small round devices.

#### HOW

Remove bottom padding.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
Should we have some small round previews, it looks like these are large.

yschimke (approved):

luizgrp:
> Should we have some small round previews, it looks like these are large.

The previews are using WearPreviewDevices annotation, so all of them are available for small devices too. 

I think you mean snapshot tests? As far as I remember, long time ago we made a decision to only have tests for large round devices. Doing a quick test I can see only 1 test for small round. Should we review this, and improve all the tests?

yschimke:
We probably should, we'll have more variety of sizes over time.  But not an issue for now.

PR #1580, opened 2023-08-12
Add icon color based on enabled field

plksharma30:
This pull request introduces an enhancement that adds visual indicators to the custom command icons based on their enabled/disabled state.


luizgrp (commented):
Looking good, it failed in CI as it needs to update the metalava files, please run the gradle task `metalavaGenerateSignatureRelease`

PR #1589, opened 2023-08-18
Adds support for setup complete state to DataLayer app helper

garanj:
#### WHAT

Adds the ability to store / read app setup complete via Data Layer app helper

#### WHY

For some apps, it is useful for the phone to know whether the user has completed all the necessary setup steps on the watch, such as logging in. In the case where the user hasn't logged in, the phone may show different options, or prompt the user to log in for example.

#### HOW

Moves the `ActivityLaunched` state into an enum to cover the following states:

- Unspecified
- Launched (at least) once
- Setup complete

With the behaviour:
- Marking setup as complete trumps launched once (so setting launched once after setup complete has no affect - the implication for now is that the app would need to be launched at least once to complete setup)
- Unmarking setup as complete returns to the launched once state - if setup had previously been completed, but has no effect if the state is currently unspecified

TODO: Add use of the `markSetupComplete` and `markSetupNoLongerComplete` hooks to the sign-in flows in the auth sample.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/core/src/main/proto/app_helper_pb.proto
+++ b/datalayer/core/src/main/proto/app_helper_pb.proto
@@ -54,6 +54,11 @@ message TileInfo {
   .google.protobuf.Timestamp timestamp = 3;
 }
 
+message ActivityLaunched {
+  bool activityLaunchedOnce = 1;
+  .google.protobuf.Timestamp timestamp = 2;
+}
+
 enum UsageStatus {
   USAGE_STATUS_UNSPECIFIED = 0;
   USAGE_STATUS_LAUNCHED_ONCE = 1;
@@ -69,6 +74,7 @@ message SurfacesInfo {
   reserved 1;
   repeated ComplicationInfo complications = 2;
   repeated TileInfo tiles = 3;
-  reserved 5;
-  UsageInfo usageInfo = 6;
+  // Temporarily keep [ActivityLaunched] - remove at a later date
+  ActivityLaunched activityLaunched = 4;
+  UsageInfo usageInfo = 5;
 }
--- a/datalayer/core/src/main/proto/app_helper_pb.proto
+++ b/datalayer/core/src/main/proto/app_helper_pb.proto
@@ -54,8 +54,14 @@ message TileInfo {
   .google.protobuf.Timestamp timestamp = 3;
 }
 
-message ActivityLaunched {
-  bool activityLaunchedOnce = 1;
+enum UsageStatus {
+  USAGE_STATUS_UNSPECIFIED = 0;
+  USAGE_STATUS_LAUNCHED_ONCE = 1;
+  USAGE_STATUS_SETUP_COMPLETE = 2;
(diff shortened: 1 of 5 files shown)


datalayer/core/src/main/proto/app_helper_pb.proto
```diff
@@ -54,14 +54,21 @@ message TileInfo {
...
-message ActivityLaunched {
-  bool activityLaunchedOnce = 1;
+enum UsageStatus {
+  USAGE_STATUS_UNSPECIFIED = 0;
+  USAGE_STATUS_LAUNCHED_ONCE = 1;
+  USAGE_STATUS_SETUP_COMPLETE = 2;
+}
+
+message UsageInfo {
+  UsageStatus usageStatus = 1;
```
yschimke, line 64: burn the 1 (unused), or keep it for some transition period.

Either way make this id 3.

datalayer/core/src/main/proto/app_helper_pb.proto
```diff
@@ -54,14 +54,21 @@ message TileInfo {
...
   .google.protobuf.Timestamp timestamp = 2;
 }
 
 message SurfacesInfo {
   reserved 1;
   repeated ComplicationInfo complications = 2;
   repeated TileInfo tiles = 3;
-  ActivityLaunched activityLaunched = 4;
+  reserved 5;
```
yschimke, line 72: why reserved?

yschimke, line 72: If someone is using this, we may want to keep it temporarily. populate both for a bit.

datalayer/core/src/main/proto/app_helper_pb.proto
yschimke, line 64: Do we want any other info here? Like logged in status? User may setup and choose not to log in.

yschimke (approved):

datalayer/core/src/main/proto/app_helper_pb.proto
garanj: Great point. Just so I understand, this means keep this proto entry and then also ensure the code is populating both?

garanj: (temporarily, that is)

datalayer/core/src/main/proto/app_helper_pb.proto
garanj: At the moment, it is intentionally vague - the app determines what constitutes setup complete - whether that means logged in, all necessary permissions granted etc.

We could break it out into separate categories (permissions might be another one?), but maybe simple is best to start?

datalayer/core/src/main/proto/app_helper_pb.proto
yschimke: Yep.

PR #1609, opened 2023-08-24
Avoid deprecated preview APIs

yschimke:
#### WHAT

Avoid deprecated APIs.

#### WHY

Clean builds.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):
Thanks

PR #1612, opened 2023-08-25
Integrate media app with ExoPlayer / Media3-ui playback suppression on unsuitable output solution

anubhavkakkar:
#### WHAT
Integrate media app with ExoPlayer / Media3-ui playback suppression on unsuitable output solution

#### WHY
The playback suppression solution on unsuitable output helps handling the cases of preventing unintended playback on the builtin speaker of watch. This includes launching the system output switcher when playback is suppression due to this reasom, resuming the playback automatically after the situation of unavailability of the suitable outputs is resolved, and also pausing the playback automatically when all the suitable outputs have been removed during an ongoing playback.

#### HOW
The media3 library version is changed to 1.2.0-alpha01 and the corresponding APIs namely ExoPlayer.setSuppressPlaybackOnUnsuitableOutput & WearUnsuitableOutputPlaybackSuppressionResolverListener have been used for the ExoPlayer module of PlaybackServiceModule. Besides, the internal implementations has been removed.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media/backend-media3/src/main/java/com/google/android/horologist/media3/WearConfiguredPlayer.kt
```diff
@@ -82,40 +53,12 @@ public class WearConfiguredPlayer(
     }
 
     private suspend fun attemptPlay(mediaItem: MediaItem) = coroutineScope {
-        val currentAudioOutput = audioOutputRepository.audioOutput.value
-
         if (!playbackRules.canPlayItem(mediaItem)) {
```
yschimke, line 56: I don't think it's worth keeping just for this.  This was there in case you want to avoid playing a streaming song, checking for downloads first.  But it was secondary always to the Speaker case.

Should we remove this class entirely?

media/sample/src/main/AndroidManifest.xml
```diff
@@ -123,7 +123,8 @@
 
         <service
             android:name=".data.service.download.MediaDownloadServiceImpl"
-            android:exported="false">
+            android:exported="false"
+            android:foregroundServiceType="dataSync">
```
yschimke, line 127: Is dataSync recommended for media downloads? I guess makes sense.

media/backend-media3/src/main/java/com/google/android/horologist/media3/rules/PlaybackRules.kt
```diff
@@ -31,23 +30,13 @@ public interface PlaybackRules {
      */
     public suspend fun canPlayItem(mediaItem: MediaItem): Boolean
 
-    /**
-     * Can Media be played with the given audio target.
-     */
-    public fun canPlayWithOutput(audioOutput: AudioOutput): Boolean
-
     @ExperimentalHorologistApi
     public object Normal : PlaybackRules {
```
yschimke, line 34: We can probably remove now.

media/sample/src/main/AndroidManifest.xml
anubhavkakkar: Just for background, we need to specify one out of many foregroundServiceType listed in https://developer.android.com/about/versions/14/changes/fgs-types-required#data-sync for the foreground services. From the description of dataSync in documentation (ref: https://developer.android.com/about/versions/14/changes/fgs-types-required#data-sync), it looked closest to me.

media/backend-media3/src/main/java/com/google/android/horologist/media3/WearConfiguredPlayer.kt
anubhavkakkar: yes, it looks to be making sense to remove this class.
I've removed this now.
Please take a look.

yschimke:
So much red, lovely.

You will need to merge or rebase the changes in.

yschimke (approved):

yschimke:
Run `./gradlew spotlessApply` to fix these violations.

PR #1638, opened 2023-09-01
Add fixup commands

yschimke:
#### WHAT

Add fixup commands

#### WHY

Try a manual workflow instead of automatic.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/.github/workflows/fixup.yml
+++ b/.github/workflows/fixup.yml
@@ -1,0 +1,65 @@
+name: Fixups
+
+on:
+  workflow_dispatch:
+
+jobs:
+  screenshots:
+    runs-on: ubuntu-latest
+    timeout-minutes: 40
+
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          lfs: 'true'
+
+      - name: Copy CI gradle.properties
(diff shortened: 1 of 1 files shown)


yschimke:
@luizgrp going to land to test, I have a broken PR

.github/workflows/fixup.yml
```diff
@@ -0,0 +1,65 @@
...
+
+      - name: Reformat
+        uses: gradle/gradle-build-action@v2
+        with:
+          arguments: spotlessApply
+
+      - name: Push reformatted code
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          file_pattern: '**/*.kt'
```
luizgrp, line 63: this reminded me that we don't have groovy files anymore since they were ported to kotlin dsl, so [this](https://github.com/google/horologist/blob/9726603e4a54b803fa0ca58345189301416fda44/build.gradle.kts#L100-L108) should be updated to validate the kts files

luizgrp (commented):
nice!

PR #1640, opened 2023-09-01
Modify to launch output switcher from the volume screen if that is available

anubhavkakkar:
Modify to launch output switcher from the volume screen if that is available

#### WHAT
Changes to check if output switcher is present and launch that from the output switcher UI button on the volume screen.

#### WHY
Currently, the code only launches the BT settings whereas output switcher may also be present which is could be a more appropriate system dialog to switcher between the media outputs so it is better to check for the same launch that if needed.

#### HOW
The modified code will check for a system package of the output switcher and launch that if that is found.

#### Checklist :clipboard:
- [+] Add explicit visibility modifier and explicit return types for public declarations
- [+] Run spotless check
- [+] Run tests
- [+] Update metalava's signature text files


media/audio/src/main/java/com/google/android/horologist/audio/OutputSwitcher.kt
```diff
@@ -0,0 +1,68 @@
...
+import android.content.Intent
+import android.content.pm.ApplicationInfo
+
+/**
+ * Support for launching the system output switcher.
+ *
+ * https://developer.android.com/guide/topics/media/media-routing#output-switcher.
+ *
+ */
+public object OutputSwitcher {
```
yschimke, line 30: These presumably need API updates.

yschimke (approved):

yschimke:
I'll fix the API in a follow up

PR #1644, opened 2023-09-04
Add some initial Data Layer docs

yschimke:
#### WHAT

Add some initial Data Layer docs

#### WHY

Missing docs, so not advertised.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):
Nice!

Issue #1550, opened 2023-08-04
Metalava check not failing for some modules


luizgrp:
_Originally posted by @luizgrp in https://github.com/google/horologist/pull/1548#discussion_r1284209481_

PR #1647, opened 2023-09-05
Run metalava and fix media audio api

laiyichin:
#### WHAT
Run metalava and fix media audio api

#### HOW
`./gradlew spotlessApply spotlessCheck compileDebugSources compileReleaseSources metalavaGenerateSignature metalavaGenerateSignatureRelease lintDebug`

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/media/audio/api/current.api
+++ b/media/audio/api/current.api
@@ -90,6 +90,11 @@ package com.google.android.horologist.audio {
     field public static final com.google.android.horologist.audio.BluetoothSettings INSTANCE;
   }
 
+  public final class OutputSwitcher {
+    method public boolean launchSystemMediaOutputSwitcherUi(android.content.Context);
+    field public static final com.google.android.horologist.audio.OutputSwitcher INSTANCE;
+  }
+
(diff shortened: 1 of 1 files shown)


yschimke:
@laiyichin is there a extra check we can add into a workflow file?

https://github.com/google/horologist/blob/main/.github/workflows/build.yml#L134

yschimke (approved):

laiyichin:
> @laiyichin is there a extra check we can add into a workflow file?
> 
> https://github.com/google/horologist/blob/main/.github/workflows/build.yml#L134

I tried running ` ./gradlew --stacktrace media:audio:metalavaCheckCompatibilityDebug media:audio:metalavaCheckCompatibilityRelease` and it just passes. Running `./gradlew --stacktrace media:audio:metalavaGenerateSignatureRelease` would generate the change, though. 

Will need more time to investigate why the check does not work. luizgrp@ pointed out to me that there is an [existing issue filed](https://github.com/google/horologist/issues/1550) already. I will merge this PR in first, and if I found anything i will associate with that issue.

PR #1649, opened 2023-09-06
Add a test of WorkManager

yschimke:
#### WHAT

Robolectric test of SyncWorker

#### WHY

Confirm Work isn't broken.

#### HOW

Robolectric App test.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/media/sample/src/test/java/com/google/android/horologist/mediasample/work/FakeNetworkModule.kt
+++ b/media/sample/src/test/java/com/google/android/horologist/mediasample/work/FakeNetworkModule.kt
@@ -23,13 +23,13 @@ import com.google.android.horologist.mediasample.di.NetworkModule
 import com.google.android.horologist.networks.data.DataRequestRepository
 import com.google.android.horologist.networks.data.InMemoryDataRequestRepository
 import com.google.android.horologist.networks.data.NetworkType
-import com.google.android.horologist.networks.testdoubles.FakeNetworkRepository
 import com.google.android.horologist.networks.highbandwidth.HighBandwidthConnectionLease
 import com.google.android.horologist.networks.highbandwidth.HighBandwidthNetworkMediator
 import com.google.android.horologist.networks.logging.NetworkStatusLogger
 import com.google.android.horologist.networks.okhttp.impl.FailedCall
 import com.google.android.horologist.networks.request.HighBandwidthRequest
 import com.google.android.horologist.networks.status.NetworkRepository
+import com.google.android.horologist.networks.testdoubles.FakeNetworkRepository
 import com.google.android.horologist.test.toolbox.testdoubles.FakeUampService
 import dagger.Module
 import dagger.Provides


network-awareness/core/api/current.api
```diff
@@ -247,6 +247,28 @@ package com.google.android.horologist.networks.data {
 
 }
 
+package com.google.android.horologist.networks.fake {
+
+  public final class FakeNetworkRepository implements com.google.android.horologist.networks.status.NetworkRepository {
```
yschimke, line 252: It's not worth a test artifact, and they are useful, we already had two of them.

media/sample/build.gradle.kts
```diff
@@ -250,6 +256,10 @@ dependencies {
     testImplementation(libs.kotlinx.coroutines.test)
     testImplementation(libs.turbine)
     testImplementation(libs.robolectric)
+    testImplementation(libs.androidx.work.ktx)
+    testImplementation("com.google.dagger:hilt-android-testing:2.48")
```
luizgrp, line 260: nit: move to toml file?

network-awareness/core/api/current.api
luizgrp: do you mean "it is worth"?

media/sample/src/test/java/com/google/android/horologist/mediasample/work/BaseAppTest.kt
```diff
@@ -0,0 +1,33 @@
...
+import dagger.hilt.android.testing.HiltTestApplication
+import org.junit.Rule
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import org.robolectric.annotation.GraphicsMode
+
+@RunWith(RobolectricTestRunner::class)
+@Config(application = HiltTestApplication::class, sdk = [30])
+@GraphicsMode(GraphicsMode.Mode.NATIVE)
```
luizgrp, line 29: is this needed?

network-awareness/core/src/main/java/com/google/android/horologist/networks/fake/FakeNetworkRepository.kt
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.google.android.horologist.networks.rules.helpers
+package com.google.android.horologist.networks.fake
```
luizgrp, line 17: nit: maybe "testdouble" instead of "fake" so no need to create separate packages for mocks, stubs etc

luizgrp (approved):
LGTM

network-awareness/core/api/current.api
yschimke: I mean't test module, an extra jar.

luizgrp: yeah I get it, but do you think it is **not** worth it having it? I think we should, it's used in many places, just wondering why you think otherwise

PR #1670, opened 2023-09-11
Preview bolded content

yschimke:
#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


composables/src/debug/java/com/google/android/horologist/composables/DatePickerPreview.kt
```diff
@@ -17,18 +17,55 @@
...
+@Composable
+@Preview(
+    device = WearDevices.LARGE_ROUND,
+    showSystemUi = true,
+    backgroundColor = 0xff000000,
+    showBackground = true,
+    group = "Fonts - Small",
+    fontScale = 0.94f,
+)
+fun DatePickerPreviewLarge() {
```
luizgrp, line 66: nit: small

here and in the other previews

composables/src/test/kotlin/com/google/android/horologist/composables/DatePickerTest.kt
```diff
@@ -36,4 +40,20 @@ class DatePickerTest : ScreenshotBaseTest(
             )
         }
     }
+
+    @Test
+    @Config(
+        qualifiers = "+w192dp-h192dp",
+        fontScale = 1.24f,
+    )
+    fun datePickerSmallBold() {
```
luizgrp, line 49: nit: "LargeFontBold"?

here and in the other tests

composables/src/test/kotlin/com/google/android/horologist/composables/DatePickerTest.kt
```diff
@@ -36,4 +40,20 @@ class DatePickerTest : ScreenshotBaseTest(
...
+        screenshotTestRule.setContent(takeScreenshot = true) {
+            MaterialTheme(typography = MaterialTheme.typography.copy { this.copy(fontWeight = FontWeight.Bold) }) {
```
luizgrp, line 51: I wonder if this can be done via TestHarness as we do with fontScale. I see a `fontWeightAdjustment` param but haven't tried it

composables/build.gradle.kts
```diff
@@ -109,6 +109,7 @@ dependencies {
     implementation(libs.androidx.corektx)
 
     implementation(libs.compose.ui.toolingpreview)
+    implementation(libs.androidx.wear.tooling.preview)
```
luizgrp, line 112: nit: debugImplementation, and if it's used for tests, testImplementation?

luizgrp (approved):
LGTM

composables/src/test/kotlin/com/google/android/horologist/composables/DatePickerTest.kt
yschimke: It won't work on API 30, so I think this is simpler here.  I'll add a comment.

composables/build.gradle.kts
yschimke: I think implementation is right here.  The ui.tooling is the debug one.

composables/src/debug/java/com/google/android/horologist/composables/DatePickerPreview.kt
yschimke: I meant large device, so I made it clearer.

composables/src/test/kotlin/com/google/android/horologist/composables/DatePickerTest.kt
yschimke: Made it clearer, this meant devices, which was confusing.

PR #1672, opened 2023-09-12
Adds AmbientAware and an example usage, AmbientAwareTime

garanj:
#### WHAT

Provides the ability to show different UI in interactive vs ambient mode, and respond to ambient update events, when the app is configured to support Always-On

#### WHY

It can be useful to show some basic UI whilst in ambient mode

#### HOW

Implements a wrapper around AmbientLifecycleObserver

#### Checklist :clipboard:
- [ x] Run spotless check
- [ x] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2023 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.compose.ambient
+
+import android.app.Activity
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.LocalLifecycleOwner
+import androidx.wear.ambient.AmbientLifecycleObserver
+
+/**
+ * Composable for general handling of changes and updates to ambient status. A new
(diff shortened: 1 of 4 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
```diff
@@ -0,0 +1,77 @@
...
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.LocalLifecycleOwner
+import androidx.wear.ambient.AmbientLifecycleObserver
+
+/**
+ * Composable for general handling of changes and updates to ambient status. A new
+ * [AmbientStateUpdate] is generated with any change of ambient state, as well as with any periodic
+ * update generated whilst the screen is in ambient mode.
```
yschimke, line 33: We should probably comment that it should be applied high up near the activity, since it affects the state of the activity, calling setAmbientMode.

compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientStateUpdate.kt
```diff
@@ -0,0 +1,32 @@
...
+ * Represent Ambient as updates, with the state and time of change. This is necessary to ensure that
+ * when the system provides a (typically) 1min-frequency callback to onUpdateAmbient, the developer
+ * may wish to update composables, but the state hasn't changed.
+ */
+data class AmbientStateUpdate(
+    val ambientState: AmbientState,
+    val changeTimeMillis: Long = System.currentTimeMillis(),
+)
+
+enum class AmbientState {
```
yschimke, line 29: Should these be richer?

```
sealed interface AmbientState {
  class Ambient(val details: AmbientDetails)
  object Interactive
} 
```

Simplifying and removing the AmbientDetails seems limiting. Unless they aren't useful?

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/wear/src/main/java/androidx/wear/ambient/AmbientLifecycleObserver.kt;l=128

compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAwareTime.kt
```diff
@@ -0,0 +1,89 @@
...
+ *      }
+ *  }
+ *
+ * @param stateUpdate The state update from [AmbientAware]
+ * @param updatePeriodMillis The update period, whilst in interactive mode
+ * @param block The developer-supplied composable for rendering the date and time.
+ */
+@RequiresApi(Build.VERSION_CODES.O)
+@Composable
+fun AmbientAwareTime(
```
yschimke, line 63: Since sort of feels in between two worlds.

a) AmbientAware - that works for anything?
b) AmbientAwareTimestamp - that could handle the formatter for you.

yschimke (approved):
Suggest landing as-is and iterating so it doesn't blow up in size.

garanj:
Thanks @yschimke, landing now, follow ups to address the issues raised.

PR #1674, opened 2023-09-13
Expands documentation for AmbientAware, adds support for AmbientDetails

garanj:
#### WHAT

Provides further guidance on using `AmbientAware` and adds support for `AmbientDetails`

#### WHY

Using the `AmbientAware` composable changes the behavior of the `Activity`, so should ideally be used high up in the composable tree. This needs clarifying in the documentation.

`AmbientDetails` are provided by the system when entering ambient mode, and can be used by the app to make decisions about how to render the UI.

#### HOW

Updates kdoc and adds `AmbientDetails` as a property of the `Ambient` state.

#### Checklist :clipboard:
- [ x] Run spotless check
- [ x] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
@@ -31,6 +31,12 @@ import androidx.wear.ambient.AmbientLifecycleObserver
  * Composable for general handling of changes and updates to ambient status. A new
  * [AmbientStateUpdate] is generated with any change of ambient state, as well as with any periodic
  * update generated whilst the screen is in ambient mode.
+ *
+ * This composable changes the behavior of the activity, enabling Always-On. See:
+ *
+ *     https://developer.android.com/training/wearables/views/always-on).
+ *
+ * It should therefore be used high up in the tree of composables.
  */
 @Composable
 fun AmbientAware(block: @Composable (AmbientStateUpdate) -> Unit) {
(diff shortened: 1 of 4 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
```diff
@@ -41,23 +47,23 @@ fun AmbientAware(block: @Composable (AmbientStateUpdate) -> Unit) {
...
             }
 
             override fun onExitAmbient() {
-                ambientUpdate = AmbientStateUpdate(AmbientState.INTERACTIVE)
+                ambientUpdate = AmbientStateUpdate(AmbientState.Interactive)
             }
 
             override fun onUpdateAmbient() {
-                ambientUpdate = AmbientStateUpdate(AmbientState.AMBIENT)
+                ambientUpdate = AmbientStateUpdate(AmbientState.Ambient())
```
yschimke, line 58: Can we retain the previous value?

yschimke (approved):

compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
garanj: Yeah, I wondered about that, do these things change? I will add it.

yschimke: I don't think it's about them changing, I think more about having immediate access to it via this type.

compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
```diff
@@ -31,33 +31,48 @@ import androidx.wear.ambient.AmbientLifecycleObserver
...
  */
 @Composable
 fun AmbientAware(block: @Composable (AmbientStateUpdate) -> Unit) {
     val activity = LocalContext.current as Activity
     val lifecycle = LocalLifecycleOwner.current.lifecycle
     var ambientUpdate by remember { mutableStateOf<AmbientStateUpdate?>(null) }
+    // On entering ambient mode, the [onEnterAmbient] callback method provides details of burn-in
+    // protection and low-bit support. These are stored here such that they can be provided as a
+    // convenience with each [AmbientStateUpdate] as [onUpdateAmbient] does not provide them.
+    var lastAmbientDetails by remember {
```
yschimke, line 49: Do you need this extra field if you have the field above for the overall state?

compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
garanj: Added. Unfortunately, there still remains the chance of it being null, if the device is already in ambient, so `onEnterAmbient` has not been called. I don't think this seems very likely in practice?

yschimke (approved):

garanj:
@yschimke - modified to remove the need for the extra field for the last ambient details.

compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
```diff
@@ -64,13 +55,20 @@ fun AmbientAware(block: @Composable (AmbientStateUpdate) -> Unit) {
             }
 
             override fun onUpdateAmbient() {
+                val lastAmbientDetails = ambientUpdate?.ambientState?.let {
+                    if (it is AmbientState.Ambient) {
+                        it.ambientDetails
+                    } else {
+                        null
+                    }
+                }
```
yschimke, line 64:
```suggestion
                val lastAmbientDetails = (ambientUpdate?.ambientState as? AmbientState.Ambient).ambientDetails
```

PR #1679, opened 2023-09-14
AmbientAware isAlwaysOnScreen flag

garanj:
#### WHAT

Adds support for selectively enabling or disabling always-on

#### WHY

Some screens will require always-on, but some will not, for example, a workout screen (yes) vs a workout summary screen (no). 

#### HOW

`AmbientAware` now takes an optional parameter to specify whether the screen should enable always-on or not. This can be hoisted above the use of `AmbientAware` e.g.

```
AmbientAware(isAlwaysOnScreen == currentScreen.useAlwaysOn) { ambientStateUpdate ->

// ...

}
```

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Really elegant approach.  Nice!

PR #1682, opened 2023-09-15
Update RepeatableClickable to include a onLongRepeatableClickEnd 

laiyichin:
#### WHAT
- Update RepeatableClickable to include a onLongRepeatableClickEnd. 
- Update the parameter names to be more accurately describing the behaviours.
- Update the repeatable clickable to be internal only.

#### WHY
This allows clients using this modifier to have a different trigger when the repeatable clickable calls end. For example, repeated calls could be accumulating a value until the finger up event and only send the update to the backend with one finish call. 

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Optional, but is it worth adding a sample, or will you add to the VolumeScreen?

laiyichin:
> Optional, but is it worth adding a sample, or will you add to the VolumeScreen?

Yeap, follow-up PR in media coming!

PR #1684, opened 2023-09-17
Improve Galaxy Watch Classic detection for haptics

Dubzer:
#### WHAT
Improvement to the regex used to detect the Samsung Galaxy Watch model with a bezel. 

#### WHY
Because the previous regex missed 4 Classic 42mm and the newer 6 Classic models.

#### HOW
Match all Classic model numbers specified in:
https://en.wikipedia.org/wiki/Samsung_Galaxy_Watch_4
https://en.wikipedia.org/wiki/Samsung_Galaxy_Watch_6

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
@@ -346,7 +346,7 @@ private class PixelWatchRotaryHapticFeedback(private val view: View) : RotaryHap
 }
 
 /**
- * Implementation of [RotaryHapticFeedback] for Galaxy Watch 4 Classic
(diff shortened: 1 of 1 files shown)


yschimke:
It would be good to make this method visible for testing (internal?) and have a unit test listing devices, but doesn't need to be in your PR.

yschimke (approved):

Thomas-Vos:
Could this be backported to the release-0.4.x branch, please?

yschimke:
PR here https://github.com/google/horologist/pull/1742, can release tomorrow if it lands.

yschimke:
0.4.14 is released, should show up in a couple of hours.  Let me know if there are any issues, I haven't tested on device, but on emulator it's working ok for scrolling, rsb, position indicator.

PR #1686, opened 2023-09-18
Remove copy paste of Pickers code

yschimke:
#### WHAT

Cleanup pickers

#### WHY

Since update to Wear Compose 1.3, slightly more options to implement pickers.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


composables/src/main/java/com/google/android/horologist/composables/picker/PickerRotaryScrollAdapter.kt
```diff
@@ -16,19 +16,21 @@
...
  * An extension function for creating [RotaryScrollAdapter] from [Picker]
  */
-internal fun PickerState.toRotaryScrollAdapter(): PickerRotaryScrollAdapter =
+fun PickerState.toRotaryScrollAdapter(): RotaryScrollAdapter =
     PickerRotaryScrollAdapter(this)
 
 /**
  * An implementation of rotary scroll adapter for [Picker]
  */
+@Suppress("INVISIBLE_MEMBER")
```
yschimke, line 33: I need to test importing this into Google repo, make sure build doesn't fail on this.

Also should scalingLazyListState be public?

PR #1688, opened 2023-09-18
Update animated media buttons to have repeated calls

laiyichin:
#### WHAT
Update animated media buttons to have repeated calls for seeking forward and seeking backward

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media/sample/src/main/java/com/google/android/horologist/mediasample/ui/player/UampMediaPlayerScreen.kt
```diff
@@ -95,8 +95,10 @@ fun UampMediaPlayerScreen(
                         playPauseButtonEnabled = playerUiState.playPauseEnabled,
                         playing = playerUiState.playing,
                         onSeekToPreviousButtonClick = { playerUiController.skipToPreviousMedia() },
+                        onSeekToPreviousLongRepeatableClick = { playerUiController.seekBack() },
```
yschimke, line 98: That is lovely!

media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/AnimatedMediaButton.kt
```diff
@@ -48,6 +48,8 @@ public fun AnimatedMediaButton(
     compositionResult: LottieCompositionResult,
     contentDescription: String,
     modifier: Modifier = Modifier,
+    onLongRepeatableClick: () -> Unit = {},
```
yschimke, line 51: Does this end up being an onClick when left as defaults?

media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/AnimatedMediaControlButtons.kt
```diff
@@ -49,6 +49,10 @@ public fun AnimatedMediaControlButtons(
     onSeekToNextButtonClick: () -> Unit,
     seekToNextButtonEnabled: Boolean,
     modifier: Modifier = Modifier,
+    onSeekToPreviousLongRepeatableClick: () -> Unit = {},
+    onSeekToPreviousLongRepeatableClickEnd: () -> Unit = {},
+    onSeekToNextLongRepeatableClick: () -> Unit = {},
+    onSeekToNextLongRepeatableClickEnd: () -> Unit = {},
```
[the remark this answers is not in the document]
laiyichin: I was thinking to keep them separate, so if client does not pass in any value for these repeatable calls it'd just be no-op.

media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/AnimatedMediaButton.kt
laiyichin: This separate from `onClick` in line 47 and would just be no-op if client decided not to pass in a value.

PR #1689, opened 2023-09-18
Implementation of RotaryScrollAdapter for AndroidX Picker

yschimke:
#### WHAT

Small Picker API for RotaryScrollAdapter

#### WHY

Picker and PickerState are public APIs, so the internal implementations used by DatePicker and TimePicker should remain so.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/composables/src/main/java/com/google/android/horologist/composables/picker/PickerRotaryScrollAdapter.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/picker/PickerRotaryScrollAdapter.kt
@@ -58,7 +58,8 @@ internal class PickerRotaryScrollAdapter(
 }
 
 /**
- * Temporary implementation of RotaryScrollAdapter for PickerState.
+ * Temporary implementation of RotaryScrollAdapter for PickerState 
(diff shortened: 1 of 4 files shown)


Kpeved:
Why it's called "Small picker" ? I guess would be better to have something like " Implementation of RotaryScrollAdapter for AndroidX Picker. "

composables/src/main/java/com/google/android/horologist/composables/picker/PickerRotaryScrollAdapter.kt
```diff
@@ -55,3 +56,41 @@ internal class PickerRotaryScrollAdapter(
     override fun totalItemsCount(): Int =
         scrollableState.scalingLazyListState.layoutInfo.totalItemsCount
 }
+
+/**
+ * Temporary implementation of RotaryScrollAdapter for PickerState.
```
Kpeved, line 61: Better to add "PickerState  from AndroidX wear-compose library"

Kpeved (approved):

PR #1690, opened 2023-09-18
Add a seek projection track position model

laiyichin:
#### WHAT
Add a new type of `TrackPositionUiModel` for displaying seeking projection. 

#### WHY
This gives clients an opportunity to show seek projection in addition to the Actual, Loading, Predictive ui models. For example, this could enable `PlayerRepositoryImpl` to accumulate seeking duration and displaying it without making multiple repeated calls to `seekBack` or `seekForward` on long press and only flush the position on finger up.  

#### HOW
When the seek projection is passed in (as duration), use the `TrackPositionUiModel`

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media/ui/src/main/java/com/google/android/horologist/media/ui/state/mapper/TrackPositionUiModelMapper.kt
+++ b/media/ui/src/main/java/com/google/android/horologist/media/ui/state/mapper/TrackPositionUiModelMapper.kt
@@ -20,6 +20,7 @@ import com.google.android.horologist.annotations.ExperimentalHorologistApi
 import com.google.android.horologist.media.model.PlaybackStateEvent
 import com.google.android.horologist.media.model.PlayerState
 import com.google.android.horologist.media.ui.state.model.TrackPositionUiModel
+import kotlin.time.Duration
 
 /**
  * Functions to map a [TrackPositionUiModel] based on data from other layers.
@@ -38,7 +39,7 @@ public object TrackPositionUiModelMapper {
             return TrackPositionUiModel.Hidden
         }
         event.playbackState.seekProjection?.let { seek ->
-            if (seek.inWholeMilliseconds >= 0) {
+            if (seek > Duration.ZERO) {
                 val percent = seek.inWholeMilliseconds.toFloat() / durationMs.toFloat()
                 return TrackPositionUiModel.SeekProjection(percent, duration, seek)
             }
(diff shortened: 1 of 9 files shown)


yschimke:
I'd defer to @fstanis . I guess I'm not sure how best to model it, or whether it could be handled closer to the UI.  But this design would seems nicely flexible.

fstanis (approved):

media/ui/src/main/java/com/google/android/horologist/media/ui/state/mapper/TrackPositionUiModelMapper.kt
```diff
@@ -37,6 +38,12 @@ public object TrackPositionUiModelMapper {
         if (currentPositionMs == null || durationMs == null || durationMs <= 0) {
             return TrackPositionUiModel.Hidden
         }
+        event.playbackState.seekProjection?.let { seek ->
+            if (seek > Duration.ZERO) {
```
fstanis, line 42: I would maybe merge this with the line above: `event.playbackState.seekProjection?.takeIf { it > Duration.ZERO }?.let { seek -> ...`

media/ui/src/main/java/com/google/android/horologist/media/ui/state/model/TrackPositionUiModel.kt
```diff
@@ -26,6 +27,21 @@ public sealed class TrackPositionUiModel {
     public abstract val shouldAnimate: Boolean
     public abstract val isLoading: Boolean
 
+    public fun getCurrentPlaybackDuration(): Duration =
+        when (this) {
+            is Actual -> duration
+            is SeekProjection -> duration
+            is Predictive -> predictor.predictDuration(SystemClock.elapsedRealtime())
```
fstanis, line 34: This makes these methods harder to test, since they're relying on the global clock. I'd make this method take `elapsedRealtime: Long` as a parameter - but ideally, this logic shouldn't be in the model itself but in the call site.

laiyichin: Removed these get properties from models.

media/ui/src/main/java/com/google/android/horologist/media/ui/state/mapper/TrackPositionUiModelMapper.kt
laiyichin: Done.

PR #1728, opened 2023-10-04
Adds foreground check to DataLayerAppHelper methods

garanj:
#### WHAT

Adds a foreground check to methods provided by DataLayerAppHelper

#### WHY

Methods in DataLayerAppHelper can launch activities on the remote device. This should only be possible where the calling process is in the foreground.

#### HOW

Checks foreground status of process invoking the method

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


datalayer/core/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,19 @@
...
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<resources>
+    <string name="datalayerapphelper_foreground_error">This method can only be called from the foreground.</string>
```
yschimke, line 18: I'm not sure we want the exception translated. Maybe better being in english in log files.

yschimke (approved):

datalayer/core/src/main/res/values/strings.xml
garanj: Sure. Shall I move it straight out of the resource?

garanj: Fixed.

datalayer/phone/src/main/java/com/google/android/horologist/datalayer/phone/PhoneDataLayerAppHelper.kt
```diff
@@ -47,6 +48,7 @@ public class PhoneDataLayerAppHelper(
 
     @CheckResult
     override suspend fun startCompanion(node: String): AppHelperResultCode {
+        super.checkIsForegroundOrThrow()
```
yschimke, line 51: why `super.` prefix?

garanj: Fixed

yschimke:
Landing, doesn't seem blocked.

Issue #1748, opened 2023-10-08
Rotary input support for InlineSlider


cvb941:
I was attempting to use rotary input to control the InlineSlider value.

From the documentation [url](https://developer.android.com/training/wearables/compose/rotary-input#discrete-values), it seems like it should only be necessary to simply give focus to the Slider component and it would automatically receive rotary events.

However, this did not work and I had to manually setup `onRotaryScrollEvent()` and adjust the Slider value from there.

yschimke:
Yep - not really a Horologist issue, but I can see what you drew that conclusion from the docs.

I'll fix docs next week.

You probably want to use Horologist's onRotaryInputAccumulatedWithFocus, which handles the focus for you.

https://github.com/google/horologist/blob/a84dee88bb8aa0747206ba27aedf61689697c5c6/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/AccumulatedRotaryInputModifier.kt#L43

We could build a wrapper for InlineSlider that does all this for you.  Would that work for you?

cvb941:
Hello, thank you for your response and proactivity.

You're right, InlineSlider is a material component and does not have anything to do with Horologist. I think I was just reading about some other rotary input-related issues of Horologist and mixed up the two.

The `onRotaryInputAccumulatedWithFocus` modifier accumulates a Float which would somehow need to be tied with the slider value state. For that reason, I'm using the raw `onRotaryScrollEvent`.

I increment the slider value when triggered and ignore the scroll value. This feels fine for Galaxy Watch 4 and the Emulator, but may be unusable for other devices.

If there is any built in code support for this, I think it should be part of the Wear material library itself. It is up to you though to judge that.

I think in an ideal scenario, the current InlineSlider should handle rotary input automatically when it gets focused.

yschimke:
PR above is what it would look like for Stepper.  Do you need InlineSlider particularly? How are you using it?

cvb941:
![image](https://github.com/google/horologist/assets/2351527/81a312fa-5f32-4157-bf13-336c1c946b74)

I have an InlineSlider acting as a difficulty selector.

I alone wouldn't bother adding the rotary input to it, but my app kept getting rejected at Play Store reviews due to missing rotary input, even though it was present on every scrollable screen. I then tried adding rotary to this screen and the review issue has disappeared, who knows if that was the problem though.
It would be great to have some quick feedback from the reviewers with a screenshot of the problem that they have encountered.

[...]

PR #1749, opened 2023-10-09
Support for Steppers with RSB built in

yschimke:
#### WHAT

Support for Steppers with RSB built in

#### WHY

Feedback from https://github.com/google/horologist/issues/1748

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
nice!

PR #1751, opened 2023-10-12
Fix issue when rotary modifiers were stopping receiving new rotary events

Kpeved:
#### WHAT
Issue with not recomposed rotary modifiers were fixed.

#### WHY
Specifically that was causing an issue with AlarmPicker b/301894733.
Problem was that rotary modifiers were not recomposed when state was changed, so they continue to fire events even though the new modifiers were already created. 

#### HOW
Was fixed by updating Rotary modifier to the latest version from GOB.
https://team-review.git.corp.google.com/c/wear-developer-eng/wear-compose/+/2005597
Some Rotary apis also were removed in favour of more simplified main apis.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
```diff
@@ -323,57 +364,44 @@ public object RotaryDefaults {
      */
     @ExperimentalHorologistApi
     @Composable
-    public fun rememberSnapHandler(
+    internal fun rememberSnapHandler(
```
luizgrp, line 367: nit: experimental annotation could be removed given visibility was changed to internal

same to the other changes in this file in this PR

luizgrp (approved):
LGTM

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Rotary.kt
Kpeved: Done

Issue #1750, opened 2023-10-11
rotaryWithScroll library doesn't respect use crown haptic settings


blakadder:
I am directed by the creators of the Home Assistant Companion app to file a bug report;

On TicWatch Pro 5 with OS 3.5 the app doesn't respect the "Use crown" vibration system settings and keeps vibration on crown use. Furthermore the type of vibrations are not the usual ones used when Use crown in enabled bit instead the strong vibration that's used for touch events is used

yschimke:
Thanks for the report. We can add specific haptics but we would need to know which ones are preferred. I'll see if it's doing something stupid and using the ones for a different model.

Is there documentation for this setting? Or it's a standard wear one?

jpelgrom:
I can confirm the same 'issue' exists on a Samsung Galaxy Watch 4 with the touch bezel (= rotary input on this device). In the system settings app > General > Touch bezel, there is an option "Vibration feedback". Sounds like it is kind of standard Wear but depends on the device what it's called and where it is located.

Turning it off doesn't change the behavior of apps using the `rotaryWithScroll` modifier, while other apps that (presumably) use a classic view do change and stop vibrating when rotating. Easy to spot as thin vs thick scroll bar.

blakadder:
> Thanks for the report. We can add specific haptics but we would need to know which ones are preferred. I'll see if it's doing something stupid and using the ones for a different model.
> 
> Is there documentation for this setting? Or it's a standard wear one?

Specific or not the existing one is overly aggressive and should be toned down.

The setting also exists on the Pixel Watch as "Using crown" - https://www.verizon.com/support/knowledge-base-303381/

yschimke:
So I think we need to do one of two things

1) Find the relevant haptics code for this device, or hopefully class of devices.
2) Fallback to reading the setting, possibly `Settings.Global.getInt(context.contentResolver, "wear_vibrate_for_rotary_input")`

It's possible neither are public APIs.

A quick test of 1, are you able to add this in and test on that device?

```
public class LikePixelWatchRotaryHapticFeedback(private val view: View) : RotaryHapticFeedback {

    @ExperimentalHorologistApi
    override fun performHapticFeedback(
        type: RotaryHapticsType,
    ) {
        when (type) {
            RotaryHapticsType.ScrollItemFocus -> {
                view.performHapticFeedback(
                    if (Build.VERSION.SDK_INT >= 33) {
                        ROTARY_SCROLL_ITEM_FOCUS
                    } else {
[...]

PR #1752, opened 2023-10-16
Use Wear 3/4 haptics by default

yschimke:
#### WHAT

Use Wear 3/4 haptics by default. Assume OEMs haven't customised these. 

#### WHY

We should have nice haptics on all devices, not just PixelWatch and Galaxy Watch.
To fix https://github.com/google/horologist/issues/1750

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
@@ -282,15 +282,39 @@ public class DefaultRotaryHapticFeedback(private val view: View) : RotaryHapticF
     ) {
         when (type) {
             RotaryHapticsType.ScrollItemFocus -> {
-                view.performHapticFeedback(HapticFeedbackConstants.LONG_PRESS)
+                view.performHapticFeedback(
+                    if (Build.VERSION.SDK_INT >= 33) {
+                        PixelWatchRotaryHapticFeedback.ROTARY_SCROLL_ITEM_FOCUS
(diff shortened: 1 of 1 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
```diff
@@ -282,15 +282,39 @@ public class DefaultRotaryHapticFeedback(private val view: View) : RotaryHapticF
...
+                    },
+                )
             }
 
             RotaryHapticsType.ScrollLimit -> {
-                view.performHapticFeedback(HapticFeedbackConstants.LONG_PRESS)
+                view.performHapticFeedback(
+                    if (Build.VERSION.SDK_INT >= 33) {
+                        PixelWatchRotaryHapticFeedback.ROTARY_SCROLL_LIMIT
+                    } else if (Build.VERSION.SDK_INT >= 30) {
```
Kpeved, line 312: We can't filter Wear 3.5 with that check - as both Wear 3 and 3.5 have the same SDK. But these constants were added only in Wear 3.5

Kpeved: I used this setting to check that, but couldn't test it as I don't have Wear 3.5 device 
https://github.com/google/horologist/commit/40a7940e0b142447e862f417705e69037046a324#diff-99fe5c62c28ba9bce84ee988b61b5dd53e74f53958c7d7032f55812462e8c4c4R408

PR #1757, opened 2023-10-17
Fix TrackPositionUiModelMapper for SeekProjection at 0 position

laiyichin:
#### WHAT
Fix TrackPositionUiModelMapper for SeekProjection to work for seek to position 0

#### WHY
Currently, when seeking at to the beginning and seek duration is set to 0, the mapper would map back to predictive model instead of seek projected model. 

#### HOW
add the `=` to accommodate seeking to position 0

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke (approved):
Consider adding a test

laiyichin:
> Consider adding a test

Done.

PR #1776, opened 2023-10-27
Seeking

laiyichin:
#### WHAT
Add seek in current media item enabled signal to the ui state

#### WHY
This allows as to add one more signals to the UI with the seeking in current media item command (such as enabling or disabling the seek button)

#### Checklist :clipboard:
- [] Add explicit visibility modifier and explicit return types for public declarations
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
LGTM but @fstanis maybe a better reviewer

PR #1782, opened 2023-10-31
Add ScalingLazyColumnDefaults.responsive() and make it the default

yschimke:
#### WHAT

Added `ScalingLazyColumnDefaults.responsive()` which has a 5% horizontal margin. 

image

#### WHY

Mobvoi TicWatch Pro 5 is 233dp, other new devices are similar.  Solve this without user intervention.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnDefaults.kt
```diff
@@ -128,4 +131,76 @@ public object ScalingLazyColumnDefaults {
...
+            override fun create(): ScalingLazyColumnState {
+                val density = LocalDensity.current
+                val configuration = LocalConfiguration.current
+                val screenWidthDp = configuration.screenWidthDp.toFloat()
+                val screenHeightDp = configuration.screenHeightDp.toFloat()
+                val padding = screenWidthDp * horizontalPaddingPercent
+                val topPaddingDp: Dp = if (firstItemIsFullWidth) {
+                    calculateVerticalOffsetForChip(screenWidthDp, horizontalPaddingPercent)
+                } else {
+                    32.dp
```
ssancho14, line 172: Should this be percentage based?

ssancho14 (approved):

compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnDefaults.kt
```diff
@@ -128,4 +131,76 @@ public object ScalingLazyColumnDefaults {
...
+
+    @ExperimentalHorologistApi
+    public fun responsive(
+        firstItemIsFullWidth: Boolean = true,
+        verticalArrangement: Arrangement.Vertical =
+            Arrangement.spacedBy(
+                space = 4.dp,
+                alignment = Alignment.Top,
+            ),
+        horizontalPaddingPercent: Float = 0.052f,
```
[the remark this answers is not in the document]
yschimke: Why? Do we have other examples? How strongly do you feel about it?

compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnDefaults.kt
yschimke: Maybe, the current intent is putting it just below the time text.  But maybe it should also be positioned so it isn't scaled.  What percent would you configure for a button?

yschimke: Let's land and follow up. So you don't have to solve it now.

PR #1791, opened 2023-11-06
Change responsive to work better on Square devices

yschimke:
#### WHAT

Change responsive to work better on Square devices

#### WHY

No point having a larger top margin on full width square devices

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
@luizgrp Can you post review?

PR #1793, opened 2023-11-06
Change Default back to belowTimeText

yschimke:
#### WHAT

Change default back from responsive to belowTimeText.  The horologist screens known to be safe are overriding to keep using responsive.

#### WHY

Early iteration on responsive, not ready as default.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
Based on it being a RNG bug, going to close this.

PR #1801, opened 2023-11-10
Testing mirror

yschimke:
#### WHAT
| Direction | Left    | Right    |
| :---:   | :---: | :---: |
|  | AutoMirror   | AutoMirror + rtlMode=Mirrored   |
| RTL | AutoMirror   | AutoMirror + rtlMode=Mirrored   |
|  | Filled   | AutoMirror   |
| RTL | Filled   | AutoMirror   |

image

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/sample/build.gradle.kts
+++ b/sample/build.gradle.kts
@@ -154,6 +154,7 @@ dependencies {
     implementation(projects.datalayer.core)
     implementation(projects.datalayer.watch)
     implementation(projects.logo)
+    implementation(libs.accompanist.testharness)
 
     implementation(libs.compose.ui.util)
 
--- a/sample/src/main/java/com/google/android/horologist/scratch/ScratchPreview.kt
+++ b/sample/src/main/java/com/google/android/horologist/scratch/ScratchPreview.kt
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2023 The Android Open Source Project
+ *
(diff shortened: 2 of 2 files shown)


yschimke:
@luizgrp I suspect we should consider whether we need rtlMode any more.  Or use it only in extreme cases?

luizgrp:
I think the rtlMode param might still be useful when using an Icon that is not "AutoMirrored" or when calling Icon with a [drawable](https://github.com/google/horologist/blob/879118c2d83f1016261c2cdb0b70b8d4282ed007/compose-material/src/main/java/com/google/android/horologist/compose/material/Icon.kt#L66)?

If you agree then I think we should just update our code to not pass rtlMode when using AutoMirrored Icons, and maybe update the kdocs to:

```kotlin
/**
 * This component is an alternative to [Icon], providing the following:
 * - a convenient way of setting the icon to be mirrored in RTL mode;
 * 
 * Leave [rtlMode] set to [IconRtlMode.Default] when the Icon used is "auto mirrored", 
 * e.g. from the [Icons.AutoMirrored] set, or with `android:autoMirrored="true"` set in 
 * the XML vector file.
 */
```

yschimke:
Maybe warn/fail if autoMirror is true? It's available in the api.

luizgrp:
sounds good!

PR #1815, opened 2023-11-15
Add default icon size to custom action media button aligning to the media button

laiyichin:
#### WHAT
Add default icon size for custom action media button

#### WHY
Currently the custom action sizes do not match the previous / next media control buttons

#### HOW
Add default icon size and alignment matching the media button (`media/ui/src/main/java/com/google/android/horologist/media/ui/components/controls/MediaButton.kt`)

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media/ui/src/main/java/com/google/android/horologist/media/ui/components/CustomActionMediaButton.kt
+++ b/media/ui/src/main/java/com/google/android/horologist/media/ui/components/CustomActionMediaButton.kt
@@ -16,10 +16,13 @@
 
 package com.google.android.horologist.media.ui.components
 
+import androidx.compose.foundation.layout.offset
 import androidx.compose.foundation.layout.size
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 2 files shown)


yschimke (approved):

media/ui/src/main/java/com/google/android/horologist/media/ui/components/CustomActionMediaButton.kt
```diff
@@ -52,6 +57,22 @@ public fun CustomActionMediaButton(
             model = iconUri,
             contentDescription = contentDescription,
             colorFilter = ColorFilter.tint(colors.contentColor(enabled = enabled).value),
+            modifier = Modifier
```
yschimke, line 60: Should any of these be backported to 0.4.x? Probably not, but if they are bugs, please flag.

PR #1816, opened 2023-11-15
Volume color

laiyichin:
#### WHAT
Make color configurable in volume position indicator

#### HOW
Adding an optional volume parameter

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media/audio-ui/api/current.api
+++ b/media/audio-ui/api/current.api
@@ -8,7 +8,7 @@ package com.google.android.horologist.audio.ui {
   }
 
   public final class VolumePositionIndicatorKt {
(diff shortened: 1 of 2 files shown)


media/ui/src/main/java/com/google/android/horologist/media/ui/components/CustomActionMediaButton.kt
```diff
@@ -52,6 +57,22 @@ public fun CustomActionMediaButton(
             model = iconUri,
             contentDescription = contentDescription,
             colorFilter = ColorFilter.tint(colors.contentColor(enabled = enabled).value),
+            modifier = Modifier
```
yschimke, line 60: Dupe from https://github.com/google/horologist/pull/1815/files

laiyichin: Synced.

yschimke (approved):

PR #1818, opened 2023-11-16
Add sample for install app prompt

luizgrp:
#### WHAT

Add sample for install app prompt.

| Screenshot_20231116_150458 | Screenshot_20231116_150506 | Screenshot_20231116_150515 | 
|--------|--------|--------|

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
+++ b/datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
@@ -16,6 +16,7 @@
 
 package com.google.android.horologist.datalayer.sample
 
+import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
@@ -28,8 +29,13 @@ import androidx.compose.material3.Card
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
@@ -40,6 +46,7 @@ import com.google.android.horologist.data.complicationInfo
 import com.google.android.horologist.data.surfacesInfo
 import com.google.android.horologist.data.tileInfo
 import com.google.android.horologist.data.usageInfo
+import com.google.android.horologist.datalayer.sample.prompt.InstallAppDialog
 import com.google.android.horologist.datalayer.sample.ui.theme.HorologistTheme
 import com.google.protobuf.Timestamp
 
@@ -49,7 +56,10 @@ fun AppHelperNodeStatusCard(
     onInstallClick: (String) -> Unit,
     onLaunchClick: (String) -> Unit,
     onCompanionClick: (String) -> Unit,
+    onInstallAppClick: (packageName: String) -> Unit,
 ) {
+    var showInstallAppDialog by remember { mutableStateOf(false) }
+
     Box(
         modifier = Modifier
             .padding(16.dp)
@@ -129,8 +139,46 @@ fun AppHelperNodeStatusCard(
                         Text(stringResource(id = R.string.app_helper_companion_button_label))
                     }
                 }
+                /*
+                This button should only be displayed when AppHelperNodeStatus.isAppInstalled returns
+                false. However, given our sample app is not published on Play, we are using Gmail as
+                a "sample" in order to display the full flow.
+                 */
+                Button(
+                    modifier = Modifier.wrapContentHeight(),
+                    onClick = {
+                        showInstallAppDialog = true
+                    },
+                ) {
+                    Text(stringResource(id = R.string.app_helper_install_app_prompt_button_label))
+                }
             }
         }
+    }
+
+    if (showInstallAppDialog) {
+        val sampleAppPackage =
+            stringResource(id = R.string.app_helper_install_app_prompt_sample_app_package)
+
+        InstallAppDialog(
+            appName = stringResource(id = R.string.app_helper_install_app_prompt_sample_app_name),
+            watchName = nodeStatus.displayName,
+            message = stringResource(id = R.string.app_helper_install_app_prompt_sample_message),
+            icon = {
+                Image(
+                    painter = painterResource(id = R.drawable.sample_app_wearos_screenshot),
+                    contentDescription = null,
+                )
+            },
+            onDismissRequest = {
+                showInstallAppDialog = false
+            },
+            onConfirmation = {
+                showInstallAppDialog = false
+
+                onInstallAppClick(sampleAppPackage)
+            },
+        )
     }
 }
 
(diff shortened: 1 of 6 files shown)


datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
```diff
@@ -129,9 +145,72 @@ fun AppHelperNodeStatusCard(
...
+                }
             }
         }
     }
+
+    if (showInstallAppPromptDialog.value) {
+        val context = LocalContext.current
+
+        InstallAppPromptDialog(
+            appName = "Gmail",
```
kul3r4, line 171: Can we move this and the message string into constants? to make it easier also to localize for apps?

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
```diff
@@ -129,9 +145,72 @@ fun AppHelperNodeStatusCard(
...
+                    contentDescription = null,
+                )
+            },
+            onDismissRequest = {
+                showInstallAppPromptDialog.value = false
+            },
+            onConfirmation = {
+                showInstallAppPromptDialog.value = false
+
+                val appPackageName = "com.google.android.gm"
```
kul3r4, line 186: can we move constant strings in one place?

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
luizgrp: Done 👍  The app and package names are marked as not translatable.

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/prompt/InstallAppDialog.kt
```diff
@@ -0,0 +1,96 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import com.google.android.horologist.datalayer.sample.R
+
+@Composable
+fun InstallAppDialog(
```
yschimke, line 37: Would this be actively usable if in library form? Or it's really just a UX sample?

yschimke (commented):
Looks great.

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
```diff
@@ -129,9 +147,67 @@ fun AppHelperNodeStatusCard(
...
         }
     }
+
+    if (showInstallAppDialog) {
+        val context = LocalContext.current
+        val sampleAppPackage = stringResource(id = R.string.app_helper_install_app_prompt_sample_app_package)
+
+        InstallAppDialog(
+            appName = stringResource(id = R.string.app_helper_install_app_prompt_sample_app_name),
+            watchName = nodeStatus.displayName,
```
yschimke, line 175: The current implementation, isn't really specific to a watch.  Would a more realistic example be to have this outside the `AppHelperNodeStatusCard` and enabled based on the presence of a watch without the app?

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
```diff
@@ -129,9 +147,67 @@ fun AppHelperNodeStatusCard(
                         Text(stringResource(id = R.string.app_helper_companion_button_label))
                     }
                 }
+                Row(
+                    modifier = Modifier
+                        .padding(8.dp)
+                        .fillMaxWidth(),
+                    horizontalArrangement = Arrangement.SpaceBetween,
+                ) {
+                    Button(
```
yschimke, line 156: Should this be enabled based on whether the app is already installed? I think that is contained in `AppHelperNodeStatus`

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
```diff
@@ -129,9 +147,67 @@ fun AppHelperNodeStatusCard(
...
+                try {
+                    startActivity(
+                        context,
+                        Intent(
+                            Intent.ACTION_VIEW,
+                            Uri.parse(PLAY_STORE_APP_URI_PREFIX + sampleAppPackage),
+                        ),
+                        Bundle(),
+                    )
+                } catch (anfe: ActivityNotFoundException) {
```
yschimke, line 198: Worth a comment for what case this is handling? I assume it will handle it via Chrome instead of Play. But does that allow users to click to install then?

Is that actually possible, given the mobile must have Play Services in order to have a paired Wear device?

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
```diff
@@ -129,9 +147,67 @@ fun AppHelperNodeStatusCard(
...
+                    contentDescription = null,
+                )
+            },
+            onDismissRequest = {
+                showInstallAppDialog = false
+            },
+            onConfirmation = {
+                showInstallAppDialog = false
+
+                try {
```
yschimke, line 189: Is it worth moving this play launch code to a helper function?

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/prompt/InstallAppDialog.kt
luizgrp: Just ux sample for these PRs

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
luizgrp: good point!

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
luizgrp: > I assume it will handle it via Chrome instead of Play. 

Yeah it would launch the default browser or the dialog to choose the app to open it with.

> But does that allow users to click to install then?

If the user is logged in, yes.

> Is that actually possible, given the mobile must have Play Services in order to have a paired Wear device?

Good point, in that case this would only happen if the version of the Google Play installed, for whatever reason, does not handle that URI. It seem extra cautious, I will remove the code.

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
luizgrp: sure, and when we do, then we have to keep the code that I said in my previous comment that I would remove 😂

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/AppHelperNodeStatusCard.kt
luizgrp: Can't be done with what we have available - see https://github.com/google/horologist/pull/1827

PR #1826, opened 2023-11-21
Fix bottom padding for responsive() SLC

yschimke:
#### WHAT

Fix bottom padding for SLC.

#### WHY

Bottom item if larger than a Chip, such as a Card will be obscured.

AutoCenteringParams doesn't allow adjusting the bottom item, and with Anchor = Start will only ensure the start of the final item can get to the center, which is not enough.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


Kpeved:
Can you please add more details what exactly this change fixes?

compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnDefaults.kt
```diff
@@ -198,11 +203,12 @@ public object ScalingLazyColumnDefaults {
                             index = 0,
                             offsetPx = topScreenOffsetPx,
                         ),
+                        autoCentering = null,
```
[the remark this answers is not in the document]
yschimke: AutoCentering API is this

```
public class AutoCenteringParams(
    // @IntRange(from = 0)
    internal val itemIndex: Int = 1,
    internal val itemOffset: Int = 0,
)
```

You can ensure that the top is correctly working. However it doesn't ensure that the bottom can be completely scrolled into view.

So better to disable it, and do it manually with enough content padding at the bottom to allow the last item to be unobscured.

sample/src/test/snapshots/images/com.google.android.horologist.screensizes_ScalingLazyColumnDefaultsTest_belowTimeText_end[0]_mobvoiticwatchpro5.png
Comment (yschimke) [inline sample/src/test/snapshots/images/com.google.android.horologist.screensizes_ScalingLazyColumnDefaultsTest_belowTimeText_end[0]_mobvoiticwatchpro5.png:1]: @Kpeved without this change, the `responsive` screens hit this issue.

With the change, they can scroll so the last item is mostly not cutoff.

sample/src/test/snapshots/images/com.google.android.horologist.screensizes_ScalingLazyColumnDefaultsTest_responsive_end[6]_smalldevicebigfonts.png
Comment (yschimke) [inline sample/src/test/snapshots/images/com.google.android.horologist.screensizes_ScalingLazyColumnDefaultsTest_responsive_end[6]_smalldevicebigfonts.png:1]: This seems like a bug...

sample/src/test/snapshots/images/com.google.android.horologist.screensizes_MediaPlayerLibraryTest_screenshot[2]_samsunggalaxywatch6large.png
Comment (yschimke) [inline sample/src/test/snapshots/images/com.google.android.horologist.screensizes_MediaPlayerLibraryTest_screenshot[2]_samsunggalaxywatch6large.png:1]: This is a test bug, I don't have the code to change density in robolectric right.

yschimke:
cc @garanj

Kpeved (approved):
LGTM

PR #1827, opened 2023-11-21
Improve AppHelperNodeStatus

luizgrp:
#### WHAT

Improve `AppHelperNodeStatus` to make `nodeType` attribute available only when the app is installed on the node.

#### WHY

Currently the code in place can't check the type of the node without having the app installed, and `nodeType` in `AppHelperNodeStatus` might lead one to think that it can.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
My brain works the other way, but no strong objection and you've talked it though with Garan. LGTM

Other way being the existing way.

```
    val isAppInstalled: Boolean,
    val nodeType: AppHelperNodeType,
```

garanj:
This definitely makes it clear, thanks @luizgrp

garanj (approved):

PR #1831, opened 2023-11-23
Change datalayer sample app to use available API

luizgrp:
#### WHAT

Change datalayer sample app to use available API.

#### WHY

Showcase the available API + convenience.

#### HOW

Change sample to use API from `PhoneDataLayerAppHelper` instead of `WearableApiAvailability`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/MainActivity.kt
+++ b/datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/MainActivity.kt
@@ -49,7 +49,6 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.lifecycle.lifecycleScope
 import com.google.android.horologist.data.ProtoDataStoreHelper.protoDataStore
 import com.google.android.horologist.data.WearDataLayerRegistry
-import com.google.android.horologist.data.WearableApiAvailability
 import com.google.android.horologist.data.apphelper.AppHelperNodeStatus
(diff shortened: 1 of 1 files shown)


datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/MainActivity.kt
```diff
@@ -96,7 +95,7 @@ class MainActivity : ComponentActivity() {
                     var apiAvailable by remember { mutableStateOf(false) }
                     LaunchedEffect(Unit) {
                         coroutineScope.launch {
-                            apiAvailable = WearableApiAvailability.isAvailable(registry.dataClient)
+                            apiAvailable = phoneDataLayerAppHelper.isAvailable()
```
luizgrp, line 98: @garanj @yschimke `PhoneDataLayerAppHelper` checks the availability of `CapabilityClient`, whereas the current code is checking specifically the availability of `DataClient`. 

Is this a suitable replacement or is the intention to explicitly check for the availability of `DataClient`?

garanj: I'm not sure what the original intention was, but it seems reasonable to me. Though I suppose the DataClient is required for correct operation?

Out of interest, is there ever a case where one client could exist without another?

luizgrp:
done in https://github.com/google/horologist/pull/1872

PR #1836, opened 2023-11-27
Update media player screenshots

yschimke:
#### WHAT

Update media player screenshots

#### WHY

Show the Marquee text which has an offset and fade.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
good idea!

Issue #1671, opened 2023-09-11
Add iconRtlMode param to Chip


luizgrp:
This surfaces that these components were not correct in RTL mode before. We should add `iconRtlMode` param to `Chip` component, similarly to what we have for others like [CompactChip](https://github.com/google/horologist/blob/180d15950566236757901e06e2c555fb59c59dd2/compose-material/src/main/java/com/google/android/horologist/compose/material/CompactChip.kt#L58)

_Originally posted by @luizgrp in https://github.com/google/horologist/pull/1665#discussion_r1321961910_

MohitMandalia:
I can pick this up

MohitMandalia:
Sure I'll check the comments for context and ask if I need some clarification, Thanks

oas004:
Are you still working on this @MohitMandalia? Otherwise, I can also check this out :)

oas004:
@luizgrp Is it correct that we should add `iconRtlMode: IconRtlMode = IconRtlMode.Default` to the `Chip` and `OutlinedChip` component and then move the implementation to use `com.google.android.horologist.compose.material.Icon` wrapper instead? :)  Or am I missing something here?

PR #1850, opened 2023-12-04
Add iconRtlMode param to Chip API

oas004:
#### WHAT
Add iconRtlMode param to Chip API and move chip to use the internal horologist material icon implementation.

#### WHY
#1671 

#### HOW
Add iconRtlModeParam to chip. Then start using it in the icon implementation. Then add preview and Screenshot test.

image

Plan is to do the same with OutlinedChip in a separate PR :) 

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


compose-material/src/debug/java/com/google/android/horologist/compose/material/ChipPreview.kt
```diff
@@ -178,6 +182,24 @@ fun ChipPreviewWithImageBackgroundChipColors() {
     )
 }
 
+@Preview(
+    name = "With icon and forced rtl mode",
+    backgroundColor = 0xff000000,
+    showBackground = true,
+)
+@Composable
+private fun ChipPreviewWithIconMirroredMode() {
```
oas004, line 191: @luizgrp is this sufficient for previews? Or is there other permutations that I should add for as well? :)

compose-material/src/test/java/com/google/android/horologist/compose/material/ChipTest.kt
```diff
@@ -353,6 +355,21 @@ class ChipTest : ScreenshotBaseTest() {
         }
     }
 
+    @Test
+    fun withIconMirrored() {
```
oas004, line 359: @luizgrp Is this sufficient for screenshot tests? Or should I add tests for other permutations as well? :)

compose-material/src/debug/java/com/google/android/horologist/compose/material/ChipPreview.kt
luizgrp: I'd say these forced RTL should be only added in the tests and not in previews - I know we've added them in the past and might still have some lingering around, sorry if that caused confusion!

compose-material/src/test/java/com/google/android/horologist/compose/material/ChipTest.kt
luizgrp: We have the other permutations being covered by the Icon component tests, I'd say this one is sufficient 👍

compose-material/src/test/java/com/google/android/horologist/compose/material/ChipTest.kt
```diff
@@ -353,6 +355,21 @@ class ChipTest : ScreenshotBaseTest() {
...
+        screenshotTestRule.setContent(isComponent = true, takeScreenshot = true) {
+            CompositionLocalProvider(LocalLayoutDirection provides LayoutDirection.Rtl) {
+                Chip(
+                    label = "Primary label",
+                    onClick = { },
+                    secondaryLabel = "Secondary label",
+                    icon = Icons.AutoMirrored.Outlined.VolumeDown,
+                    iconRtlMode = IconRtlMode.Mirrored,
```
luizgrp, line 367: `iconRtlMode = Mirrored` should not be used with icons that are auto mirrored already, like these ones from `Icons.AutoMirrored.*` and XML vector files with `android:autoMirrored="true"`. I thought we had this info in the kdocs of the components, but can't find it in any 🤔 sorry!

So we should use `Icons.Outlined.VolumeDown` for this test and for the preview added in this PR.

compose-material/src/test/snapshots/images/com.google.android.horologist.compose.material_ChipTest_withIconMirrored.png
luizgrp, line 1: This screenshot shows that the volume icon is not pointing to the left, as it should, because it's being mirrored "twice".

compose-material/src/test/java/com/google/android/horologist/compose/material/ChipTest.kt
oas004: oh Okay, that makes sense! :) I can change it in the test to use `Icons.Outlined.VolumeDown` :) Thanks for pointing that out! :)

compose-material/src/test/snapshots/images/com.google.android.horologist.compose.material_ChipTest_withIconMirrored.png
oas004: Changed the icon to not be the automirrored version in [dd82ea2](https://github.com/google/horologist/pull/1850/commits/dd82ea2b2d847dd9dcd7cfc71304e4f3e7e795b9) and I think that fixed it :)

Issue #1808, opened 2023-11-13
Create separated sample apps for datalayer


luizgrp:
### WHAT

Add a phone and wear sample app to showcase the datalayer features.

### WHY

1. Currently it is not intuitive to find the datalayer sample, as it is in the "auth" sample.
2. Unify the datalayer samples, as currently, they are in two places: the root's sample (Data Layer Nodes menu) and the "auth" sample.

### HOW

- [x] Add new wear project to folder datalayer/sample/wear.
- [x] Add new phone project to folder datalayer/sample/phone.
- [x] Add new library project to folder sample/shared.
- [x] Move datalayer sample code from auth/sample/shared project to datalayer/sample/shared project.
- [x] Move datalayer sample code from auth/sample/phone project to datalayer/sample/phone project.
- [x] Move datalayer sample code from auth/sample/wear project to datalayer/sample/wear project.
- [ ] Move datalayer sample code from sample project to datalayer/sample/wear project.

PR #1852, opened 2023-12-05
Move datalayer sample to datalayer sample app

luizgrp:
#### WHAT

Move datalayer sample to datalayer sample app.

#### WHY

Fixes https://github.com/google/horologist/issues/1808

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


datalayer/sample/wear/build.gradle.kts
```diff
@@ -87,6 +90,42 @@ android {
...
+                "build/generated/source/proto/debug/kotlin",
+                "build/generated/source/proto/debug/grpckt",
+            )
+        }
+    }
+}
+
+protobuf {
+    protoc {
+        artifact = "com.google.protobuf:protoc:3.25.1"
```
yschimke, line 108: For later - we should grab these from the version catalog also.

yschimke (approved):

PR #1858, opened 2023-12-06
Add title secondary to sample application

### LINKED ISSUE #1347: List components from compose-material to sample app
Add all the variants of each component in `compose-material` to the sample app.

oas004:
#### WHAT
Add title secondary variants to the sample application

#### WHY
#1347 

This should resolves #1347 unless any objections. I think I have listed all the material components and variants in the sample app.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/sample/src/main/java/com/google/android/horologist/materialcomponents/SampleTitleScreen.kt
+++ b/sample/src/main/java/com/google/android/horologist/materialcomponents/SampleTitleScreen.kt
@@ -17,10 +17,14 @@
 package com.google.android.horologist.materialcomponents
 
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Add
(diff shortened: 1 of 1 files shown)


sample/src/main/java/com/google/android/horologist/materialcomponents/SampleTitleScreen.kt
```diff
@@ -33,7 +38,24 @@ internal fun SampleTitleScreen(
...
         item {
-            Title("Title")
+            Title(
+                modifier = Modifier.fillMaxWidth(),
+                text = "Title",
+            )
+        }
+        item {
+            SecondaryTitle(
+                modifier = Modifier.fillMaxWidth(),
```
oas004, line 48: Don't think I need to fillMaxWidth as they do behind the component implementation

luizgrp (approved):

PR #1864, opened 2023-12-12
Adds icon to MediaUiModel and display components

fstanis:
#### WHAT
#1863 follow-up, adds the icon to `MediaUiModel`.

#### WHY
Makes it possible for apps to add an icon next to the media title.

#### HOW
Adds `titleIcon` optional property to `MediaUiModel` and consumes it in `MarqueeTextMediaDisplay` and `TextMediaDisplay`.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests

### DIFF
--- a/media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/MarqueeTextMediaDisplay.kt
+++ b/media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/MarqueeTextMediaDisplay.kt
@@ -25,12 +25,15 @@ import androidx.compose.animation.fadeOut
 import androidx.compose.animation.slideInHorizontally
 import androidx.compose.animation.slideOutHorizontally
 import androidx.compose.animation.togetherWith
+import androidx.compose.foundation.Image
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.ColorFilter
+import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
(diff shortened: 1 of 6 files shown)


media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/MarqueeTextMediaDisplay.kt
```diff
@@ -49,6 +53,7 @@ public fun MarqueeTextMediaDisplay(
     modifier: Modifier = Modifier,
     title: String? = null,
     artist: String? = null,
+    titleIcon: ImageVector? = null,
```
yschimke, line 56: Is an AppIcon actually going to be an ImageVector? Or should be left as a slot or made flexible? Can apps choose to use album artwork via Coil?

yschimke (approved):
LGTM - but a question if you are sure about ImageVector

media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/MarqueeTextMediaDisplay.kt
fstanis: Changed to `Paintable?`.

media/ui/src/main/java/com/google/android/horologist/media/ui/state/model/MediaUiModel.kt
```diff
@@ -26,6 +27,7 @@ public data class MediaUiModel(
     val subtitle: String = "",
     val artworkUri: String? = null,
     val artworkColor: Color? = null,
+    val titleIcon: Paintable? = null,
```
yschimke, line 30: Where does this get used? I'm curious if we need both this and artworkUri/artworkColor?

yschimke (approved):

media/ui/src/main/java/com/google/android/horologist/media/ui/state/model/MediaUiModel.kt
fstanis: This is meant to be a small image / icon that can be displayed next to the media title, inside the marquee. The artwork one is usually the album artwork, which is larger.

Ideally, I'd make the artwork one a `Paintable` or maybe even a type that can return just the color (`Colorable`?).

PR #1868, opened 2023-12-13
Adds Paintable class to represent generic image

fstanis:
#### WHAT
Adds a new `Paintable` interface that can be used to represent "any kind of image".

#### WHY
This decouples most parts of the library from `coil` while making it trivial to still use it (see `CoilPaintable`) and removes the need to use `Any` across the codebase.

#### HOW
By replacing uses of `Any` with `Paintable` in the codebase, which itself has only one method, which returns a `Painter` that can be used inside both `Icon` and `Image` elements. This logic previously lived in individual components (see `Chip` delta) and is now encapsulated in the various implementations of this interface.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


auth/composables/src/main/java/com/google/android/horologist/auth/composables/chips/AccountChip.kt
```diff
@@ -89,9 +90,8 @@ public fun AccountChip(
             label = email,
             onClick = onClick,
             modifier = modifier,
-            icon = avatar ?: defaultAvatar,
+            icon = CoilPaintable(avatar ?: defaultAvatar, placeholder),
```
luizgrp, line 93: nice, this is an improvement for the scenarios when the placeholder doesn't make sense

compose-material/src/main/java/com/google/android/horologist/compose/material/Paintable.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    @Composable
+    override fun rememberPainter(): Painter = painterResource(id = id)
+}
+
+@JvmInline
+public value class BitmapPaintable(private val bitmap: Bitmap) : Paintable {
+
+    @Composable
+    override fun rememberPainter(): Painter = remember { BitmapPainter(bitmap.asImageBitmap()) }
```
luizgrp, line 59: should the bitmap be used as key for remember? asImageBitmap doesn't seem to be costly though

compose-material/src/main/java/com/google/android/horologist/compose/material/Paintable.kt
```diff
@@ -0,0 +1,70 @@
...
+}
+
+@Stable
+public class CoilPaintable(private val model: Any?, private val placeholder: Painter? = null) : Paintable {
```
luizgrp, line 63: do we want to have "Coil" in the name, exposing the implementation details? ImageLoadingPaintable would be my suggestion but doesn't read great..

yschimke (approved):
I'm a bit concerned we are inventing an alternative Slot API....

media/ui/src/debug/java/com/google/android/horologist/media/ui/components/actions/ShowPlaylistChipPreview.kt
```diff
@@ -21,20 +21,23 @@ import androidx.compose.material.icons.automirrored.filled.FeaturedPlayList
...
 import com.google.android.horologist.compose.tools.WearPreview
 
 @WearPreview
 @Composable
 fun ShowPlaylistChipPreview() {
     ShowPlaylistChip(
-        artworkUri = "artworkUri",
+        artworkPaintable = CoilPaintable(
+            "artworkUri",
+            rememberVectorPainter(
```
luizgrp, line 34: should the placeholder also be a Paintable instance?

```kotlin
        artworkPaintable = CoilPaintable(
            model = "artworkUri",
            placeholder = ImageVectorPaintable(
                image = Icons.AutoMirrored.Default.FeaturedPlayList,
                tintColor = Color.Green,
            ).rememberPainter(),
        ),
```

media/ui/src/main/java/com/google/android/horologist/media/ui/screens/browse/PlaylistDownloadBrowseScreen.kt
```diff
@@ -106,8 +107,7 @@ public fun PlaylistDownloadBrowseScreen(
                                 id = R.string.horologist_browse_downloads_progress,
                                 download.percentage,
                             ),
-                            icon = Icons.Default.Downloading,
-                            placeholder = downloadItemArtworkPlaceholder,
```
luizgrp, line 110: nice, it is surfacing places where the placeholder was not needed 👍

auth/composables/src/main/java/com/google/android/horologist/auth/composables/dialogs/SignedInConfirmationDialog.kt
```diff
@@ -121,18 +121,17 @@ private fun SignedInConfirmationDialogContent(
...
                 .size(60.dp)
                 .background(color = Color(AVATAR_BACKGROUND_COLOR), shape = CircleShape),
             contentAlignment = Alignment.Center,
         ) {
-            if (hasAvatar) {
+            if (avatar != null) {
                 Image(
                     modifier = Modifier.clip(CircleShape),
-                    painter = rememberAsyncImagePainter(model = avatar),
+                    painter = avatar.rememberPainter(),
```
luizgrp, line 134: I like having a single remember function to call regardless of the implementation

compose-material/src/main/java/com/google/android/horologist/compose/material/ChipIconWithProgress.kt
```diff
@@ -150,32 +142,27 @@ private fun ChipIconWithProgressInternal(
...
-                    contentDescription = DECORATIVE_ELEMENT_CONTENT_DESCRIPTION,
-                    modifier = Modifier
-                        .align(Alignment.Center)
-                        .size(iconSize - indicatorPadding)
-                        .clip(CircleShape),
-                    contentScale = ContentScale.Crop,
-                    alpha = LocalContentAlpha.current,
-                )
-            }
+        icon ?: return
```
luizgrp, line 145: concise way of returning! :D

luizgrp (approved):
I like the improvements, even though it might look like it is making the API a bit less convenient.

compose-material/src/main/java/com/google/android/horologist/compose/material/ToggleButton.kt
```diff
@@ -87,8 +87,8 @@ public fun ToggleButton(
  */
 @Composable
 public fun ToggleButton(
-    checkedIcon: Any,
-    notCheckedIcon: Any,
+    checkedIcon: PaintableIcon,
```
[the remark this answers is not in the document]
luizgrp: this might be surfacing the lack of implementation in the current code of ToggleButton and Button to also support Image

compose-material/src/main/java/com/google/android/horologist/compose/material/Paintable.kt
luizgrp: If this get extracted to a separate module as previously suggested, then it should be fine to keep the name, and having "-coil" in the name of the lib.
Worth doing that extraction before next release (not necessarily in this PR) to avoid having 2 separated "big" API changes in the same area?

compose-material/src/main/java/com/google/android/horologist/compose/material/Paintable.kt
yschimke: I don't think so. It's not a known Stable type. And Composables have an implicit `this` param also, so effectively there from the fact it's an immutable type.

compose-material/src/main/java/com/google/android/horologist/compose/material/ToggleButton.kt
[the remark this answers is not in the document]
fstanis: `PaintableIcon` is a subtype used just to indicate "this could be used in an `Icon` instead of an `Image`". This is to keep compatibility with existing code that in some places uses `Icon` or `Image` and in some only `Icon` (like here).

auth/composables/src/main/java/com/google/android/horologist/auth/composables/chips/CreateAccountChip.kt
```diff
@@ -71,7 +72,7 @@ public fun CreateAccountChip(
             label = label,
             onClick = onClick,
             modifier = modifier,
-            icon = Icons.AutoMirrored.Outlined.SendToMobile,
+            icon = ImageVectorPaintable(Icons.AutoMirrored.Outlined.SendToMobile),
```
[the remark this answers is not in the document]
fstanis: Done.

media/ui/src/debug/java/com/google/android/horologist/media/ui/components/actions/ShowPlaylistChipPreview.kt
fstanis: I'm worried about the implications of this - what if the placeholder needs to load? This is just passing `coil` arguments verbatim.

compose-material/src/main/java/com/google/android/horologist/compose/material/Paintable.kt
fstanis: Agreed, moving this to a separate `-coil` library will be a quick follow-up. I only added it here to stay backwards-compatible.

PR #1870, opened 2023-12-14
Moves coil code into new library

fstanis:
#### WHAT
Follow up for #1868, creates a new library that will host coil-specific implementation of `Paintable` and other image-related implementations in the future.

#### WHY
To remove the hard dependency on coil for use-cases where simpler image loading is possible (e.g. apps only using local resources).

#### HOW
By moving `CoilPaintable` and `rememberArtworkColor` into a new `compose-material-coil` library.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


compose-material-coil/src/main/java/com/google/android/horologist/compose/material/coil/ArtworkColor.kt
```diff
@@ -0,0 +1,66 @@
...
+import androidx.core.graphics.drawable.toBitmap
+import androidx.palette.graphics.Palette
+import androidx.wear.compose.material.MaterialTheme
+import coil.imageLoader
+import coil.request.ImageRequest
+import com.google.android.horologist.annotations.ExperimentalHorologistApi
+
+@Composable
+@ExperimentalHorologistApi
+public fun rememberArtworkColor(
```
yschimke, line 35: Should we replace `Any?` with `String` here? 

Make a blanket rule that we avoid Any in composables?

yschimke (approved):

compose-material-coil/src/main/java/com/google/android/horologist/compose/material/coil/ArtworkColor.kt
fstanis: Missed this - fixed!

fstanis:
I went with `images/base` and `images/coil`instead of `compose-material` and `compose-material-coil` in the latest revision, PTAL.

yschimke (approved):
LGTM - I'll kick the tires once it's landed.

PR #1884, opened 2023-12-15
Optimize background rendering.

clickxu:
#### WHAT
Optimize gradient background rendering. 

#### WHY
Drawing background by Brush.radialGradient with Marquee is a bit costly.

#### HOW
Draw Radial gradient background into an offscreen bitmap within a Modifier.drawWithCache.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/auth/sample/wear/build.gradle.kts
+++ b/auth/sample/wear/build.gradle.kts
@@ -99,9 +99,6 @@ dependencies {
     implementation(projects.composables)
     implementation(projects.composeLayout)
     implementation(projects.composeMaterial)
-    implementation(projects.datalayer.core)
-    implementation(projects.datalayer.watch)
-    implementation(projects.datalayer.grpc)
 
     implementation(libs.androidx.activity.compose)
     implementation(libs.androidx.complications.data)
--- a/composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
@@ -29,7 +29,6 @@ import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.size
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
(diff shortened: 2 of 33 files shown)


media/ui/src/main/java/com/google/android/horologist/media/ui/components/background/ArtworkColorBackground.kt
```diff
@@ -103,15 +98,25 @@ public fun ColorBackground(
...
-                    listOf(
-                        animatedBackgroundColor.value
-                            .copy(alpha = 0.3f)
-                            .compositeOver(background),
-                        background,
-                    ),
-                ),
-            ),
+            .drawWithCache {
+                val bitmap = ImageBitmap(size.width.toInt(), size.height.toInt())
```
yschimke, line 102: [Does the type of Bitmap make any difference?](https://developer.android.com/reference/kotlin/androidx/compose/ui/graphics/ImageBitmap#prepareToDraw()) is interesting and might explain the performance improvement?

composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
```diff
@@ -160,6 +161,12 @@ private class MarqueeController(edgeGradientWidth: Dp) {
                 placeable.placeRelative(IntOffset.Zero)
             }
         }
+        .graphicsLayer {
+            // Required to make the faded edges only clear the alpha for the marquee content's
+            // pixels and not punch a hole through whatever is beneath this composable.
+            compositingStrategy = CompositingStrategy.Offscreen
```
yschimke, line 167: If you are happy with performance with this in place, then it's a great win.  Nice work.

yschimke (approved):

yschimke:
To fix the screenshot failures, run https://github.com/clickxu/horologist/actions/workflows/fixup.yml against your branchy

composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
clickxu: Yes, I copied this piece and (most of)  your pull/1857, since the Bitmap approach is drawn behind the Marquee, if I don't include this, the black fading area will be obvious。

yschimke: I'll close that other one.

auth/sample/wear/build.gradle.kts
```diff
@@ -99,9 +99,6 @@ dependencies {
     implementation(projects.composables)
     implementation(projects.composeLayout)
     implementation(projects.composeMaterial)
-    implementation(projects.datalayer.core)
```
yschimke, line 102: Something went wrong with your merge, but not sure if it's really an issue.

clickxu: I think I missed up the comment history. Would be happy to redo it.

PR #1886, opened 2023-12-15
App/Screen/Page Scaffold

yschimke:
#### WHAT

First draft of App/Screen/Page Scaffold with TimeText animation.

https://github.com/google/horologist/assets/231923/d57da0df-5bca-4aeb-91ad-473c9962f41d

#### WHY

See what this API would look like if we deprecate (not delete) WearNavScaffold.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


compose-layout/src/main/java/com/google/android/horologist/compose/layout/AppScaffold.kt
```diff
@@ -0,0 +1,65 @@
...
+        }
+    })
+}
+
+@Composable
+fun ScreenScaffold(
+    timeText: (@Composable () -> Unit)? = null,
+    scrollState: ScrollableState? = null,
+    pageIndicatorState: PageIndicatorState? = null,
+    positionIndicator: (@Composable () -> Unit)? = null,
```
rajat4064g, line 48: Why are these taken as params if they're not used in the method body?

yschimke: It's in progress, I wanted to start with whether the API is right.

Some alternatives, delete PageScaffold, and just move the PositionIndicator into smarter components like ScalingLazyColumn.

yschimke: I put an implementation in to start.  Using a Composition Local to coordinate.

sample/src/main/java/com/google/android/horologist/sample/SampleWearApp.kt
```diff
@@ -72,230 +72,310 @@ fun SampleWearApp() {
...
+            composable(
+                route = Screen.MaterialChipIconWithProgressScreen.route,
+            ) {
+                val columnState = rememberColumnState()
+
+                ScreenScaffold(scrollState = columnState) {
+                    SampleChipIconWithProgressScreen(columnState = columnState)
+                }
+            }
+            composable(
```
yschimke, line 204: @BowerSteve What are your thoughts on this? If I leave scrollable, then I need to still deal with passing the scroll state down.  So this seems cleaner to make apps responsible for creating the state and putting it into ScreenScaffold.

compose-layout/src/main/java/com/google/android/horologist/compose/layout/AppScaffold.kt
```diff
@@ -0,0 +1,205 @@
...
+import androidx.wear.compose.foundation.OnFocusChange
+import androidx.wear.compose.foundation.lazy.ScalingLazyListState
+import androidx.wear.compose.material.HorizontalPageIndicator
+import androidx.wear.compose.material.PageIndicatorState
+import androidx.wear.compose.material.PositionIndicator
+import androidx.wear.compose.material.Scaffold
+import androidx.wear.compose.material.TimeText
+
+/**
+ * App Scaffold to place *above* the SwipeDismissableNavHost.
```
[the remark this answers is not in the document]
luizgrp: Maybe elaborate with the advantages of/why using this component

compose-layout/src/main/java/com/google/android/horologist/compose/layout/AppScaffold.kt
```diff
@@ -0,0 +1,205 @@
...
+                PositionIndicator(scrollState)
+            }
+        },
+        content = content,
+    )
+}
+
+/**
+ * Pager Scaffold to place *inside* a single page of HorizontalPager.
+ * The TimeText if set will override the AppScaffold timeText.
```
luizgrp, line 133:
```suggestion
 * Pager Scaffold to be placed *inside* a single page of [HorizontalPager].
 * If set, the [TimeText] will override the [AppScaffold]'s timeText.
```

luizgrp: Maybe elaborate why this component should be used, instead of using the ScreenScaffold directly, or any other component. It seems to me that the advantage is that it handles the TimeText in a better way, but I'm not sure based on the current doc. If that's the case, it can be also made more explicitly that if the TimeText is not set, then there is no real advantage in using it?

media/sample/src/main/java/com/google/android/horologist/mediasample/ui/app/UampWearApp.kt
```diff
@@ -186,52 +203,68 @@ fun UampWearApp(
...
+                    ScreenScaffold(scrollState = columnState) {
+                        AudioDebugScreen(
+                            columnState = columnState,
+                            audioDebugScreenViewModel = hiltViewModel(),
+                        )
+                    }
                 }
 
-                scrollable(
+                composable(
```
luizgrp, line 220: nit: could we have a function (composableScreenScaffold?) that wraps the content in a `ScreenScaffold`?

luizgrp (approved):
I like the new API. Left few comments in order to improve the kdocs.

compose-layout/src/main/java/com/google/android/horologist/compose/layout/AppScaffold.kt
```diff
@@ -0,0 +1,205 @@
...
+    val scaffoldState = LocalScaffoldState.current.apply {
+        appTimeText.value = timeText
+    }
+
+    Scaffold(
+        modifier = modifier,
+        timeText = scaffoldState.timeText,
+    ) {
+        Box(modifier = Modifier.fillMaxSize()) {
+            content()
```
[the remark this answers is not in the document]
yschimke: Because there are so many params, part of the design here is to avoid wrapping and hiding the Wear Compose composables.

media/sample/src/main/java/com/google/android/horologist/mediasample/ui/app/UampWearApp.kt
yschimke: Yes, but I think I want to start without things that wrap other things and then work out where to draw the line. For follow up.

compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScrollAway.kt
```diff
@@ -16,128 +16,213 @@
...
-        scrollState = scalingLazyColumnState.state,
-        itemIndex = scalingLazyColumnState.initialScrollPosition.index,
-        offset = offset,
-    )
 }
 
-internal fun Modifier.scrollAway(
-    scrollState: State<ScrollableState?>,
+private fun Modifier.scrollAwayImpl(
+    scrollFn: Density.() -> ScrollParams,
```
yschimke, line 115: I think this is executed on every frame. Should we work out how to only execute when state changes? Maybe my bug, but I'm not sure where it is.

rajat4064g: Aren't all LayoutModifier dependents gets executed at every frame?

yschimke: Probably. I didn't think about it that way.  More that it only needs to do work when

a) The TimeText composable changes
b) The scrollState changes
c) the scroll position changes within some range (doesn't need to do anything while definitely out of range, something something `derivedStateOf`)

PR #1889, opened 2023-12-18
Fix chip layout

yschimke:
#### WHAT

Fix chip layout in sample app.

image

#### WHY

Should be using the material conventions.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


kul3r4 (approved):
thank you for fixing this!

PR #1897, opened 2023-12-19
Fix Pager usage of Scaffold

yschimke:
#### WHAT

Fix Pager usage of Scaffold.

#### WHY

All screens should be using the new scaffold types.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScreenScaffold.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScreenScaffold.kt
@@ -14,26 +14,30 @@
  * limitations under the License.
  */
 
-@file:OptIn(ExperimentalWearFoundationApi::class)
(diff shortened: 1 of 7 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScreenScaffold.kt
```diff
@@ -43,7 +47,7 @@ import androidx.wear.compose.material.Scaffold
  * @param modifier the Scaffold modifier.
  * @param timeText the page specific time text.
  * @param scrollState the ScrollableState to show in a default PositionIndicator.
- * @param pageIndicatorState state for a HorizontalPager.
+ * @param pagerState state for a HorizontalPager.
```
yschimke, line 50: @BowerSteve you were right, PageIndicatorState wasn't the right type.

luizgrp (approved):

yschimke:
Will follow up with comments.

PR #1906, opened 2023-12-20
Cleanup Data Layer Sample

yschimke:
#### WHAT

Cleanup Data Layer Sample

#### WHY

Avoid suggesting WearNavScaffold

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


datalayer/sample/wear/src/main/java/com/google/android/horologist/datalayer/sample/WearApp.kt
```diff
@@ -36,37 +44,59 @@ fun WearApp(
...
+                }
+            }
+            composable(route = Screen.AppHelperNodesActionsScreen.route) {
+                val columnState = rememberColumnState()
+
+                ScreenScaffold(scrollState = columnState) {
+                    NodesActionsScreen(columnState = columnState)
+                }
+            }
+            infoScreen(
```
luizgrp, line 96: infoScreen uses scrollable, should that be updated?

luizgrp (approved):

PR #1915, opened 2023-12-22
Make DataLayerAppHelper.startRemoteActivity more strict

luizgrp:
#### WHAT

Change `DataLayerAppHelper.startRemoteActivity` to only launch activities of the same package.

#### WHY

It was not the initial intention of this function to be able to launch activities from other apps.

#### HOW

- Remove `packageName` attribute from `ActivityConfig`.
- Set the package name of the running app on the `Intent` used to launch the activity.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/sample/phone/src/main/AndroidManifest.xml
+++ b/datalayer/sample/phone/src/main/AndroidManifest.xml
@@ -42,6 +42,13 @@
             </intent-filter>
         </activity>
 
+        <activity
+            android:name=".screens.startremote.StartRemoteSampleActivity"
+            android:exported="true"
+            android:taskAffinity=".main"
+            android:theme="@style/Theme.Horologist">
+        </activity>
+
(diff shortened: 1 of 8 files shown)


datalayer/sample/phone/src/main/AndroidManifest.xml
```diff
@@ -42,6 +42,13 @@
             </intent-filter>
         </activity>
 
+        <activity
+            android:name=".screens.startremote.StartRemoteSampleActivity"
+            android:exported="true"
```
yschimke, line 47: should we demo with exported = false?

yschimke (approved):

datalayer/sample/phone/src/main/AndroidManifest.xml
luizgrp: nice, thanks!

PR #1918, opened 2023-12-29
Simplify icons in marquee using InlineTextContent

fstanis:
#### WHAT
- Simplify #1863 to not require a `Row` when an icon is present.
- Makes `MarqueeTextMediaDisplay` only apply tint on `PaintableIcon`.

#### WHY
- Simpler code and simpler composable, but also an easier way to size the icon correctly using the text size.
- Image icons shouldn't be tinted and the type (`PaintableIcon` vs `Paintable`) conveys this (similar pattern is used elsewhere in the codebase).

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests

### DIFF
--- a/media/ui/src/test/java/com/google/android/horologist/media/ui/components/MarqueeTextMediaDisplayTest.kt
+++ b/media/ui/src/test/java/com/google/android/horologist/media/ui/components/MarqueeTextMediaDisplayTest.kt
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/MarqueeTextMediaDisplay.kt
```diff
@@ -114,3 +131,22 @@ public fun MarqueeTextMediaDisplay(
         }
     }
 }
+
+@Composable
+private fun MediaTitleIcon(icon: Paintable) {
+    if (icon is PaintableIcon) {
```
yschimke, line 137: Would be good to have both of these covered by a test.

yschimke (approved):

media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/MarqueeTextMediaDisplay.kt
fstanis: Done.

Issue #1920, opened 2024-01-02
Talkback speaking device output text twice in VolumeScreen


luizgrp:

luizgrp:
Issue doesn't seem to be surfaced in the a11y screenshot tests: 
- https://github.com/google/horologist/blob/bccd49197af93ae1c863b2ef1888dc1b334aaaf7/media/audio-ui/src/test/snapshots/images/com.google.android.horologist.audio.ui_VolumeScreenA11yScreenshotTest_volumeScreenAtMinimums.png
- https://github.com/google/horologist/blob/main/media/audio-ui/src/test/snapshots/images/com.google.android.horologist.audio.ui_VolumeScreenA11yScreenshotTest_volumeScreenNotConnected.png

PR #1922, opened 2024-01-02
Fix double talkback on VolumeScreen

yschimke:
#### WHAT

Fix double talkback on VolumeScreen.

#### WHY

Even when merged, two nodes in the A11y tree, one with text, one with contentDescription will both be seen by talkback on Wear 4 devices.

#fixes https://github.com/google/horologist/issues/1920

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
+++ b/media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
@@ -45,6 +45,7 @@ import com.google.android.horologist.audio.ui.components.DeviceChip
 import com.google.android.horologist.audio.ui.components.toAudioOutputUi
 import com.google.android.horologist.compose.material.Icon
 import com.google.android.horologist.compose.material.IconRtlMode
+import com.google.android.horologist.compose.material.util.DECORATIVE_ELEMENT_CONTENT_DESCRIPTION
 import com.google.android.horologist.compose.rotaryinput.RotaryDefaults.isLowResInput
 import com.google.android.horologist.images.base.paintable.ImageVectorPaintable.Companion.asPaintable
 
@@ -117,7 +118,7 @@ public fun VolumeScreen(
                 icon = {
                     Icon(
                         paintable = audioOutputUi.imageVector.asPaintable(),
-                        contentDescription = null,
+                        contentDescription = DECORATIVE_ELEMENT_CONTENT_DESCRIPTION,
                         tint = MaterialTheme.colors.onSurfaceVariant,
                     )
                 },
--- a/media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
+++ b/media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
@@ -117,7 +117,7 @@ public fun VolumeScreen(
                 icon = {
                     Icon(
                         paintable = audioOutputUi.imageVector.asPaintable(),
-                        contentDescription = audioOutputUi.displayName,
+                        contentDescription = null,
                         tint = MaterialTheme.colors.onSurfaceVariant,
                     )
                 },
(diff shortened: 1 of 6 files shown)


media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/DeviceChip.kt
```diff
@@ -45,7 +45,7 @@ public fun DeviceChip(
     Chip(
         modifier = modifier
             .width(intrinsicSize = IntrinsicSize.Max)
-            .semantics {
+            .semantics(mergeDescendants = true) {
```
luizgrp, line 48: Is it needed? I would expect it to be redundant, aren't they already merged?

media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
```diff
@@ -117,7 +117,7 @@ public fun VolumeScreen(
                 icon = {
                     Icon(
                         paintable = audioOutputUi.imageVector.asPaintable(),
-                        contentDescription = audioOutputUi.displayName,
+                        contentDescription = null,
```
luizgrp, line 120: nit: use const `DECORATIVE_ELEMENT_CONTENT_DESCRIPTION`

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/a11y/A11ySnapshotTransformer.kt
```diff
@@ -128,7 +128,8 @@ internal class A11ySnapshotTransformer : SnapshotTransformer {
             }
             if (it.contentDescription != null) {
                 drawItem("Content Description \"${it.contentDescription.joinToString(", ")}\"")
-            } else if (it.text != null) {
+            }
+            if (it.text != null) {
```
luizgrp, line 132: This change by itself will introduce wrong screenshots, for example, in the scenario of a single node with both text and content description.

media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/DeviceChip.kt
yschimke: In practice it seems so, but I couldn't see where. https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/compose/compose-material/src/main/java/androidx/wear/compose/material/Chip.kt;l=1399?q=ChipImpl&sq=&ss=androidx%2Fplatform%2Fframeworks%2Fsupport:wear%2Fcompose%2F

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/a11y/A11ySnapshotTransformer.kt
yschimke: I don't see how to easily detect this here.  So I'll see what fails in our tests.

media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/DeviceChip.kt
luizgrp: Is in practice making any difference? In my run, just setting the content description to null had fixed the issue. 

Is this needed due to the changes made to A11ySnapshotTransformer in this PR?

luizgrp (approved):
LGTM - I still don't understand the need for the mergeDescendants, also left a suggestion.

media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/components/DeviceChip.kt
yschimke: I think it's correct anyway.

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/a11y/A11ySnapshotTransformer.kt
yschimke: Tests are passing, I'd like to land with an effective test.

PR #1923, opened 2024-01-02
Update wear-remote-interactions to 1.1.0-alpha01

luizgrp:
#### WHAT

Update wear-remote-interactions to `1.1.0-alpha01`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


datalayer/phone/build.gradle.kts
```diff
@@ -93,6 +93,7 @@ dependencies {
     implementation(libs.kotlinx.coroutines.core)
 
     api(libs.playservices.wearable)
+    implementation(libs.kotlinx.coroutines.guava)
```
yschimke, line 96: maybe use androidx.concurrent-futures-ktx?

See https://cs.android.com/androidx/platform/frameworks/support/+/4a8de953ddb7ed73c321a0559464d49d4e5bb4cb and https://b.corp.google.com/issues/275580742

yschimke (approved):

datalayer/phone/build.gradle.kts
luizgrp: it didn't seem to find the extension function with that one, but found in kotlinx-coroutines-core instead, lmk

PR #1928, opened 2024-01-03
Normal png res file

yschimke:
#### WHAT

Stop sample_image.png being lfs.

#### WHY

Causing issues with repo synching.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


.gitattributes
```diff
@@ -1,5 +1,5 @@
 **/test/snapshots/** filter=lfs diff=lfs merge=lfs -text
 **/figma/** filter=lfs diff=lfs merge=lfs -text
+**/docs/**/*.png filter=lfs diff=lfs merge=lfs -text
```
luizgrp, line 3: we also want the pngs for the tests, just have to make sure that the filter of line 1 is working

luizgrp (approved):
LGTM

.gitattributes
yschimke: I'll test.

yschimke: Worked

yschimke:
Used https://stackoverflow.com/questions/35011366/move-git-lfs-tracked-files-under-regular-git

PR #1929, opened 2024-01-03
Add datalayer-phone-ui with Install App prompt feature.

luizgrp:
#### WHAT

Add `datalayer-phone-ui` library with Install App prompt feature.

https://github.com/google/horologist/assets/878134/a6a2c488-3a76-4984-851b-0ca30d30becd

#### HOW

- Add `datalayer-phone-ui` module;
- Add `PhoneUiDataLayerHelper` class as API entry point;
- Add `InstallAppBottomSheet` composable which is public and can also be used directly without the API;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


datalayer/phone-ui/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,17 @@
...
+    xmlns:tools="http://schemas.android.com/tools"
+    package="com.google.android.horologist.datalayer.phone.ui">
+
+    <uses-sdk
+        android:minSdkVersion="21"
+        tools:ignore="GradleOverrides" />
+
+    <application>
+        <activity
+            android:name=".prompt.installapp.InstallAppBottomSheetActivity"
```
yschimke, line 12: Nice!

yschimke:
LGTM - but Garan and Chiara are proper reviewers.

kul3r4 (approved):
LGTM, we can think about renaming the package name to not include datalayer at least for the prompts but we can do this later.
Also, worth thinking about if keeping or not the screenshots if those can't be automated to avoid additional burden on developers for manual updates.

PR #1930, opened 2024-01-03
Bump to metalava 0.3.5

yschimke:
#### WHAT

Bump to metalava 0.3.5

#### WHY

Hopefully fix kotlin 1.9 warnings

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


auth/data-phone/api/current.api
```diff
@@ -1,16 +1,9 @@
 // Signature format: 4.0
-package com.google.android.horologist.auth.data.phone.common {
-
-  public final class LoggingKt {
```
luizgrp, line 4: finally those empty classes are not published anymore!

luizgrp (approved):
Thanks!

PR #1941, opened 2024-01-05
Updating AS files we are fighting over

yschimke:
#### WHAT

Updating AS files we are fighting over

#### WHY

Some updates after dependency updates

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
Landing as non-controversial.

PR #1945, opened 2024-01-08
AI UI library

yschimke:
#### WHAT

AI UI library

#### WHY

Show some samples of AI such as Gemini on Wear.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


amankhoza:
LGTM

ai/ui/src/main/java/com/google/android/horologist/ai/ui/components/PromptDisplay.kt
```diff
@@ -0,0 +1,62 @@
...
+public fun TextPromptDisplay(
+    prompt: TextPromptUiModel,
+    modifier: Modifier = Modifier,
+    onClick: () -> Unit = {},
+) {
+    Card(
+        modifier = modifier.fillMaxWidth(),
+        onClick = onClick,
+        backgroundPainter = CardDefaults.cardBackgroundPainter(
+            Color(0xff8ab4f8),
```
fstanis, line 46: Maybe extract this color into a named `const`?

ai/ui/src/main/java/com/google/android/horologist/ai/ui/model/PromptUiModel.kt
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.ai.ui.model
+
+import java.time.Instant
+
+public sealed interface PromptUiModel : PromptResponseUiModel
```
fstanis, line 21: Maybe I'm missing something, but prompt being a subclass of prompt response seems a bit odd, is this international?

yschimke: PromptResponse means PromptOrResponse.  I'll rename.

yschimke: Will follow up with this.

ai/ui/src/main/java/com/google/android/horologist/ai/ui/screens/PromptScreen.kt
```diff
@@ -0,0 +1,108 @@
...
+    promptEntry: @Composable () -> Unit,
+) {
+    ScalingLazyColumn(columnState = columnState, modifier = modifier) {
+        item {
+            ListHeader(modifier = Modifier.fillMaxWidth(0.8f)) {
+                Text(text = uiState.modelInfo?.name ?: "...")
+            }
+        }
+        uiState.messages.forEach {
+            item {
```
fstanis, line 65: Is there something we can use as a key here? May not be needed yet, just thinking about future-proofing.

fstanis (approved):

ai/ui/src/main/java/com/google/android/horologist/ai/ui/screens/PromptScreen.kt
yschimke: Good question.  Not sure.

Also a bigger question of how big these will get.  Should they use paging to lazy load them?

Not going to think about this yet.

ai/ui/src/main/java/com/google/android/horologist/ai/ui/components/PromptDisplay.kt
yschimke: I really don't think I typed that value.  wonder if refactoring replaced a constant or something....

```
public class Colors(
    primary: Color = Color(0xFFAECBFA),
    primaryVariant: Color = Color(0xFF8AB4F8),
```

yschimke: Color can't be const can it? just a val. I'll change to MaterialTheme.colors.primaryVariant

PR #1946, opened 2024-01-08
Rely on automatically created launchers

yschimke:
#### WHAT

Rely on automatically created launchers

#### WHY

Duplicates since AS makes them as well.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
config only, landing.

PR #1950, opened 2024-01-10
Update CustomMediaActionButton UI to be in line with MediaButton.

kpaikena:
#### WHAT
Update CustomMediaActionButton UI.

#### WHY
To align UI with other media buttons.

#### HOW
Changed default icon size, and use UnboundedRippleButton to show custom ripple.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
Surprised it isn't covered by a screenshot test

PR #1961, opened 2024-01-12
Fix time text on NodeDetailsScreen

luizgrp:
#### WHAT

Fix time text on `NodeDetailsScreen`.

#### WHY

It was not hiding when scrolling through the list.

#### HOW

Use `ScreenScaffold`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/sample/wear/src/main/java/com/google/android/horologist/datalayer/sample/screens/nodesactions/NodeDetailsScreenNavigation.kt
+++ b/datalayer/sample/wear/src/main/java/com/google/android/horologist/datalayer/sample/screens/nodesactions/NodeDetailsScreenNavigation.kt
@@ -21,7 +21,9 @@ import androidx.navigation.NavController
 import androidx.navigation.NavGraphBuilder
 import androidx.navigation.NavType
 import androidx.navigation.navArgument
-import com.google.android.horologist.compose.navscaffold.scrollable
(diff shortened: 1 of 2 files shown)


datalayer/sample/wear/src/main/java/com/google/android/horologist/datalayer/sample/screens/nodesactions/NodeDetailsScreenNavigation.kt
```diff
@@ -21,7 +21,9 @@ import androidx.navigation.NavController
 import androidx.navigation.NavGraphBuilder
 import androidx.navigation.NavType
 import androidx.navigation.navArgument
-import com.google.android.horologist.compose.navscaffold.scrollable
```
yschimke, line 24: We should probably kill these off, to stop them getting used....

yschimke: I'll deprecate.

yschimke:
InfoScreenNavigation is also using the wrong DSL method.  My fault, should deprecate these.

luizgrp:
oh I didn't realise that this PR was using the wrong one

yschimke (approved):

PR #1964, opened 2024-01-12
Improve explanation of install prompt demo for Gmail

kul3r4:
#### WHAT
Improve explanation of install prompt demo for Gmail

#### WHY
Not clear what it does before running it

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):
Thanks!

PR #1972, opened 2024-01-15
Reduce libraries to minSdk 25

### LINKED ISSUE #1968: Revert minSdk from 26 back to 25
This is COMPLETLY INSANE!!! You can't just arbitrarily introduce breaking changes without incrementing the major version let alone even MENTIONING it on the changelog. Please follow semantic versioning like any sane person would do! I don't know why this change was needed in the first place but you need to revert it ASAP since minSdk 25 is the defacto default for Wear OS applications and many people rely on this library.

yschimke:
#### WHAT

Reduce libraries to minSdk 25.

#### WHY

Wear OS exists for SDK 25+. Horologist doesn't need to be the limiting factor.

#fixes #1968 

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
2 things we are clearly overrding/ignoring

```
        public suspend fun markComplicationAsActivated(
            complicationName: String,
            complicationInstanceId: Int,
            complicationType: ComplicationType,
        ) {
```

This API relies on ComplicationType which is minSdk 26

```

This API is also using complications.

```
public class MediaStatusTemplate(
    context: Context,
) :
    TypedComplicationTemplate<MediaStatusTemplate.Data>(context) {
```

**Comment by yschimke** [comment]
Hitting

``` /home/runner/work/horologist/horologist/composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt:86: Error: Call requires API level 26 (current min is 25): java.time.LocalDate#now [NewApi]
      date: LocalDate = LocalDate.now(),
                                  ~~~
  
  The full lint text report is located at:
    /home/runner/work/horologist/horologist/composables/build/intermediates/lint_intermediate_text_report/debug/lint-results-debug.txt
```

ai/ui/build.gradle.kts
```diff
@@ -25,7 +25,7 @@ android {
     compileSdk = 34
 
     defaultConfig {
-        minSdk = 26
+        minSdk = 25
```
luizgrp, line 28: minSdkVersion="30" in androidmanifest of this module:

https://github.com/google/horologist/blob/0248c3b6330a6a5d4f191566f7d490c1c7d866e2/ai/ui/src/main/AndroidManifest.xml#L24

compose-layout/build.gradle.kts
```diff
@@ -25,7 +25,7 @@ android {
     compileSdk = 34
 
     defaultConfig {
-        minSdk = 26
+        minSdk = 25
```
luizgrp, line 28: missing changes in androidmanifest of this module

datalayer/watch/src/main/AndroidManifest.xml
```diff
@@ -19,8 +19,15 @@
...
-        android:minSdkVersion="26"
-        tools:ignore="GradleOverrides" />
+        android:minSdkVersion="25"
+        tools:ignore="GradleOverrides"
+        tools:overrideLibrary="
+            androidx.wear.watchface.data,
+            androidx.wear.watchface.complications.datasource,
+            androidx.wear.watchface.complications.datasource.ktx,
+            androidx.wear.protolayout.expression,
+            androidx.wear.protolayout.expression.pipeline"
```
luizgrp, line 29: Is there a way of avoiding this override?
E.g. marking the functions that require API > 25 with `@RequiresApi(Build.VERSION_CODES.O)`?

media/sample/src/main/AndroidManifest.xml
```diff
@@ -133,7 +133,7 @@
     </application>
 
     <uses-sdk
-        android:minSdkVersion="26"
+        android:minSdkVersion="25"
```
luizgrp, line 136: missing changes in build.gradle of this module

roboscreenshots/build.gradle.kts
```diff
@@ -25,7 +25,7 @@ android {
     compileSdk = 34
 
     defaultConfig {
-        minSdk = 26
+        minSdk = 25
```
luizgrp, line 28: missing androidmanifest for the module - ignore comment if we are not planning to import this module internally

luizgrp (approved):

roboscreenshots/build.gradle.kts
yschimke: No plans to import. I'd love to kill it off.  Use AGP screenshotTest when available.

datalayer/watch/src/main/AndroidManifest.xml
yschimke: No, it's not the APIs, it's merging the manifests that is the problem.

In this case, we could probably change the param we use to a string and have no hard dependency.

Worth changing?

compose-layout/build.gradle.kts
yschimke: Was already 25!

yschimke:
This will also required coreLibraryDesugaring enabled in all apps using this. Previously not required as API 26 allowed java.time.

datalayer/watch/src/main/AndroidManifest.xml
luizgrp: probably not, let's see how it goes with this?

PR #1978, opened 2024-01-15
Renaming for app prompts and updated text in counter sample.

kul3r4:
#### WHAT
Renamed demo1 and demo2 to describe what they do. Also renamed text for counter sample to match phone and watch. 

#### WHY
To make it easy for developer running the sample to understand the purpose of the samples.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


kul3r4:
Closing this pull request as we need a different strategy for renaming, will send a separate one for improving the counter sample.

PR #1992, opened 2024-01-18
Improve installOnNode to return AppHelperResultCode

luizgrp:
#### WHAT

Improve `installOnNode` to return `AppHelperResultCode`.

#### WHY

Fixes https://github.com/google/horologist/issues/1902

#### Checklist :clipboard:
- [x] Add explicit visibility modifier and explicit return types for public declarations
- [x] Run spotless check
- [x] Run tests
- [x] Update metalava's signature text files

PR #1993, opened 2024-01-18
Fix implementation of installOnNode for phones

luizgrp:
#### WHAT

Fix implementation of installOnNode for phones.

#### WHY

- From [RemoteActivityHelper#getAvailabilityStatus](https://developer.android.com/reference/androidx/wear/remote/interactions/RemoteActivityHelper#getAvailabilityStatus()): 
> On phone devices, it will always return STATUS_UNKNOWN

- Missed adding comment to the watch implementation, requested in https://github.com/google/horologist/pull/1992#discussion_r1457584821 

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):

yschimke:
So not broken, just pointless to check?

luizgrp:
> So not broken, just pointless to check?

Yes, based on my understanding of the docs. Just when called by phones, for watches it is advised to check.

PR #1997, opened 2024-01-19
Screenshot test for responsive dialogs

yschimke:
#### WHAT

Screenshot test for responsive dialogs.

#### WHY

Use the preferred version of dialog layout for testing.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


compose-material/src/main/java/com/google/android/horologist/compose/material/ResponsiveDialog.kt
```diff
@@ -55,12 +56,14 @@ public fun ResponsiveDialogContent(
     onCancelButtonClick: (() -> Unit)? = null,
     okButtonContentDescription: String = stringResource(R.string.ok),
     cancelButtonContentDescription: String = stringResource(R.string.cancel),
+    state: ScalingLazyColumnState = rememberColumnState(responsive(firstItemIsFullWidth = icon == null)),
```
yschimke, line 59: state is exposed because Dialog wants it already? Unless we don't want to pass it in, and keep handling here?

same for showPositionIndicator?

compose-material/src/main/java/com/google/android/horologist/compose/material/ResponsiveDialog.kt
```diff
@@ -55,12 +56,14 @@ public fun ResponsiveDialogContent(
...
     content: (ScalingLazyListScope.() -> Unit)? = null,
 ) {
-    val state = rememberColumnState(
-        ScalingLazyColumnDefaults.responsive(firstItemIsFullWidth = icon == null),
-    )
-    ScreenScaffold(modifier = modifier.fillMaxSize(), scrollState = state) {
+    ScreenScaffold(
+        modifier = modifier.fillMaxSize(),
+        scrollState = if (showPositionIndicator) state else null,
+        timeText = {}) {
```
yschimke, line 66: disable TimeText when this is showing.

ssancho14: Isn't no time text the default?

compose-material/src/main/java/com/google/android/horologist/compose/material/ResponsiveDialog.kt
ssancho14: Fair, the issue is that the default is overridden then the `firstItemIsFullWidth` may be set wrong?

compose-material/src/main/java/com/google/android/horologist/compose/material/ResponsiveDialog.kt
yschimke: Not for most screens, time text is on.

I think Dialog doesn't, so maybe you are right.

compose-material/src/main/java/com/google/android/horologist/compose/material/ResponsiveDialog.kt
yschimke: Should we internalise it, but also disable this in Dialog?

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/compose/compose-material/src/main/java/androidx/wear/compose/material/dialog/Dialog.android.kt;l=82?q=f:wear%20Dialog&sq=

passing in null?

ssancho14: But then we lose the PositionIndicator, right?

yschimke: We can add it back. I mean we cant achieve both with current APIs. 

Feel free to change the APIs to what you want to achieve.

PR #1998, opened 2024-01-22
Add sample screen for connectedAndInstalledNodes function

luizgrp:
#### WHAT

Add sample screen for `connectedAndInstalledNodes` function.

https://github.com/google/horologist/assets/878134/cdd5a224-718d-4176-bde7-ce7fef263840

#### WHY

Show case the functionality.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/nodeslistener/NodesListenerScreen.kt
+++ b/datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/nodeslistener/NodesListenerScreen.kt
@@ -1,0 +1,175 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.datalayer.sample.screens.nodeslistener
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxSize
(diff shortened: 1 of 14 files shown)


datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/nodeslistener/NodesListenerScreen.kt
```diff
@@ -0,0 +1,175 @@
...
+
+@Composable
+fun NodesListenerScreen(
+    modifier: Modifier = Modifier,
+    viewModel: NodesListenerViewModel = hiltViewModel(),
+) {
+    val state by viewModel.uiState.collectAsStateWithLifecycle()
+
+    if (state == NodesListenerScreenState.Idle) {
+        viewModel.initialize()
```
yschimke, line 50: Should this be a SideEffect?

yschimke (approved):

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/nodeslistener/NodesListenerScreen.kt
luizgrp: Could you please elaborate why would you think this should be a SideEffect? I would go with your suggestion, just would like to understand it for my own learning

yschimke: https://developer.android.com/jetpack/compose/side-effects

> A side-effect is a change to the state of the app that happens outside the scope of a composable function. Due to composables' lifecycle and properties such as unpredictable recompositions, executing recompositions of composables in different orders, or recompositions that can be discarded, composables [should ideally be side-effect free](https://developer.android.com/jetpack/compose/mental-model).
>
> However, sometimes side-effects are necessary, for example, to trigger a one-off event such as showing a snackbar or navigate to another screen given a certain state condition. These actions should be called from a controlled environment that is aware of the lifecycle of the composable. In this page, you'll learn about the different side-effect APIs Jetpack Compose offers.

...

> To share Compose state with objects not managed by compose, use the [SideEffect](https://developer.android.com/reference/kotlin/androidx/compose/runtime/package-summary#SideEffect(kotlin.Function0)) composable. Using a SideEffect guarantees that the effect executes after every successful recomposition. On the other hand, it is incorrect to perform an effect before a successful recomposition is guaranteed, which is the case when writing the effect directly in a composable.

PR #1999, opened 2024-01-23
Add sample screen for connectedAndInstalledNodes function

luizgrp:
#### WHAT

Add sample screen in wear sample app for `connectedAndInstalledNodes` function.

Screenshot_20240123_105045

#### WHY

To demo the functionality.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/nodes/NodesScreen.kt
+++ b/datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/nodes/NodesScreen.kt
@@ -230,7 +230,7 @@ fun NodesScreenPreview() {
 
 @Preview(showBackground = true)
 @Composable
-fun NodesActionsScreenPreviewEmptyNodes() {
+fun NodesScreenPreviewEmptyNodes() {
     NodesScreen(
         state = NodesScreenState.Loaded(emptyList()),
         onRefreshClick = { },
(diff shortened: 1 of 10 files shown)


datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/nodeslistener/NodesListenerScreen.kt
```diff
@@ -71,7 +71,7 @@ fun NodesListenerScreen(
         item {
             Text(
                 text = stringResource(id = R.string.nodes_listener_screen_header),
-                modifier = Modifier.padding(vertical = 10.dp),
+                modifier = Modifier.padding(10.dp),
```
yschimke, line 74: Why custom padding? Should `ListHeader {.Text() }` suffice?

yschimke (approved):

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/nodeslistener/NodesListenerScreen.kt
luizgrp: This line is in the phone sample so there is no ListHeader as far as I know, but will address the one on the watch sample

luizgrp: For wear we have `Title` in Horologist compose material, which doesn't have a hard coded height as `ListHeader` does, which seems to be obsolete, see https://issuetracker.google.com/251166127.

So I'd go with Title + padding, which doesn't change much from what's already here.

yschimke: Ignore, was thinking Wear.

PR #2000, opened 2024-01-23
Improve connectedAndInstalledNodes to only query/listen to a single capability

luizgrp:
#### WHAT

Improve `connectedAndInstalledNodes` to only query/listen to a single capability.

#### WHY

- Remove assumption that a phone won't see another phone node - phone will now only listen to nodes with the capability meant for watches (`data_layer_app_helper_device_watch`);
- Remove assumption that a watch won't see another watch node nearby - watch will now only listen to nodes with the capability meant for phones (`data_layer_app_helper_device_phone`);
- Avoid issues in the future if another capability is added with prefix `data_layer_app_helper_device_`;

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/core/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
+++ b/datalayer/core/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
@@ -19,7 +19,6 @@ package com.google.android.horologist.data.apphelper
 import android.app.ActivityManager
 import android.app.ActivityManager.RunningAppProcessInfo.IMPORTANCE_FOREGROUND
 import android.content.Context
-import android.net.Uri
 import android.os.Process
 import androidx.annotation.CheckResult
 import androidx.wear.remote.interactions.RemoteActivityHelper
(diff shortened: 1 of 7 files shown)


datalayer/core/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
```diff
@@ -106,32 +105,48 @@ abstract class DataLayerAppHelper(
...
+                    val unused = trySend(capability.nodes.filter { it.isNearby }.toSet())
+                }
 
-        val allCaps = registry.capabilityClient.getAllCapabilities(
-            CapabilityClient.FILTER_REACHABLE,
-        ).await()
-        val installedCaps =
-            allCaps.filter { it.key.startsWith(CAPABILITY_DEVICE_PREFIX) }.values.flatMap { it.nodes }
-                .filter { it.isNearby }.toSet()
+            val installedDeviceCapabilityUri = "wear://*/$capability"
```
yschimke, line 135: Maybe worth a comment why you are using this form instead of the simpler

```
  public abstract Task<Void> addListener(
      @NonNull OnCapabilityChangedListener listener, @NonNull String capability);
```

yschimke (approved):

datalayer/core/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
luizgrp: it was tunnel vision when changing the code - thanks, changed to use the simpler one 👍

PR #2007, opened 2024-01-25
Fixes responsive SLC docs

garanj:
#### WHAT

Fixes responsive example for ScalingLazyColumn

#### WHY

Example doesn't build when copy/pasted

#### HOW

Fixed and confirmed build and execute

#### Checklist :clipboard:
- [N/A] Run spotless check
- [N/A] Run tests

### DIFF
--- a/docs/compose-layout.md
+++ b/docs/compose-layout.md
@@ -1,5 +1,10 @@
 # Compose Layout library
 
+> [!IMPORTANT]
+> All the examples on this page make use of imports from Horologist's `compose-layout` and `compose-material`.
+> Add these dependencies and imports to ensure you are using the Horologist version of components with the same
+> name in Wear Compose, for example `Chip`.
+
 ## ScalingLazyColumn responsive() layout.
 
 The `responsive()` layout factory will ensure that your ScalingLazyColumn is positioned correctly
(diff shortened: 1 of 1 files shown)


docs/compose-layout.md
```diff
@@ -41,7 +41,10 @@ Scaffold(
             }
         }
         items(10) {
-            Chip("Item $it", onClick = {})
+            Chip(
```
yschimke, line 44: You need the import from compose-material

yschimke (changes requested):
show an import?

docs/compose-layout.md
garanj: Do you mean the original would work with an additional import? I have androidx.wear.compose:compose-material 1.3.0.

Maybe 

```
Chip(onClick = {}) {
    Text("Item $it")
}
```

is better?

yschimke: See https://google.github.io/horologist/compose-material/#compose-material

garanj: Got it!!

garanj:
Updated - apologies, makes sense now!😀

docs/compose-layout.md
```diff
@@ -17,6 +17,8 @@ with `responsive()` will handle all the following:
 - Size side margins based on a percentage, adapting to different screen sizes.
 
 ```kotlin
+import com.google.android.horologist.compose.material.Chip
```
yschimke, line 20: TIOLI: should we import * here to not fix on specific classes. You are missing a lot of other imports.

yschimke:
Should we add a banner on top once showing imports come from compose-material?

garanj:
Added a note!

PR #2011, opened 2024-01-26
Remove isNearby filter from AppHelper functions

luizgrp:
#### WHAT

1) Remove `isNearby` filter from `connectedNodes` and `connectedAndInstalledNodes` functions from AppHelper.

2) Add `isNearby` property to `AppHelperNodeStatus` and update samples to display it.

<details><summary>Phone sample</summary>
<p>

https://github.com/google/horologist/assets/878134/c1e911da-2665-4508-b196-c36b95568b52

</p>
</details> 

<details><summary>Wear sample</summary>
<p>

https://github.com/google/horologist/assets/878134/a5861b25-7634-4eed-af28-767fedce3182

</p>
</details> 

#### WHY

I think it was being used as a way to filter "connected" nodes, however, this is not correct.
Retrieving reachable nodes from `CapabilityClient` already returns nodes that are connected. 
IsNearby is only filtering by those connected directly to the caller device.

Use case which shows that it was not working as intended:
1) Switch off bluetooth from the phone;
2) Have a paired watch device connected to wifi;
3) Launch datalayer sample on the phone, go the Nodes or Nodes Listener screen;

Expected result: watch device should be listed.
Actual result: watch device is not listed.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/core/api/current.api
+++ b/datalayer/core/api/current.api
@@ -93,19 +93,22 @@ package com.google.android.horologist.data {
 package com.google.android.horologist.data.apphelper {
 
   public final class AppHelperNodeStatus {
-    ctor public AppHelperNodeStatus(String id, String displayName, com.google.android.horologist.data.apphelper.AppInstallationStatus appInstallationStatus, optional error.NonExistentClass surfacesInfo);
+    ctor public AppHelperNodeStatus(String id, String displayName, boolean isNearby, com.google.android.horologist.data.apphelper.AppInstallationStatus appInstallationStatus, optional error.NonExistentClass surfacesInfo);
     method public String component1();
     method public String component2();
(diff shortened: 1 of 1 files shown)


garanj:
The original intention of using `isNearby` was to only include those devices that were connected via BT (as that's the only 1-hop transport supported right now?). This was with the assumptions that:

1. The user would be needing their phone and watch with them (so in close proximity) for whatever they were doing
2. Bluetooth was enabled

Maybe (2) doesn't hold? Maybe some people have BT off?

yschimke (approved):

luizgrp:
@yschimke @garanj added more changes after our discussion - updated the PR description, including video to the samples - please review again when you get a chance

yschimke (approved):

PR #2019, opened 2024-02-01
Represent the invalid dates in DatePicker in grey color.

rajat4064g:
#### WHAT
Make the invalid dates (dates before fromDate and after toDate) visible in DatePicker but represent them in grey color.
This also solves the retaining of previous day value in the day picker if month picker is scrolled.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
@@ -14,12 +14,9 @@
  * limitations under the License.
  */
 
-@file:OptIn(ExperimentalFoundationApi::class)
-
 package com.google.android.horologist.composables
 
 import androidx.compose.animation.core.Animatable
-import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.focusable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.BoxWithConstraints
@@ -95,13 +92,7 @@ public fun DatePicker(
     }
 
     val datePickerState = remember(date) {
-        if (fromDate != null && toDate == null) {
-            DatePickerState(date = date, fromDate = fromDate, toDate = fromDate.plusYears(3000))
-        } else if (fromDate == null && toDate != null) {
-            DatePickerState(date = date, fromDate = toDate.minusYears(3000), toDate = toDate)
-        } else {
-            DatePickerState(date, fromDate, toDate)
-        }
+        DatePickerState(date, fromDate, toDate)
     }
 
     val touchExplorationStateProvider = remember { DefaultTouchExplorationStateProvider() }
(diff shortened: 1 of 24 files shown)


yschimke:
Can you add a test to com.google.android.horologist.screensizes.DatePickerTest that covers this?

Would be good to see the result visually.

But I'm probably not the right person to review.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -588,19 +589,23 @@ internal fun pickerGroupItemWithRSB(
     )
 }
 
-internal fun pickerTextOption(textStyle: TextStyle, indexToText: (Int) -> String):
-    (@Composable PickerScope.(optionIndex: Int, pickerSelected: Boolean) -> Unit) = { value: Int, pickerSelected: Boolean ->
+internal fun pickerTextOption(
+    isValid: (Int) -> Boolean = { true },
```
BowerSteve, line 593: Maybe add isValid as the last parameter (it's optional), then textStyle wouldn't need to be specified as a named parameter?

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -332,6 +334,11 @@ public fun DatePicker(
                         }
                         .focusRequester(focusRequesterConfirmButton)
                         .focusable(),
+                    enabled = if (pickerGroupState.selectedIndex >= 2) {
+                        datePickerState.isDayValid()
```
BowerSteve, line 338: Seems wrong - shouldn't it be checking for valid month and year as well?

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
```diff
@@ -413,72 +415,105 @@ internal class DatePickerState constructor(
...
     private val toDate: LocalDate?,
 ) {
-    private val yearOffset = fromDate?.year ?: 1
-    val numOfYears = (toDate?.year ?: 3000) - (yearOffset - 1)
     val yearState =
         PickerState(
-            initialNumberOfOptions = numOfYears,
-            initiallySelectedOption = date.year - yearOffset,
+            initialNumberOfOptions = 200,
+            initiallySelectedOption = date.year - 1901,
```
BowerSteve, line 421: Worth commenting that the year range 1900-2100 has come from b/277885199 (and should it be 201 years to include both ends of the range?)

BowerSteve (commented):
Please add from-date and to-date samples into the Horologist sample app.

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
rajat4064g: Done

composables/src/main/java/com/google/android/horologist/composables/DatePicker.kt
rajat4064g: `datePickerState.isDayValid()` internally checks validity of month and year as well.

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
rajat4064g: Done

PR #2020, opened 2024-02-02
Add string resources for AI-UI module

yschimke:
#### WHAT

Add string resources for AI-UI module

#### WHY

Good citizen

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/ai/ui/src/main/java/com/google/android/horologist/ai/ui/model/FailedResponseUiModel.kt
+++ b/ai/ui/src/main/java/com/google/android/horologist/ai/ui/model/FailedResponseUiModel.kt
@@ -16,9 +16,12 @@
 
 package com.google.android.horologist.ai.ui.model
 
+import android.content.Context
+import com.google.android.horologist.ai.ui.R
+
 public data class FailedResponseUiModel(public val message: String) : PromptOrResponseUiModel {
     public companion object {
-        public val NoCompanion: FailedResponseUiModel =
-            FailedResponseUiModel("No companion app found")
+        public fun NoCompanion(context: Context): FailedResponseUiModel =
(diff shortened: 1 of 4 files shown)


ai/ui/src/main/java/com/google/android/horologist/ai/ui/model/FailedResponseUiModel.kt
```diff
@@ -16,9 +16,12 @@
...
 package com.google.android.horologist.ai.ui.model
 
+import android.content.Context
+import com.google.android.horologist.ai.ui.R
+
 public data class FailedResponseUiModel(public val message: String) : PromptOrResponseUiModel {
     public companion object {
-        public val NoCompanion: FailedResponseUiModel =
-            FailedResponseUiModel("No companion app found")
+        public fun NoCompanion(context: Context): FailedResponseUiModel =
```
luizgrp, line 24: nit: usually it is avoided to pass reference of context to viewmodels in order to prevent further mistakes of [holding an instance of it](https://developer.android.com/topic/libraries/architecture/viewmodel#best-practices) (which is not the case here)

luizgrp (approved):
LGTM

ai/ui/src/main/java/com/google/android/horologist/ai/ui/model/FailedResponseUiModel.kt
yschimke: True, I guess to use this I'd have to implement AndroidViewModel which gets the application context.

yschimke: It's more a helper method, I might give it some thought and follow up just removing this, or moving to a Screen.

PR #2022, opened 2024-02-05
Remove animation from the InstallAppBottomSheetActivity when finishing it

luizgrp:
#### WHAT

Remove animation from the `InstallAppBottomSheetActivity` when finishing it.

#### WHY

Before

https://github.com/google/horologist/assets/878134/73c63747-34b4-4f04-9d1a-d97d76f11a83

After

https://github.com/google/horologist/assets/878134/d379ecb5-ec84-40f4-8d22-f5b548be5573

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke (approved):
LGTM - is this a common problem, worth a comment to some medium post explaining how to fix?

But not blocking.

PR #2023, opened 2024-02-06
Make the buttons in a responsive dialog be responsive.

ssancho14:
* Buttons get wider after the 225.dp breakpoint
* Added dialog activity as a sample to try different dialogs

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


compose-material/src/main/java/com/google/android/horologist/compose/material/ResponsiveDialog.kt
```diff
@@ -128,18 +142,42 @@ public fun ResponsiveDialogContent(
                             verticalAlignment = Alignment.CenterVertically,
                         ) {
                             onCancel?.let {
-                                Button(
-                                    imageVector = Icons.Default.Close,
-                                    contentDescription = cancelButtonContentDescription,
+                                androidx.wear.compose.material.Chip(
```
yschimke, line 145: Can we refactor this into a private ResponsiveButton and keep this Dialog code simpler?

compose-material/src/main/java/com/google/android/horologist/compose/material/ResponsiveDialog.kt
```diff
@@ -115,6 +122,13 @@ public fun ResponsiveDialogContent(
                 }
                 if (onOk != null || onCancel != null) {
                     item {
+                        val width = LocalConfiguration.current.screenWidthDp
+                        val buttonWidth = if (width < 225 || onOk == null || onCancel == null) {
```
BowerSteve, line 126: Add a comment that single buttons are not intended to be responsive?

compose-material/src/main/java/com/google/android/horologist/compose/material/ResponsiveDialog.kt
ssancho14: Done

compose-material/src/main/java/com/google/android/horologist/compose/material/ResponsiveDialog.kt
ssancho14: Done

PR #2042, opened 2024-02-19
Fix phone previews

yschimke:
#### WHY

Don't show up without the debug dependency

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):
Thanks!!

PR #2043, opened 2024-02-19
Screenshots for Chips

yschimke:
#### WHAT

Screenshots of layout for Chips also

#### WHY

Ensure we aren't getting apps rejected

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
Landing so I can unlock some bug fixes I need to test.

PR #2045, opened 2024-02-19
Additional padding at end of SLC

yschimke:
#### WHAT

Default to an additional 10dp at end of `responsive()` to avoid clipping.
Can be overridden. 

#### WHY

Clipping of a tall Card may lead to apps getting rejected.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


BowerSteve (approved):
LGTM

PR #2053, opened 2024-02-20
Adds ResponsiveTimeText

garanj:
#### WHAT

Adds `ResponsiveTimeText` to allow `TimeText` components to be easily set at 2.1% from top.

#### WHY

To help conform with larger devices.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests


compose-material/src/main/java/com/google/android/horologist/compose/material/ResponsiveTimeText.kt
```diff
@@ -0,0 +1,69 @@
...
+import androidx.wear.compose.material.TimeTextDefaults
+import androidx.wear.compose.material.TimeTextDefaults.CurvedTextSeparator
+import androidx.wear.compose.material.TimeTextDefaults.TextSeparator
+import androidx.wear.compose.material.TimeTextDefaults.timeFormat
+
+/**
+ * Provides a [TimeText] component with a responsive padding of 2.1%.
+ */
+@Composable
+public fun ResponsiveTimeText(
```
yschimke, line 38: As a follow up, can you make this the default in AppScaffold?

yschimke (approved):

PR #2055, opened 2024-02-21
Move sample Tile preview to debug folder

luizgrp:
#### WHAT

Move sample Tile preview to debug folder.

#### WHY

Nightly builds are failing with:
```
e: file:///home/runner/work/horologist/horologist/media/sample/src/main/java/com/google/android/horologist/mediasample/data/service/tile/MediaCollectionsTileService.kt:27:28 Unresolved reference: tooling

> Task :media:sample:compileReleaseKotlin FAILED
```

#### Checklist :clipboard:
- [N/A] Add explicit visibility modifier and explicit return types for public declarations
- [x] Run spotless check
- [x] Run tests
- [N/A] Update metalava's signature text files

PR #2056, opened 2024-02-21
Fix Tiles preview dependencies

yschimke:
#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):
Do we want to ship release with that dependency? 
In this case it is just a sample app which we are not going to publish, so it is fine.

This would be the alternative: https://github.com/google/horologist/pull/2055

yschimke:
I think for Compose UI, we ship with this dependency. We assume R8 strips it, I think.

yschimke:
Let's go with yours

PR #2059, opened 2024-02-22
Fix issue with Firebase auth and FirebaseUrlFactory

yschimke:
#### WHAT

Fix issue with Firebase auth and FirebaseUrlFactory. 

#### WHY

Failing when clearing a header.

#### HOW

Treat null as removeHeader

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


luizgrp (approved):
LGTM - failing with metalava

PR #2060, opened 2024-02-23
Add ReEngage prompt

luizgrp:
#### WHAT

Add ReEngage prompt + sample.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/PhoneUiDataLayerHelper.kt
+++ b/datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/PhoneUiDataLayerHelper.kt
@@ -25,6 +25,7 @@ import androidx.activity.result.ActivityResultLauncher
 import androidx.annotation.DrawableRes
 import com.google.android.horologist.annotations.ExperimentalHorologistApi
 import com.google.android.horologist.datalayer.phone.ui.prompt.installapp.InstallAppBottomSheetActivity
+import com.google.android.horologist.datalayer.phone.ui.prompt.reengage.ReEngageBottomSheetActivity
 
 private const val NO_RESULT_REQUESTED_REQUEST_CODE = -1
 
@@ -48,7 +49,7 @@ public class PhoneUiDataLayerHelper {
         bottomMessage: String,
         requestCode: Int = NO_RESULT_REQUESTED_REQUEST_CODE,
     ) {
-        val intent = getInstallPromptIntent(
+        val intent = getInstallAppPromptIntent(
             context = activity,
             appPackageName = appPackageName,
             image = image,
@@ -62,7 +63,7 @@ public class PhoneUiDataLayerHelper {
     }
 
     /**
-     * Returns the [Intent] to display an install prompt to the user.
+     * Returns the [Intent] to display an install app prompt to the user.
      *
      * This can be used in Compose with [rememberLauncherForActivityResult] and
      * [ActivityResultLauncher.launch]:
@@ -76,7 +77,7 @@ public class PhoneUiDataLayerHelper {
      *     }
      * }
      *
-     * launcher.launch(getInstallPromptIntent(/*params*/))
+     * launcher.launch(getInstallAppPromptIntent(/*params*/))
      * ```
      *
      * It can also be used directly in an [ComponentActivity] with
@@ -90,10 +91,10 @@ public class PhoneUiDataLayerHelper {
      *      }
      *  }
      *
-     * launcher.launch(getInstallPromptIntent(/*params*/))
+     * launcher.launch(getInstallAppPromptIntent(/*params*/))
      * ```
      */
-    public fun getInstallPromptIntent(
+    public fun getInstallAppPromptIntent(
         context: Context,
         appPackageName: String,
         @DrawableRes image: Int,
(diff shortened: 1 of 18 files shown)


datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/PhoneUiDataLayerHelper.kt
```diff
@@ -106,4 +107,81 @@ public class PhoneUiDataLayerHelper {
         topMessage = topMessage,
         bottomMessage = bottomMessage,
     )
+
+    /**
+     * Display a re-engage prompt to the user.
```
yschimke, line 112: Is this for non compose users? And the other documentation shows for compose with rememberLauncherForActivityResult?

datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/prompt/reengage/ReEngageBottomSheetActivity.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    private lateinit var phoneDataLayerAppHelper: PhoneDataLayerAppHelper
+    private lateinit var coroutineAppScope: CoroutineScope
+    private lateinit var nodeId: String
+
+    @OptIn(ExperimentalMaterial3Api::class)
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        coroutineAppScope = CoroutineAppScope.getInstance()
```
yschimke, line 60: Is this a common pattern for launchers and compose? What do other libraries do?

yschimke (approved):
This seems really nice, great sample and library.

datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/PhoneUiDataLayerHelper.kt
luizgrp: Not necessarily, both functions can be used with Compose or View system. You can see that `showInstallAppPrompt` is used by the sample app, which is built with compose. 

The other function can also be called without rememberLauncherForActivityResult: https://developer.android.com/training/basics/intents/result#launch - I will add this to the docs

datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/prompt/reengage/ReEngageBottomSheetActivity.kt
luizgrp: Not related to the launchers and compose, it's explained here: https://github.com/google/horologist/pull/2060/files#diff-a09261cf6a3ea4ce576403e04687dc8e1df1c1398b2369747b65d2e93ca17476R117-R121 - please let me know if it needs improvement for clarity

datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/PhoneUiDataLayerHelper.kt
luizgrp: Oh it's already there: 

```
     * It can also be used directly in an [ComponentActivity] with
     * [ComponentActivity.registerForActivityResult]:
```

luizgrp: I might remove these functions as they are not providing much value, rather than showing how to call the getIntent ones without the Activity contracts. They were initially added to handle the result themselves from the placeholder API.

datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/prompt/reengage/ReEngageBottomSheetActivity.kt
yschimke: I had seen the comment, but a dangling CoroutineScope is worrying. I was curious if there are clean alternatives for similar things in other androidx libraries?

If it's unavoidable, or typical, then no concerns at all.

luizgrp: I will check.

Would you be raising the same concern if this scope was being injected into the Activity via Hilt? If so, what would be the difference of what's done here and in: https://github.com/google/horologist/blob/dc6c662e56ade11f18f4cd89ca4ee756b561a5e9/ai/sample/core/src/main/java/com/google/android/horologist/ai/core/di/DataLayerModule.kt#L34-L37

luizgrp: ok so I can add the coroutine scope as param to the function for the prompts, so the user has to provide it, instead of being created by the lib. wdyt?

yschimke: Yeah, I think so.

datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/PhoneUiDataLayerHelper.kt
luizgrp: https://github.com/google/horologist/pull/2068

datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/prompt/reengage/ReEngageBottomSheetActivity.kt
luizgrp: https://github.com/google/horologist/pull/2064

Issue #2061, opened 2024-02-23
Compose-Layout transitive dep on compose-navigation


Tolriq:
I'm not using compose navigation in my apps but a custom solution based on Voyager.

Compose-layout currently have a transitive dependency on compose-navigation that is only used for a link in the javadoc of `AppScaffold`.

Would it be possible to remove that unnecessary dependency in a future release?

yschimke:
Fixed in next release

Tolriq:
Thanks. I guess asking for media-ui-navigation split would be too much to ask.

yschimke:
Can you raise it with a separate issue, it's probably doable, I'd like to confirm with another internal team that uses some of this.  It looks like MediaNavController, NavigationScreens, PlayerLibraryPagerScreen, MediaPlayerScaffold.

Tolriq:
I'm not far enough in the project to exactly know what I'll use from that part of horologist.

But will do if I decide to use the full ui package. Thanks for the follow up.

PR #2062, opened 2024-02-23
Remove navigation and viewmodel deps in compose-layout

yschimke:
#### WHAT

Remove Navigation dependency.

#### WHY

Fixes https://github.com/google/horologist/issues/2061

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/build.gradle.kts
+++ b/compose-layout/build.gradle.kts
@@ -89,20 +89,16 @@ dependencies {
 
     api(libs.wearcompose.material)
     api(libs.wearcompose.foundation)
-    api(libs.wearcompose.navigation)
 
     api(libs.androidx.lifecycle.runtime.compose)
     api(libs.androidx.paging)
 
     implementation(libs.kotlin.stdlib)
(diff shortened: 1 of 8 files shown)


compose-layout/build.gradle.kts
```diff
@@ -89,20 +89,16 @@ dependencies {
 
     api(libs.wearcompose.material)
     api(libs.wearcompose.foundation)
-    api(libs.wearcompose.navigation)
```
luizgrp, line 92: we have this plugin in some modules of the project that can help cleaning up and maintain with checks in CI: 

https://github.com/google/horologist/pull/1266

I see it is disabled atm: https://github.com/google/horologist/issues/1314

luizgrp (approved):

compose-layout/build.gradle.kts
yschimke: Good idea, to avoid this exploding, I'll update the compose subset now.

PR #2063, opened 2024-02-23
Add Sign-in prompt

luizgrp:
#### WHAT

Add Sign-in prompt + sample.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/prompt/signin/CoroutineScopeHolder.kt
+++ b/datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/prompt/signin/CoroutineScopeHolder.kt
@@ -1,0 +1,23 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.datalayer.phone.ui.prompt.signin
+
+import kotlinx.coroutines.CoroutineScope
+
+internal object CoroutineScopeHolder {
+    lateinit var coroutineScope: CoroutineScope
+}
--- a/datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/inappprompts/signin/SignInPromptDemoViewModel.kt
+++ b/datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/inappprompts/signin/SignInPromptDemoViewModel.kt
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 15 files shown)


datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/main/MainScreen.kt
```diff
@@ -74,7 +74,10 @@ fun MainScreen(
                     InstallAppPromptDemo2Screen()
                 }
                 composable(route = Screen.ReEngagePromptDemoScreen.route) {
-                    ReEngagePromptDemoScreen()
+                    SignInPromptDemoScreen()
```
yschimke, line 77:
```suggestion
                    ReEngagePromptDemoScreen()
```

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/inappprompts/signin/SignInPromptDemoViewModel.kt
```diff
@@ -0,0 +1,99 @@
...
+        val signInPrompt: SignInPrompt,
+    ) : ViewModel() {
+
+        private var initializeCalled = false
+
+        private val _uiState =
+            MutableStateFlow<SignInPromptDemoScreenState>(SignInPromptDemoScreenState.Idle)
+        public val uiState: StateFlow<SignInPromptDemoScreenState> = _uiState
+
+        @MainThread
```
yschimke, line 45: Curious? why MainThread on these? Avoiding locking on initializeCalled?

If so, is it worth a comment on initializeCalled, "always accessed on Main thread"

datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/prompt/signin/CoroutineScopeHolder.kt
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.datalayer.phone.ui.prompt.signin
+
+import kotlinx.coroutines.CoroutineScope
+
+internal object CoroutineScopeHolder {
```
yschimke, line 21: I'm still curious what this looks like in other similar libraries, but will take a look myself.

Not an issue in practice, or principle, I'm just curious about it.

yschimke (approved):

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/inappprompts/signin/SignInPromptDemoViewModel.kt
luizgrp: Wouldn't the annotation be enough to indicate that? From the MainThread annotation kdocs:

> Denotes that the annotated method should only be called on the main thread. If the annotated element is a class, then all methods in the class should be called on the main thread.

Are you concerned the variable could be modified, internally within the class, by a different thread? I'd say if the variable is changed outside this method, it is already a red flag.

In the sample code from [DAC](https://developer.android.com/topic/architecture/ui-layer/state-production#initializing-state-production) there is a comment:
> This function is idempotent provided it is only called from the UI thread.

Let me know if you think we should add that.

datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/prompt/signin/CoroutineScopeHolder.kt
luizgrp: please let me know your findings 👍

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/main/MainScreen.kt
luizgrp: thanks!

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/inappprompts/signin/SignInPromptDemoViewModel.kt
luizgrp: Merging but will come back to these open comments in the next iterations.

yschimke: Not an issue. More as part of the review, I was curious why it was MainThread, then reading the code realised it was the unsynchronized field.

But your call, nothing needed.

PR #2069, opened 2024-02-27
Add functions to check if prompts should be displayed

luizgrp:
#### WHAT

Add functions to check if prompts should be displayed.

#### WHY

Make the API more opinionated, while still allowing customisation.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/prompt/reengage/ReEngagePrompt.kt
+++ b/datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/prompt/reengage/ReEngagePrompt.kt
@@ -23,14 +23,27 @@ import androidx.activity.compose.rememberLauncherForActivityResult
 import androidx.activity.result.ActivityResultLauncher
 import androidx.annotation.DrawableRes
 import com.google.android.horologist.annotations.ExperimentalHorologistApi
+import com.google.android.horologist.data.apphelper.AppHelperNodeStatus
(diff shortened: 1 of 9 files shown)


datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/prompt/reengage/ReEngagePrompt.kt
```diff
@@ -23,15 +23,28 @@ import androidx.activity.compose.rememberLauncherForActivityResult
...
     init {
         CoroutineScopeHolder.coroutineScope = coroutineScope
     }
 
+    /**
+     * Returns a [AppHelperNodeStatus] that meets the criteria to show this prompt, otherwise
+     * returns null.
+     */
+    public suspend fun shouldDisplayPrompt(): AppHelperNodeStatus? =
+        phoneDataLayerAppHelper.connectedNodes().firstOrNull { it.appInstalled }
```
yschimke, line 46: Should this check it wasn't launched in some window?

yschimke (approved):

datalayer/phone-ui/src/main/java/com/google/android/horologist/datalayer/phone/ui/prompt/reengage/ReEngagePrompt.kt
luizgrp: If you mean a time window, we might add it in the future 👍

PR #2071, opened 2024-02-27
new rememberResponsiveColumnState API

yschimke:
#### WHAT

New Responsive SLC API, building on content type.

image

#### WHY

Want to get cleaner layout.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnDefaults.kt
```diff
@@ -253,4 +257,39 @@ public object ScalingLazyColumnDefaults {
...
+
+    enum class ItemType(val topPaddingDp: Float, val bottomPaddingDp: Float, val paddingCorrection: Dp = 0.dp) {
+        Card(Padding21Pct, Padding31Pct),
+        Chip(Padding21Pct, Padding31Pct),
+        CompactChip(topPaddingDp = Padding12Pct, bottomPaddingDp = Padding20Pct, paddingCorrection = (-8).dp),
+        Icon(Padding12Pct, Padding21Pct),
+        MultiButton(Padding21Pct, Padding20Pct),
+        SingleButton(Padding12Pct, Padding20Pct),
+        Text(Padding21Pct, Padding31Pct),
+        Unspecified(Padding21Pct, Padding31Pct),
```
BowerSteve, line 275: Can we leave these as zero as the plan was to use the formula when unspecified, as previously? Looks like the previous formula has been deleted.

yschimke: Yep.

BowerSteve (approved):
Approved pending update to the ItemType.Unspecified handling.

PR #2073, opened 2024-02-28
Remove AppInstallationStatusNodeType.UNKNOWN

luizgrp:
#### WHAT

Remove `AppInstallationStatusNodeType.UNKNOWN`.

#### WHY

It should not need based on current code: items in `allInstalledNodes` are either in `installedPhoneNodes` or `installedWatchNodes`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/core/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
+++ b/datalayer/core/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
@@ -77,8 +77,7 @@ abstract class DataLayerAppHelper(
             val appInstallationStatus = if (allInstalledNodes.contains(it.id)) {
                 val nodeType = when (it.id) {
                     in installedPhoneNodes -> AppInstallationStatusNodeType.PHONE
-                    in installedWatchNodes -> AppInstallationStatusNodeType.WATCH
-                    else -> AppInstallationStatusNodeType.UNKNOWN
+                    else -> AppInstallationStatusNodeType.WATCH
                 }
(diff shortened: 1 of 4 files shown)


yschimke (approved):

datalayer/core/src/main/java/com/google/android/horologist/data/apphelper/DataLayerAppHelper.kt
```diff
@@ -77,8 +77,7 @@ abstract class DataLayerAppHelper(
             val appInstallationStatus = if (allInstalledNodes.contains(it.id)) {
                 val nodeType = when (it.id) {
                     in installedPhoneNodes -> AppInstallationStatusNodeType.PHONE
-                    in installedWatchNodes -> AppInstallationStatusNodeType.WATCH
-                    else -> AppInstallationStatusNodeType.UNKNOWN
+                    else -> AppInstallationStatusNodeType.WATCH
```
yschimke, line 80: This is only true provided the capabilities are correctly installed?

luizgrp: That's why I thought when introduced the UNKNOWN value. However, after reviewing the code:

- Line 74: allInstalledNodes = installedPhoneNodes + installedWatchNodes
- Line 77: only enter this block if id is in allInstalledNodes
- Line 80: if id is not in installedPhoneNodes, then it can only be in installedWatchNodes, as allInstalledNodes = installedPhoneNodes + installedWatchNodes

Please review if I'm missing anything 🤔

PR #2076, opened 2024-02-28
Add performAction function to SignInPrompt

luizgrp:
#### WHAT

Add `performAction` function to `SignInPrompt`.
Add sample.

#### WHY

So it can be used with custom prompts.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


datalayer/sample/phone/src/main/res/values/strings.xml
```diff
@@ -28,6 +28,7 @@
     <string name="menu_screen_reengage_demo_item">Re-Engage</string>
     <string name="menu_screen_signin_demo_item">Sign-in</string>
     <string name="menu_screen_install_app_custom_demo_item">Install app (custom)</string>
```
kul3r4, line 30: (custom)-> (custom prompt)

kul3r4 (approved):

yschimke (approved):

PR #2078, opened 2024-02-29
Add performAction function to ReEngagePrompt

luizgrp:
#### WHAT

Add `performAction` function to `ReEngagePrompt`.
Add sample.

#### WHY

So it can be used with custom prompts.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/inappprompts/custom/reengage/ReEngageCustomPromptDemoScreen.kt
+++ b/datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/inappprompts/custom/reengage/ReEngageCustomPromptDemoScreen.kt
@@ -1,0 +1,227 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@file:OptIn(ExperimentalMaterial3Api::class)
+
+package com.google.android.horologist.datalayer.sample.screens.inappprompts.custom.reengage
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Watch
+import androidx.compose.material3.Button
+import androidx.compose.material3.CircularProgressIndicator
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.Icon
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.ModalBottomSheet
+import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
+import androidx.compose.material3.rememberModalBottomSheetState
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.SideEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.ui.Alignment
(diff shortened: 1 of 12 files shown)


datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/inappprompts/custom/reengage/ReEngageCustomPromptDemoScreen.kt
```diff
@@ -0,0 +1,227 @@
...
+@Composable
+fun ReEngageCustomPromptDemoScreen(
+    modifier: Modifier = Modifier,
+    viewModel: ReEngageCustomPromptDemoViewModel = hiltViewModel(),
+) {
+    val state by viewModel.uiState.collectAsStateWithLifecycle()
+
+    if (state == ReEngageCustomPromptDemoScreenState.Idle) {
+        SideEffect {
+            viewModel.initialize()
```
yschimke, line 64: I understand this is correct and our guidance, it does create a lot of visual noise.

Do we know why apps like Now in Android don't follow the guidance?

https://github.com/android/nowinandroid/blob/76d0e430f7d350b1707bb19c8e007bc5679ac729/feature/bookmarks/src/main/kotlin/com/google/samples/apps/nowinandroid/feature/bookmarks/BookmarksViewModel.kt

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/inappprompts/custom/reengage/ReEngageCustomPromptDemoViewModel.kt
```diff
@@ -0,0 +1,96 @@
...
+
+@HiltViewModel
+class ReEngageCustomPromptDemoViewModel
+    @Inject
+    constructor(
+        private val phoneDataLayerAppHelper: PhoneDataLayerAppHelper,
+        private val reEngagePrompt: ReEngagePrompt,
+    ) : ViewModel() {
+
+        private var initializeCalled = false
```
yschimke, line 38: For a follow up. Are there ways to reduce the repetition of this initialise phase?

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/inappprompts/custom/reengage/ReEngageCustomPromptDemoScreen.kt
luizgrp: I don't know, what about asking in https://github.com/android/nowinandroid/discussions ?

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/inappprompts/custom/reengage/ReEngageCustomPromptDemoViewModel.kt
luizgrp: This is the repetition:

Composable

```kotlin
        SideEffect {
            viewModel.initialize()
        }
```

ViewModel

```kotlin
        private var initializeCalled = false

        @MainThread
        fun initialize() {
            if (initializeCalled) return
            initializeCalled = true

            // pipeline initialization
        }
```

The repetition could be reduced having a base class for VM, but that's not scalable. I don't see many lines saved if composition approach is taken instead of inheritance.

For the composable, there is only 3 lines, that could be inlined.

Any thoughts?

yschimke: Nothing for us. If the guidance is good, the library should build this in.

But it seems very optional.

datalayer/sample/phone/src/main/java/com/google/android/horologist/datalayer/sample/screens/inappprompts/custom/reengage/ReEngageCustomPromptDemoScreen.kt
luizgrp: https://github.com/android/nowinandroid/discussions/1235

luizgrp: The answer seems to be in [DAC](https://developer.android.com/topic/architecture/ui-layer/state-production#initializing-state-production):

> Flow APIs allow for this with the [started](https://developer.android.com/topic/architecture/ui-layer/31) argument in the [stateIn](https://developer.android.com/topic/architecture/ui-layer/17) method. 

Which seems to be what NiA is using.

Then it continues:

> In the cases where this is inapplicable, define an [idempotent](https://en.wikipedia.org/wiki/Idempotence) initialize() function to explicitly start the state production pipeline as shown in the following snippet:

So we should re-evaluate our cases and see when the stateIn + started WhileSubscribed is applicable.

luizgrp: https://github.com/google/horologist/issues/2080

PR #2090, opened 2024-03-05
Update wear compose scrollaway with timeout changes

rajat4064g:
Adding animations to scrollAway modifier.

TimeText will appear after 3s when scrolling stops in a list of items.

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScaffoldState.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScaffoldState.kt
@@ -50,7 +50,7 @@ internal class ScaffoldState {
             Box(
                 modifier = Modifier
                     .fillMaxSize()
-                    .scrollAway {
+                    .scrollAway(true) {
                         scrollState ?: ScrollState(0)
                     },
             ) {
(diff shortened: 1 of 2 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScaffoldState.kt
```diff
@@ -50,7 +50,7 @@ internal class ScaffoldState {
             Box(
                 modifier = Modifier
                     .fillMaxSize()
-                    .scrollAway {
+                    .scrollAway(true) {
```
yschimke, line 53: Can we use named params?  

Is this meant to be defaulted on? Would false be a better default? Or move this up to a param for AppScaffold/ScreenScaffold?

yschimke (approved):
LGTM - but I don't think we should change the default at least yet.

yschimke:
@rajat4064g let me know if you are waiting on me for anything.

rajat4064g:
> @rajat4064g let me know if you are waiting on me for anything.

There are more changes that need to be dropped in this branch regarding syncing the timer for TimeOut animations between material PositionIndicator and new ScrollAway. Will merge this PR when that is finalised.

yschimke:
Going to close as it's got merge conflicts, reopen once it's ready for review.

PR #2095, opened 2024-03-06
Make 24h TimePicker more responsive on larger screens.

ssancho14:
#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


sample/src/main/java/com/google/android/horologist/scratch/PickerActivity.kt
```diff
@@ -0,0 +1,124 @@
...
+
+        setContent {
+            WearPickerApp()
+        }
+    }
+}
+
+@Composable
+fun WearPickerApp() {
+    val sizes = listOf(190, 200, 210, 220, 230, 240, 250)
```
yschimke, line 85:
```suggestion
    val sizes = listOf(190, 200, 210, 220, 230)
```

yschimke (approved):
Waiting for screenshots

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
```diff
@@ -138,13 +143,33 @@ public fun TimePicker(
...
+    val isLargeScreen = LocalConfiguration.current.screenWidthDp > 225
     val textStyle = with(LocalDensity.current) {
-        if (showSeconds) {
-            fontScaleIndependent(MaterialTheme.typography.display3)
-        } else {
-            fontScaleIndependent(MaterialTheme.typography.display1)
-        }
+        fontScaleIndependent(
+            when {
+                !showSeconds -> MaterialTheme.typography.display1
```
BowerSteve, line 150: Not sure the display1, display2, display3 match the spec...

BowerSteve (commented):
TimePickerWith12HourClock needs to be updated too

composables/src/main/java/com/google/android/horologist/composables/TimePicker.kt
ssancho14: I see no spec for hour + minutes picker... display1 is what it was before. Happy to use something else if we have a better alternative.

Well, there is no display4 on our theme, and current constants do match the sp sizes specified in the spec.

sample/src/main/java/com/google/android/horologist/scratch/PickerActivity.kt
ssancho14: Done.

yschimke (approved):

PR #2115, opened 2024-03-13
Add tile sync and sample tile

yschimke:
#### WHAT

Add tile sync and sample tile.

image

#### WHY

For checking tile registration.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


kul3r4 (approved):
thanks!

PR #2122, opened 2024-03-14
Support RNG screenshots

yschimke:
#### WHAT

Copy the code from ComposeStarter, put it in a library form that can be used widely across samples for quick and effective screenshot testing.

#### WHY

Consistent screenshot testing.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
@garanj Can you post review? I'd like to land and start using the preferred screenshot option here.

PR #2125, opened 2024-03-15
Makes progress prediction only happen when the lifecycle is resumed

fstanis:
#### WHAT
Makes `ProgressStateHolder` suspend when the view is paused.

#### WHY
To reduce CPU usage when the composable is backgrounded or otherwise not visible.

#### HOW
Via `lifecycleOwner.repeatOnLifecycle`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media/ui/src/main/java/com/google/android/horologist/media/ui/state/ProgressStateHolder.kt
+++ b/media/ui/src/main/java/com/google/android/horologist/media/ui/state/ProgressStateHolder.kt
@@ -21,9 +21,12 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.State
 import androidx.compose.runtime.derivedStateOf
-import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.mutableFloatStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.withFrameMillis
+import androidx.compose.ui.platform.LocalLifecycleOwner
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.repeatOnLifecycle
 import com.google.android.horologist.annotations.ExperimentalHorologistApi
 import com.google.android.horologist.media.model.TimestampProvider
 import com.google.android.horologist.media.ui.animation.PlaybackProgressAnimation.PLAYBACK_PROGRESS_ANIMATION_SPEC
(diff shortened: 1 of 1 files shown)


media/ui/src/main/java/com/google/android/horologist/media/ui/state/ProgressStateHolder.kt
```diff
@@ -74,13 +77,21 @@ internal class ProgressStateHolder(
...
+            }
+            LaunchedEffect(trackPositionUiModel, lifecycleOwner) {
+                // Prediction only happens when the UI is visible.
+                lifecycleOwner.repeatOnLifecycle(Lifecycle.State.RESUMED) {
+                    val percent = trackPositionUiModel.getCurrentPercent(timestampProvider.getTimestamp())
+                    stateHolder.setProgress(percent, trackPositionUiModel.shouldAnimate)
+                    if (trackPositionUiModel is TrackPositionUiModel.Predictive) {
+                        stateHolder.predictProgress(trackPositionUiModel.predictor::predictPercent)
+                    }
+                    stateHolder.setProgress(0f, false)
```
yschimke, line 94: How does this ever get hit? should it be in an else?

yschimke:
Nice catch!

yschimke:
Test failures

```
> Task :media:ui:testDebugUnitTest

com.google.android.horologist.media.ui.components.PlayPauseProgressButtonTest > givenMediaProgress_thenProgressIsCorrect FAILED
    java.lang.AssertionError at PlayPauseProgressButtonTest.kt:117

com.google.android.horologist.media.ui.components.PodcastControlButtonsWithProgressTest > givenPercentParam_thenProgressBarIsDisplayed FAILED
    java.lang.AssertionError at PodcastControlButtonsWithProgressTest.kt:232
```

media/ui/src/main/java/com/google/android/horologist/media/ui/state/ProgressStateHolder.kt
fstanis: The idea was for it to happen when the coroutine is cancelled (i.e. `isActive` is no longer true in the while loop), but I'm not sure if this is the right approach.

fstanis:
> What will the UI like like when onResume happens? onResume can happen when the UI is already visible, so is there a visible progress jump?

This would result in the progress bar being hidden all together when the UI is paused and it re-appearing when it's resumed, so there wouldn't be a progress jump (but it would disappear / reappear).

media/ui/src/main/java/com/google/android/horologist/media/ui/state/ProgressStateHolder.kt
yschimke: I think isActive is really only used in tight CPU (non suspending) loops.

My understanding, is that cancellation will kick in at any suspension points. So I'm doubtful this can ever hit.

Worth confirming.

fstanis: You're absolutely right - I neglected the fact the body of the loop is also suspending, which will throw when cancelled. So this code would only be reachable in e.g. a `finally {}` block. However, it may be unneccessary to include anyway, I'm testing this right now to confirm.

schmiphi (approved):

PR #2126, opened 2024-03-15
Fix responsive message padding

yschimke:
#### WHAT

Fix responsive message padding

#### WHY

Too cramped

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
fixing screenshots in the background

yschimke:
I suspect this is not 100% correct. There is already padding.

```
                        Row(
                            Modifier
                                .fillMaxWidth()
                                .padding(
                                    top = if (content != null || message != null) 12.dp else 0.dp,
                                ),
```

BowerSteve (commented):
Also remove '|| message != null' from line 149?

BowerSteve (approved):
LGTM

PR #2127, opened 2024-03-15
Wire Tile prompt

kul3r4:
#### WHAT
Wire Tile prompt on the phone with the watch

#### WHY
To make the editor Tile settings to show

#### HOW
Call trackInstallTile when the app starts

#### Checklist :clipboard:
- [x] Run spotless check
- [ ] Run tests

### DIFF
--- a/datalayer/sample/wear/src/main/java/com/google/android/horologist/datalayer/sample/TileSync.kt
+++ b/datalayer/sample/wear/src/main/java/com/google/android/horologist/datalayer/sample/TileSync.kt
@@ -27,10 +27,10 @@ import kotlinx.coroutines.asExecutor
 import kotlinx.coroutines.tasks.await
 
 class TileSync(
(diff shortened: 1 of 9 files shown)


datalayer/sample/wear/src/main/java/com/google/android/horologist/datalayer/sample/TileSync.kt
```diff
@@ -27,10 +27,10 @@ import kotlinx.coroutines.asExecutor
 import kotlinx.coroutines.tasks.await
 
 class TileSync(
-    val registry: WearDataLayerRegistry,
-    val wearAppHelper: WearDataLayerAppHelper,
+    private val registry: WearDataLayerRegistry,
```
yschimke, line 30: Thanks!

yschimke (approved):
Nice cleanup.

PR #2132, opened 2024-03-18
Add Auth device screenshots

yschimke:
#### WHAT

Add Auth device screenshots

#### WHY

Confirm current layouts before adjustments

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
tests only, won't wait for a review on this.

PR #2148, opened 2024-03-19
Make marquee text maintain talkback focus during marquee animation.

clickxu:
#### WHAT
Make marquee text maintain talkback focus during marquee animation.

#### WHY
Currently, when Talkback focuses on the marquee title, the marquee loses focus after completing its first round of animation.

#### HOW
Add `Modifier.semantics { focused = true }` to the marquee text.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
+++ b/composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
@@ -40,6 +40,8 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.CompositingStrategy
 import androidx.compose.ui.graphics.graphicsLayer
 import androidx.compose.ui.layout.layout
+import androidx.compose.ui.semantics.focused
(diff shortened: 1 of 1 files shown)


composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
```diff
@@ -100,6 +102,7 @@ public fun MarqueeText(
         inlineContent = inlineContent,
         modifier = modifier
             .then(controller.outsideMarqueeModifier)
+            .semantics { focused = true }
```
yschimke, line 105: Does this affect RSB at all?

yschimke (approved):

composables/src/main/java/com/google/android/horologist/composables/MarqueeText.kt
clickxu: No, this only affects the talkback green frame focus. RSB works as it is, normally controls the volume.

PR #2175, opened 2024-04-03
Updates a11y for VolumeScreen

fstanis:
#### WHAT
Makes `VolumeScreen` announce the current volume every time it changes.

#### WHY
To comply with a11y requirements around volume and to match phone behavior.

#### HOW
By setting a [`LiveRegionMode`](https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/LiveRegionMode) and disabling default `Stepper` range semantics.

#### Checklist :clipboard:
- [X] Run spotless check
- [X] Run tests

### DIFF
--- a/media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
+++ b/media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
@@ -17,8 +17,6 @@
 package com.google.android.horologist.audio.ui
 
 import android.media.AudioManager
-import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.material.icons.Icons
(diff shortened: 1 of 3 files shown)


media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
```diff
@@ -180,9 +187,18 @@ internal fun VolumeScreen(
     showVolumeIndicator: Boolean = true,
     volumeColor: Color = MaterialTheme.colors.secondary,
 ) {
-    Box(modifier = modifier.fillMaxSize())
     val volumeState = volume()
+    val volumePercent = (100f * volumeState.current / volumeState.max).roundToInt()
+    val volumeDescription = if (volumeState.current == 0) {
+        stringResource(id = R.string.horologist_volume_screen_volume_zero)
+    } else {
+        stringResource(id = R.string.horologist_volume_screen_volume_percent, volumePercent)
```
yschimke, line 195: Is it a convention, or clear current spec to use % for volume?

yschimke (approved):

PR #2177, opened 2024-04-04
Player UI state

laiyichin:
#### WHAT
Add an optional field to denote if the title should show with placeholder.

#### WHY
In addition to the player state, give the client an option to show the placeholder pill in place of title when title is not known yet.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media/ui/src/main/java/com/google/android/horologist/media/ui/state/PlayerUiState.kt
+++ b/media/ui/src/main/java/com/google/android/horologist/media/ui/state/PlayerUiState.kt
@@ -45,6 +45,7 @@ import com.google.android.horologist.media.ui.state.model.TrackPositionUiModel
  * @param media current [MediaUiModel]
  * @param seekBackButtonIncrement increment when seeking back.
  * @param seekForwardButtonIncrement increment when seeking forward.
+ * @param titlePlaceholder denotes whether the title should be using the placeholder.
  * @param connected is the player screen connected.
  */
 @ExperimentalHorologistApi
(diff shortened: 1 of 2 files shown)


yschimke (commented):
Why is this not just based on whether media model and title is set?

laiyichin:
> Why is this not just based on whether media model and title is set?

Adding this state allows to accommodate use cases when media apps do not provide metadata (e.g. a player playing in incognito mode in chrome).

media/ui/src/main/java/com/google/android/horologist/media/ui/state/PlayerUiState.kt
```diff
@@ -65,6 +66,7 @@ public data class PlayerUiState(
     val seekBackButtonIncrement: SeekButtonIncrement = SeekButtonIncrement.Unknown,
     val seekForwardButtonIncrement: SeekButtonIncrement = SeekButtonIncrement.Unknown,
     val connected: Boolean,
+    val showTitlePlaceholder: Boolean = false,
```
fstanis, line 69: IMO, this should probably be part of `MediaUiModel` since that class deals with the title, ideally `MediaUiModel` would be a `sealed class` with `object LoadingMediaUiModel : MediaUiModel` being one subclass. But I'm OK with this approach too.

fstanis (approved):

laiyichin:
Abandoning this PR in favour of making `MediaUiModel` a sealed class with a `Loading` object. https://github.com/google/horologist/pull/2179

PR #2179, opened 2024-04-07
Make MediaUiModel sealed for loading and loaded data object / class

laiyichin:
#### WHAT
Make MediaUiModel a sealed class, and move the original fields into the sub data class.
Also add a loading data object for this sealed class.

#### WHY
This allows us to distinguish the state between known media metada and loading media metadata

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media/sample/src/main/java/com/google/android/horologist/mediasample/ui/player/UampMediaPlayerScreen.kt
```diff
@@ -49,7 +50,12 @@ fun UampMediaPlayerScreen(
     PlayerScreen(
         modifier = modifier,
         background = {
-            val artworkUri = it.media?.artworkUri
+            val metadata = it.media
+            val artworkUri = if (metadata is MediaUiModel.Ready) {
```
fstanis, line 54: Nit: I think you can use `(metadata as? MediaUiMedia.Ready)?.artworkUri` to replace this `if`.

media/ui/src/main/java/com/google/android/horologist/media/ui/components/animated/AnimatedMediaInfoDisplay.kt
```diff
@@ -35,7 +35,7 @@ public fun AnimatedMediaInfoDisplay(
 ) {
     if (loading) {
         LoadingMediaDisplay(modifier)
-    } else if (media != null) {
+    } else if (media != null && media is MediaUiModel.Ready) {
```
fstanis, line 38: (same here)

fstanis (approved):

PR #2182, opened 2024-04-09
Cleanup a11y tests

yschimke:
#### WHAT

Cleanup a11y tests

#### WHY

Complete pay down of tech debt.  Just use a single screenshot approach.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


composables/src/test/snapshots/images/com.google.android.horologist.composables_DatePickerA11yTest_interactionTest_3.png
yschimke, line 1: To test manually.

PR #2191, opened 2024-04-12
Update UAMPEntity screen to use AlertDialog

kul3r4:
#### WHAT
Update the code to use Horologist AlertDialog instead of default Compose Alert/Dialog

#### WHY
Makes code much shorter

#### HOW
Use the Horologist component

#### Checklist :clipboard:
- [x] Run spotless check
- [ ] Run tests


yschimke (approved):
Nice cleanup

PR #2215, opened 2024-04-28
[0.5.x] Remove unused non-Wear navigation dependencies

cbeyls:
#### WHAT

The compose-layout library in the 0.5.x branch depends on `androidx.navigation:navigation-compose` and `androidx.navigation:navigation-ui-ktx`, two unused phone libraries. This pull request removes these dependencies.

#### WHY

These libraries are designed for Android phones and not Wear devices, which should use `androidx.wear.compose.*` libraries instead. In particular, `androidx.navigation:navigation-ui-ktx` depends on both **AppCompat** and **Material Components**, which bring along a lot of **unwanted resources and dependencies** that the shrinker is unable to discard. As a result, a Wear application apk file grows by about 1MB of unnecessary code.

The horologist code does not reference these libraries so the removal has no impact on it.

#### HOW

Remove implementation declarations in `buid.gradle.kts`.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -111,7 +111,6 @@ androidx-mediarouter = "androidx.mediarouter:mediarouter:1.7.0"
 androidx-metrics-performance = "androidx.metrics:metrics-performance:1.0.0-beta01"
 androidx-navigation-compose = { module = "androidx.navigation:navigation-compose", version.ref = "androidxNavigation" }
 androidx-navigation-testing = { module = "androidx.navigation:navigation-testing", version.ref = "androidxNavigation" }
-androidx-navigation-ui-ktx = { module = "androidx.navigation:navigation-ui-ktx", version.ref = "androidxNavigation" }
 androidx-paging = "androidx.paging:paging-compose:3.2.1"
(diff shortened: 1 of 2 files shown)


gradle/libs.versions.toml
```diff
@@ -111,7 +111,6 @@ androidx-mediarouter = "androidx.mediarouter:mediarouter:1.7.0"
 androidx-metrics-performance = "androidx.metrics:metrics-performance:1.0.0-beta01"
 androidx-navigation-compose = { module = "androidx.navigation:navigation-compose", version.ref = "androidxNavigation" }
```
yschimke, line 112: Can you remove this as well?

yschimke:
Thanks

I'll make the same change on main on monday, unless you want to. And cut a 0.5.x release as I assume you want that.

gradle/libs.versions.toml
cbeyls: This dependency is used in the phone sample app so I didn't remove it. It has no impact on the library itself.

yschimke:
Released in https://github.com/google/horologist/releases/tag/v0.5.27

PR #2217, opened 2024-04-30
Add Animated Chips and cards demos

Kpeved:
A couple of demos were added for showing animation transitions in Card and Chips

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


BowerSteve:
Let's see if we can add AnimatedLabel into Horologist as a usable utility composable, rather than just a sample.

compose-material/src/main/java/com/google/android/horologist/compose/material/AnimatedLabel.kt
```diff
@@ -0,0 +1,69 @@
...
+ */
+@Composable
+public fun AnimatedLabel(
+    label: String?,
+    content: @Composable AnimatedContentScope.(targetState: String) -> Unit,
+) {
+    var previousLabel by remember {
+        mutableStateOf(label ?: "")
+    }
+    label?.also { previousLabel = it }
```
yschimke, line 50: Is this safe if compose happens to do an additional recompose? Won't this overwrite previous with the current?

compose-material/src/main/java/com/google/android/horologist/compose/material/AnimatedLabel.kt
```diff
@@ -0,0 +1,69 @@
...
+        visible = label != null,
+        enter = fadeIn(tween(250, delayMillis = 50)) +
+            expandVertically(tween(200)),
+        exit = fadeOut(tween(250)) +
+            shrinkVertically(tween(200, delayMillis = 50)),
+    ) {
+        AnimatedContent(
+            targetState = previousLabel,
+            transitionSpec = {
+                fadeIn(tween(durationMillis = 300, delayMillis = 150))
```
yschimke, line 62: Should the size animate as well? It feels a bit jumpy, but I don't know what the spec is.

yschimke (approved):
Some questions - but nothing blocking.

compose-material/src/main/java/com/google/android/horologist/compose/material/AnimatedLabel.kt
Kpeved: Yes. When the `label` changes - `previousLabel` state changes as well. This doesn't trigger additional recomposition, because we pass `previousLabel` as a state to `AnimatedContent`.

compose-material/src/main/java/com/google/android/horologist/compose/material/AnimatedLabel.kt
Kpeved: It's not the same but very similar to material spec, which is currently used by Views .
For size transitions (shrink/ expand) we use another component on top - `AnimatedVisibility`

PR #2220, opened 2024-05-01
Fix ScalingLazyColumn Rotary issue

yschimke:
#### WHAT

Fix ScalingLazyColumn Rotary issue

#### WHY

Double applying modifier, and also snap/scroll mixup.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
@@ -15,12 +15,10 @@
  */
 
 @file:Suppress("ObjectLiteralToLambda")
-@file:OptIn(ExperimentalWearFoundationApi::class)
 
 package com.google.android.horologist.compose.layout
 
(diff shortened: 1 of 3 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
```diff
@@ -200,44 +195,37 @@ public fun ScalingLazyColumn(
     modifier: Modifier = Modifier,
     content: ScalingLazyListScope.() -> Unit,
 ) {
-    val focusRequester = rememberActiveFocusRequester()
+    val (behavior, fling) = when (columnState.rotaryMode) {
```
BowerSteve, line 198: Nit: I would rename these rotaryBehavior, flingBehavior to avoid confusion.

BowerSteve (approved):
Looks good, just one naming suggestion.

Kpeved (approved):

Issue #2226, opened 2024-05-07
[FR] Expose beyondViewportPageCount to PagerScreen


Tolriq:
As per title it would be nice if `beyondViewportPageCount` was exposed via `PagerScreen` 

On a 2 page pager, it's more efficient to keep the second page loaded then constantly rebuild on page change.

yschimke:
@Tolriq  care to review?

PR #2230, opened 2024-05-08
Add beyondViewportPageCount and friends to PagerScreen

yschimke:
#### WHAT

Expose params that don't conflict with the overall PagerScreen concept.

#### WHY

https://github.com/google/horologist/issues/2226

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/pager/PagerScreen.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/pager/PagerScreen.kt
@@ -19,9 +19,11 @@
 package com.google.android.horologist.compose.pager
 
 import androidx.compose.foundation.ExperimentalFoundationApi
+import androidx.compose.foundation.gestures.Orientation
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.pager.HorizontalPager
+import androidx.compose.foundation.pager.PagerDefaults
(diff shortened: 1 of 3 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/pager/PagerScreen.kt
```diff
@@ -46,9 +49,17 @@ import com.google.android.horologist.compose.layout.PagerScaffold
  */
 @Composable
 public fun PagerScreen(
-    modifier: Modifier = Modifier,
     state: PagerState,
-    content: @Composable ((Int) -> Unit),
+    modifier: Modifier = Modifier,
+    beyondViewportPageCount: Int = 1,
```
Tolriq, line 54: I would keep the same default at 0.

compose-layout/src/main/java/com/google/android/horologist/compose/pager/VerticalPagerScreen.kt
```diff
@@ -45,6 +48,14 @@ import com.google.android.horologist.compose.rotaryinput.rotaryWithPager
 public fun VerticalPagerScreen(
     state: PagerState,
     modifier: Modifier = Modifier,
+    beyondViewportPageCount: Int = 1,
```
Tolriq, line 51: I would keep the same default at 0.

Tolriq (approved):

PR #2232, opened 2024-05-15
Update docs

kul3r4:
#### WHAT

- Moves docs for logging into its own doc
- removes PlaybackRules which have been removed in code

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


docs/media3-backend.md
```diff
@@ -15,51 +15,6 @@ engine for Wear OS media apps.
 All functionality is demonstrated in the `media-sample` app, and as such is not described here
 with extensive code samples.
 
-### Bluetooth Headphone Connections
```
yschimke, line 18: Oops.

Should we refer to the unsuitable listener? or do we?

yschimke (approved):

docs/media3-backend.md
kul3r4: good point, we are doing that directly in the sample when we build the Exoplayer instance. I added a note in https://github.com/google/horologist/pull/2232/files#diff-3a06dd609aaa229f82ba9c4e7e00bb4f94d421268e2a6b823bc63302f8fa449aR18

PR #2242, opened 2024-05-28
Fix button padding

yschimke:
#### WHAT

Inconsequential update to the spec

#### WHY

Tiny diff from the spec

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
Trivial, not waiting for review.

Issue #251, opened 2022-06-20
Add support for WEBP or PNG with transparency to Tiles image helpers


yschimke:
The current conversion code for bitmaps and canvas elements to Tile ImageResources, uses RGB_565.

https://github.com/google/horologist/blob/40801e6e64b55b7e0caeaba96151e8577f399afc/tiles/src/main/java/com/google/android/horologist/tiles/images/images.kt#L69

This is safe format, but requires extra work and also does not support alpha channel.

We should add a flag to support other formats including png, webp.

This probably involves removing hard coded format when flag is set

```
setFormat(ResourceBuilders.IMAGE_FORMAT_RGB_565)
```

and use, or other conversion code.

```
    val bytes = ByteArrayOutputStream().apply {
        compress(Bitmap.CompressFormat.PNG, 100, this)
    }.toByteArray()
```

We can assume Coil for our conversion code if it helps, and to start with avoiding conversions, but to be confirmed.

A later change might be also this code path https://github.com/google/horologist/blob/2f15272f92291bcdbdacd66b1042e42802c45cb7/tiles/src/main/java/com/google/android/horologist/tiles/canvas/canvas.kt

yschimke:
cc @ataulm

yschimke:
We should make this a flag, useRgb565? 

Current logic when useRgb565.

When not useRgb565, remove setFormat, and pass through the existing bytes. Tile rendering should ask the system to decode.

Samples should show a png, webp or jpg working without conversion.

ataulm:
I think for a "good first issue" this needs more context:

- what's the problem and which APIs does this affect currently?
- Are we talking only [about this function](https://github.com/google/horologist/blob/40801e6e64b55b7e0caeaba96151e8577f399afc/tiles/src/main/java/com/google/android/horologist/tiles/images/images.kt#L85)? Or will further changes need to be made elsewhere?
- is this something that we can do without exposing a flag/new API? e.g. detect the type of the image and convert it with the correct compression?
- are there similar APIs in well-established libraries like Coil, Glide and Picasso that we should follow?

yschimke:
Yep, I'll update later today.  Thanks for flagging @ataulm

Tolriq:
Sorry I missed that original issue.

The `toImageResource` function itself is agnostic of the image downloader.
Being able to keep the alpha channel is important if you apply transformations like CircleCrop.

[...]

PR #2255, opened 2024-06-06
Add support and a test for image transparency

yschimke:
#### WHAT

Support format for tile image conversions to support transparency.

#### WHY

Fixing https://github.com/google/horologist/issues/251

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -39,6 +39,13 @@
         "com.mikepenz:multiplatform-markdown-renderer"
       ],
       "allowedVersions": "/coil2/"
+    },
+    {
+      "matchPackagePatterns": [
+        "org.jetbrains.kotlinx:kotlinx-serialization-.*"
+      ],
+      "groupName": "kotlin",
+      "allowedVersions": "<1.7.0"
     }
   ]
 }


Tolriq:
After a short night of sleep, I think it's simpler and better to drop the bitmap conversion and leave it to the user with doc explaining it. Most conversion up or down should be done with dithering and specific options to ensure better result depending on the kind of image anyway.

That function should get the bitmap config and simply have a mapping for the ImageFormat from the bitmap.

yschimke:
OK, I'll land this.  Agreed, I was struggling for how the library can know what the right thing to do is.

yschimke:
@Tolriq care to give me a LGTM?

Tolriq:
@yschimke what I mean is that `toImageResource` should be parameter less and not do any conversion at all.

It should just convert the bitmap.config to an Imageformat to pass down and have proper javadoc about it.

yschimke:
It needs to fail if your Bitmap has the wrong format then.  I don't see how that helps.

I can document that you should ensure it's right before calling, but that seems like a sharp edge.

yschimke:
OK, I see what you mean.  Yep, that is a better idea.

Tolriq:
> It needs to fail if your Bitmap has the wrong format then. I don't see how that helps.

This makes things explicit, automatic bitmap modifications makes it hard for the dev to figure out what is happening when there is an issue.
I took me a lot of time to search on that function and not on the bitmap generation itself why cropping was not working.

Worse is that it could work with artifact not seen on the test images and being ugly when in prod and actual user images.

PR #2264, opened 2024-06-11
Demonstrate CoilPaintable and alpha

yschimke:
#### WHAT

Add a test showing variants on drawing with CoilPaintable with alpha, clipping.

#### WHY

Confirm the behaviour is correct

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
Test only

PR #2265, opened 2024-06-11
Don't show PositionIndicator for non scrollable dialog content

yschimke:
#### WHAT

Don't show PositionIndicator for non scrollable dialog content

#### WHY

should seem non scrollable

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/sample/src/test/kotlin/com/google/android/horologist/dialog/NonScrollableAlertContent.kt
+++ b/sample/src/test/kotlin/com/google/android/horologist/dialog/NonScrollableAlertContent.kt
@@ -33,21 +33,20 @@ public fun NonScrollableAlertContent(
     message: String? = null,
     okButtonContentDescription: String = stringResource(android.R.string.ok),
     cancelButtonContentDescription: String = stringResource(android.R.string.cancel),
-    showPositionIndicator: Boolean = true,
     content: (ScalingLazyListScope.() -> Unit)? = null,
 ) {
     val columnState = centeredDialogColumnState()
 
     AlertContent(
-        onCancel,
-        onOk,
-        icon,
-        title,
-        message,
-        okButtonContentDescription,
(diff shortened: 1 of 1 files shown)


garanj:
Hey Yuri! Curious as to how non-scrollable dialogs work: Does the developer have to declare whether it is non-scrollable or not, or does Compose calculate whether scrolling is required? (apologies, I couldn't easily see the right bit)

yschimke:
@garanj Developer has to specify.  It will look poor, but still be scrollable if they get it wrong.

yschimke:
@garanj will land based on discussion, care to LGTM?

Tolriq:
@yschimke just tried to understand why it was like that while creating an M3 version and wonder why not choose an automtic solution such as :

```
    val canScroll by remember { derivedStateOf { state.canScrollForward || state.canScrollBackward } }
    ScreenScaffold(
        modifier = modifier.fillMaxSize(),
        scrollState = if (canScroll) state else null,
        timeText = {},
    ) {
```

yschimke:
@Tolriq yeah - that's a valid question. It would work, and probably safer for potential rejections.

This component definitely isn't a clean solution, and needing a scrollbar indicates a poor layout and user experience.

In this matrix of screenshots, the grey row is why a Non Scrollable variant exists, otherwise it's too high on screen.
So the previous fix changes that to the middle row. But unfortunately introduces the issue (too low on screen) in the orange row when you have too much content.

I don't have a better fix for this with the current Wear Compose APIs.

image

Tolriq:
Hum yes, I also changed the scaling lazy to use fake item at top and bottom instead of vertical content padding to fix the indicator (including the M3 one) but also requires changes to the scroll away. 

Will do more test and play with a local copy for now.

Thanks for the details.

PR #2271, opened 2024-06-14
Disabling android test workflows on PR

yschimke:
Run manually if you want, but otherwise they continually fail

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
github workflow only - landing

PR #2282, opened 2024-06-27
Navigation parameters are read only, 

Tolriq:
Use a simple solution to only change page once per backStack.

#### WHAT

Fix sample giving invalid advice.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/media/ui/api/current.api
+++ b/media/ui/api/current.api
@@ -296,7 +296,7 @@ package com.google.android.horologist.media.ui.navigation {
 
   public static final class NavigationScreens.Player extends com.google.android.horologist.media.ui.navigation.NavigationScreens {
     method public String getPage();
-    method public Integer? getPageParam(androidx.navigation.NavBackStackEntry backStack, optional boolean remove);
+    method public Integer? getPageParam(androidx.navigation.NavBackStackEntry backStack);
     method public String libraryDestination();
     method public String playerDestination();
     property public java.util.List<androidx.navigation.NamedNavArgument> arguments;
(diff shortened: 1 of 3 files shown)


yschimke:
Thanks for the fix. 

Once we switch to 0.7.x, this will be even more important as we'll start using the typesafe nav.  https://github.com/google/horologist/compare/main...pre-0.7.x#diff-715593650c80e162494bcfc5a37a7635a80cec9c3b314f447c9fba6d58524b84R53

yschimke:
Failing on

```
/home/runner/work/horologist/horologist/media/ui/api/current.api:299: error: Removed method com.google.android.horologist.media.ui.navigation.NavigationScreens.Player.getPageParam(androidx.navigation.NavBackStackEntry,boolean) [RemovedMethod]
Aborting: Found compatibility problems checking the public API (/home/runner/work/horologist/horologist/media/ui/build/metalava/current.txt) against the API in /home/runner/work/horologist/horologist/media/ui/api/current.api
```

running

```
./gradlew metalavaGenerateSignature metalavaGenerateSignatureRelease
```

Should fix, or edit the file if brave.

Tolriq:
For the API file I wonder why that part is public API so waited for your input :) I'll update the API file.

For the type safe version from a quick look this will be a different fix actually.

You already only run the launched once per page value, so it won't do it every time, specially when returning to the page. BUT I wonder if there's case where you would actually want to change the page and it would not restart the launched effect due to same page, despite the backstack changing. Would need to be tested specially how the toRoute works.

yschimke (approved):

PR #2284, opened 2024-07-01
Fix Card Padding

yschimke:
#### WHAT

Fix Card Padding and add tests.

#### WHY

Was showing incorrectly when onLongClick was missing.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
Comparison with the bug

image

yschimke:
cc @amankhoza

amankhoza (approved):
Change looks good to me, might not be worth fixing but worth noting screenshots look like the padding across buttons with and w/o longClick is still slightly inconsistent.

yschimke:
@amankhoza I think that's the custom padding vs standard padding. If not, maybe attach a picture or video of the diff.

amankhoza:
Ah yes you're right it is just the custom padding.

PR #2291, opened 2024-07-11
Use DateTimeFormatter to produce localized AM/PM labels for TimePicker

clickxu:
#### WHAT
Use DateTimeFormatter to produce localized AM/PM labels for TimePicker instead of string resource with translation.

#### WHY
In some languages, the translated AM/PM labels are too long for a watch screen.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke (commented):
Can you add a single screenshot test that would change because of this.

clickxu:
Added a screenshot test case with localized am label in en-rGB. In this case, this change will use `am` instead of `a.m.`.

yschimke (approved):

PR #2297, opened 2024-07-17
Update cards audit

yschimke:
#### WHAT

Update cards and audit screenshots

#### WHY

Show with correct defaults for image card.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


sample/src/main/java/com/google/android/horologist/audit/CardsAudit.kt
```diff
@@ -76,7 +76,7 @@ fun BackgroundImageCard() {
                 .fillMaxSize()
                 .padding(bottom = 12.dp),
         ) {
-            Text("Text coloured to stand out on the image")
+            Text("Text coloured to stand out on the image", maxLines = 2)
```
luizgrp, line 79: nit: should it be 3? 

>  Overall the [title] and [content] text should be no more than 5 rows of text combined.

https://github.com/androidx/androidx/blob/827c53bf6fde6d21d9083a02890656e5c65a9adb/wear/compose/compose-material/src/main/java/androidx/wear/compose/material/Card.kt#L263


Tricky one if the title fits in a single line, the content could be 4 lines.

luizgrp (approved):

sample/src/main/java/com/google/android/horologist/audit/CardsAudit.kt
yschimke: There is a later spec for this.

PR #2300, opened 2024-07-17
Use ResponsiveButtons API in lists audit

yschimke:
#### WHAT

Fix bugs in the screenshots for lists, and add a section padding before buttons.

#### WHY

Match the testcases.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):

yschimke:
@luizgrp can you take a look again, seems this should be available outside dialogs.

compose-material/src/main/java/com/google/android/horologist/compose/material/ResponsiveDialog.kt
```diff
@@ -186,7 +177,24 @@ public fun ResponsiveDialogContent(
...
+): Pair<Dp, Dp> {
+    val width = LocalConfiguration.current.screenWidthDp
+    val buttonSpacedBy = 12.dp
+    // Single buttons, or buttons on smaller screens are not meant to be
+    // responsive.
+    val buttonWidth = if (width < 225 || buttonCount != 2) {
+        ButtonDefaults.DefaultButtonSize
+    } else {
+        // 14.56% margin on the sides, 12.dp between.
+        ((width * (1f - 2 * 0.1456f) - buttonSpacedBy.value) / 2).dp
```
luizgrp, line 191: Is the division by 2 here because it's assuming there will be 2 buttons (buttonCount == 2)?

Should it divide by buttonCount instead? Or throw exception if buttonCount < 1 or > 2?

yschimke: Yep - but it can only get here is buttonCount == 2. I might invert this if else.

luizgrp: ah I see 👍

luizgrp (approved):

PR #2304, opened 2024-07-19
Adds initial offset for ScalingLazyColumn rememberResponsiveColumnState

### LINKED ISSUE #2298: Add possibility to set `initialCenterItemIndex` when using `rememberResponsiveColumnState`
Is there a possibility to automatically scroll to an item when using horologist's `ScalingLazyColumn/rememberResponsiveColumnState`?

With the default `ScalingLazyColumn` I can do something like this to automatically scroll to a certain position, which is very convenient:

```
val state = rememberScalingLazyListState(
    initialCenterItemIndex = 4
)
androidx.wear.compose.foundation.lazy.ScalingLazyColumn(
     state = state,
) { //... }
```

However, it seems that `rememberResponsiveColumnState` does not currently provide an option to set the `initialCenterItemIndex`. Adding this would enhance the usability of the `ScalingLazyColumn`. 

Meanwhile, I would be happy to know if there are any alternatives to achieving the same.

kul3r4:
#### WHAT
Fixes #2298

#### WHY
So that you can set the initial index in `rememberResponsiveColumnState` similarly on what is possible on `rememberScalingLazyListState`

#### HOW
Adds a parameter for the index and use it if bigger than 0. Also sets the offsetPx to 0 when index is bigger than 0.

@yschimke Shall I include the screenshots or shall I use a different job to generate those in the CI?

#### Checklist :clipboard:
- [x] Run spotless check
- [ ] Run tests


compose-layout/src/test/snapshots/ScalingLazyColumnTest_testSetInitialRememberResponsiveColumnState[7]_large_round.png
Comment (yschimke) [inline compose-layout/src/test/snapshots/ScalingLazyColumnTest_testSetInitialRememberResponsiveColumnState[7]_large_round.png:1]: This is unfortunate, I wonder if it's some limitation of animation? Or does it happen for real in an interactive preview?

Comment (yschimke) [reply compose-layout/src/test/snapshots/ScalingLazyColumnTest_testSetInitialRememberResponsiveColumnState[7]_large_round.png:1]: I think it's breaking in Modifier.scrollAway, which reads ScalingLazyColumnState.initialScrollPosition.

We might have to fix up these APIs... let's discuss.

Comment (kul3r4) [reply compose-layout/src/test/snapshots/ScalingLazyColumnTest_testSetInitialRememberResponsiveColumnState[7]_large_round.png:1]: right, so it's seem that Time text is not fading correctly? It reproduces on preview yes, I just added one

compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
```diff
@@ -149,6 +153,7 @@ public fun rememberResponsiveColumnState(
     hapticsEnabled: Boolean = true,
     reverseLayout: Boolean = false,
     userScrollEnabled: Boolean = true,
+    index: Int = 0,
```
yschimke, line 156: What about initialItemIndex?

kul3r4: done, thanks!

yschimke:
I gave you bad advice, we need libs.wearcompose.tooling, try

```
    implementation(libs.compose.ui.tooling)
    implementation(libs.androidx.wear.tooling.preview)
    implementation(libs.wearcompose.tooling)

    debugImplementation(libs.compose.ui.toolingpreview)
    debugImplementation(projects.composeTools)
    debugImplementation(libs.androidx.activity.compose)
    debugImplementation(libs.compose.ui.test.manifest)
```

yschimke (approved):

compose-layout/src/test/snapshots/ScalingLazyColumnTest_testSetInitialRememberResponsiveColumnState[7]_large_round.png
Comment (yschimke) [inline compose-layout/src/test/snapshots/ScalingLazyColumnTest_testSetInitialRememberResponsiveColumnState[7]_large_round.png:1]: I wonder why this is wrong?

PR #2307, opened 2024-07-22
Avoid using ScalingLazyColumnState initialScrollPosition for two things.

yschimke:
#### WHAT

Avoid using ScalingLazyColumnState initialScrollPosition for two things.

#### WHY

Was both the initial position and the time text scroll away offset, so impossible to start at a different place.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
@@ -44,6 +44,7 @@ import androidx.wear.compose.foundation.rotary.RotaryScrollableDefaults.snapBeha
 import com.google.android.horologist.annotations.ExperimentalHorologistApi
 import com.google.android.horologist.compose.layout.ScalingLazyColumnDefaults.responsiveScalingParams
 import com.google.android.horologist.compose.layout.ScalingLazyColumnState.RotaryMode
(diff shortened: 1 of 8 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnState.kt
```diff
@@ -53,6 +54,7 @@ import androidx.wear.compose.foundation.lazy.ScalingLazyColumnDefaults as WearSc
 @ExperimentalHorologistApi
 public class ScalingLazyColumnState(
     public val initialScrollPosition: ScrollPosition = ScrollPosition(1, 0),
+    public val timeTextHomeOffset: ScrollPosition = initialScrollPosition,
```
kul3r4, line 57: curious why are you putting this default if always overriding?

yschimke: I'm not always overriding, there are other calls to this.  I wanted to leave those as is.

yschimke: See https://github.com/google/horologist/blob/548ad4a184bfaf0166efd172e768b9d79331c2b2/compose-layout/src/main/java/com/google/android/horologist/compose/layout/ScalingLazyColumnDefaults.kt#L54

kul3r4 (approved):

PR #2326, opened 2024-08-01
Deprecates ResponsiveTimeText

garanj:
#### WHAT

Marks `ResponsiveTimeText` as deprecated and removes usage in Horologist.

#### WHY

The recommended behavior for `ResponsiveTimeText` now aligns with the default behavior for regular `TimeText`, so it is no longer necessary to maintain this component - developers should use `TimeText` instead.

#### HOW

Mark Composable as deprecated and remove usages.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


kul3r4:
@garanj has this been backported or is that only from some Wear Compose version onwards?

yschimke (approved):

yschimke:
Hold off, discussion over 2dp/4dp

yschimke:
@kul3r4  I think we should backport the 2dp to 0.6.x, but avoid deprecation. Does that work?

PR #2330, opened 2024-08-07
Picker

ssancho14:
#### WHAT
Picker updates

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
I think that test takes multiple screenshots, so the failure gets fixed one at a time.

I should improve that.

But rerun the fixup once more

yschimke:
Will fix in a follow up

PR #2340, opened 2024-08-13
Mass cleanup before 0.7.x release

yschimke:
#### WHAT

Mass cleanup before 0.7.x release

#### WHY

Avoid carrying over deprecated APIs or extra maintenance.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


compose-material/src/main/java/com/google/android/horologist/compose/material/Chip.kt
```diff
@@ -86,7 +86,7 @@ public fun Chip(
     modifier: Modifier = Modifier,
     onLongClick: (() -> Unit)? = null,
     secondaryLabel: String? = null,
-    iconRtlMode: IconRtlMode = IconRtlMode.Default,
+    iconRtlMode: Any? = null, // kept to avoid ambiguity with other Chip
```
yschimke, line 89: @luizgrp any suggestion for this? Without it, the number of args matches the one below and it's ambiguous if no icon is used.

Maybe an extra one without icon?

luizgrp: Do you mean it matches the one in line ~177 where the icon param has type of `(@Composable BoxScope.() -> Unit)? = null`?

It should not matter if no icon is used, as both would produce the same result. 
But we could change the icon param in any of them, maybe on the one below it could be renamed to `iconContent` as it is a slot param?

compose-layout/src/main/java/com/google/android/horologist/compose/rotaryinput/Haptics.kt
```diff
@@ -251,10 +251,8 @@ private fun rememberHapticChannel() =
         )
     }
 
-@Deprecated("Replaced by wear compose")
-@ExperimentalHorologistApi
 @Composable
-public fun rememberDefaultRotaryHapticFeedback(): RotaryHapticFeedback =
+internal fun rememberDefaultRotaryHapticFeedback(): RotaryHapticFeedback =
```
luizgrp, line 255: Do we want to keep a note that this should be replaced by the one in wear compose at some point, or is not the case?

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/tiles/TilePreview.kt
```diff
@@ -59,7 +59,6 @@ import kotlin.math.roundToInt
  * Any bitmaps should be preloaded from test resources and passed in via [resourceState] as
  * Bitmap or ImageResource.
  */
-@Deprecated("Use androidx.wear.tiles.tooling.preview")
```
luizgrp, line 62: Are we republishing it as non-deprecated in a new artifact?

luizgrp (approved):
Great clean-up!

roboscreenshots/src/main/java/com/google/android/horologist/screenshots/tiles/TilePreview.kt
yschimke: Yep - It's not needed for Previews, but is needed for Screenshot testing, as the AS previews don't support that.

PR #2341, opened 2024-08-13
Include alpha in version

yschimke:
#### WHAT

Indicate the stability in the version

#### WHY

Copy the approach from https://github.com/google/accompanist/releases/tag/v0.35.1-alpha, and generally follow the Wear Compose cycle for 1.5 (not released yet)

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):
Good idea 👍

PR #2342, opened 2024-08-13
Fix publishing and deps

yschimke:
#### WHAT

Fix publishing and deps

#### WHY

JDK 17 fails in AGP javadoc generation, bump to recent AGP but still need to disable dokka from AGP.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


luizgrp (approved):

yschimke:
@luizgrp heads up

```
       1.  Dependency 'androidx.navigation:navigation-compose:2.8.0-beta07' requires libraries and applications that
           depend on it to compile against version 35 or later of the
           Android APIs.   
```

and https://androiddev.social/@ianlake/112925099458782370

PR #2363, opened 2024-08-22
Fixing issues while adding SetAudioOutput button.

hardikagr:
#### WHAT
- Adding alignment parameter for SettingsButton.
- Adding Preview for SetAudioOutputButton.
- Adding Screenshot for SetAudioOutputButton

#### WHY
While adding the SetAudioOutputButton action, the alignment parameter was deleted by mistake, also at that time the preview for SetAudioOutput button was not added.


yschimke (changes requested):
Please add screenshot tests.

At minimum for the components. But maybe one player screen example using the changes you've added previously.

yschimke (approved):

PR #2366, opened 2024-08-28
Change Card content type to match Wear Material Card

### LINKED ISSUE #2359: Change Card content type to match Wear Material's Card
Card's [content](https://github.com/google/horologist/blob/d89405564985cc1bfd2c7068017d337d7087f39b/compose-material/src/main/java/com/google/android/horologist/compose/material/Card.kt#L56) has type `@Composable () -> Unit` whereas Wear Material's Card [content](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/compose/compose-material/src/main/java/androidx/wear/compose/material/Card.kt;l=107;drc=8c1694055387362719c25d76e9ef8b969671c5e1) has type `ColumnScope.() -> Unit`.

That should be matched and the Box for the long click version should be replaced by a `Column`.

oas004:
#### WHAT
Change Card content type to match Wear Material Card

#### WHY
resolves #2359 

#### HOW
Changing the card content type to a `@Composable ColumnScope.() -> Unit` and changing `Box` to `Column` in the long click one

#### Checklist :clipboard:
- [x] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-material/api/current.api
+++ b/compose-material/api/current.api
@@ -47,7 +47,7 @@ package com.google.android.horologist.compose.material {
   }
 
   public final class CardKt {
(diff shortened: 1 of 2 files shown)


yschimke (approved):
Thanks

oas004:
@yschimke Can we merge this if it looks good? :) I don't think I have the access rights to merge it

PR #2371, opened 2024-09-01
Expose icon tint for ToggleChip composable

### LINKED ISSUE #2370: Material ToggleChip should allow overriding Icon tint
Add param

  iconTint: Color = LocalContentColor.current.copy(alpha = LocalContentAlpha.current),

oas004:
#### WHAT
Expose icon tint for ToggleChip composable

#### WHY
resolves #2370 

#### HOW
Exposing the icon tint as a pram for ToggleChip

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/compose-material/src/main/java/com/google/android/horologist/compose/material/ToggleChip.kt
+++ b/compose-material/src/main/java/com/google/android/horologist/compose/material/ToggleChip.kt
@@ -26,6 +26,8 @@ import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.takeOrElse
 import androidx.compose.ui.platform.LocalLayoutDirection
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.semantics
@@ -34,6 +36,8 @@ import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.LayoutDirection
 import androidx.wear.compose.material.ChipDefaults
+import androidx.wear.compose.material.LocalContentAlpha
+import androidx.wear.compose.material.LocalContentColor
 import androidx.wear.compose.material.MaterialTheme
 import androidx.wear.compose.material.Text
 import androidx.wear.compose.material.ToggleChip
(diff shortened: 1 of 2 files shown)


compose-material/src/main/java/com/google/android/horologist/compose/material/ToggleChip.kt
```diff
@@ -64,6 +67,7 @@ public fun ToggleChip(
     colors: ToggleChipColors = ToggleChipDefaults.toggleChipColors(),
     enabled: Boolean = true,
     interactionSource: MutableInteractionSource = remember { MutableInteractionSource() },
+    iconTint: Color = LocalContentColor.current.copy(alpha = LocalContentAlpha.current),
```
yschimke, line 70: I think this might pick up the wrong value of LocalContentColor.

Because of 

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/compose/compose-material/src/main/java/androidx/wear/compose/material/ToggleChip.kt;l=139?q=ToggleChip

We might need to push this down into the icon lambda. Maybe using Color.Unspecified as the null value.

oas004: What is the expected behaviour of this? :) I think if I change it, for the preview the icon will look like this:

image

vs 

image

yschimke: Is that with the current change, or my suggested change?

oas004: First is with the suggested `Color.Unspecified` default param (that is what you meant right? 😅). Second is the current change 😊

yschimke: There is a isUnspecified or ifUnspecified, that will use the current logic.

oas004: I couldn't quite find that `isUnspecified` or `ifUnspecified` is it part of Horologist? I'm a bit unsure what you mean with that as well, can I use `LocalContentColor.current.copy(alpha = LocalContentAlpha.current)` in the iconParam lambda if the parameter is unspecified? :)

oas004: Oh I found it, its part of `androidx.compose.ui.graphics` right?

oas004:
@yschimke Can we try to run the build pipeline for this now? :)

yschimke (approved):

yschimke:
@oas004 Thanks!

PR #2372, opened 2024-09-02
Adding watch speaker as a playable device.

hardikagr:
#### WHAT
- Adding logic in SystemAudioRepository to make watch speaker as playble.
- Adding preview and screenshot test for connected watch speaker.

#### WHY
As of now watch speaker is not ever shown as connected, with this change watch speaker will be shown as connected if the user explicitly selects it.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media/audio/src/main/java/com/google/android/horologist/audio/SystemAudioRepository.kt
```diff
@@ -37,6 +46,7 @@ public class SystemAudioRepository(
     private val _available = MutableStateFlow(mediaRouter.devices)
     private val _output = MutableStateFlow(mediaRouter.output)
     private val _volume = MutableStateFlow(mediaRouter.volume)
+    private lateinit var watchSpeakerSuitabilityChecker: WatchSpeakerSuitabilityChecker
```
wrkben, line 49: Wouldn't it be less brittle if this was a nullable `val` instead of `lateinit var`, and let it be null in the case of below V?

media/audio/src/main/java/com/google/android/horologist/audio/SystemAudioRepository.kt
```diff
@@ -83,6 +93,10 @@ public class SystemAudioRepository(
     }
 
     init {
+        if (isOsVersionAtLeastV) {
```
wrkben, line 96: I would inline this logic, it's more idiomatic

media/audio/src/main/java/com/google/android/horologist/audio/SystemAudioRepository.kt
```diff
@@ -95,14 +109,41 @@ public class SystemAudioRepository(
...
+            output = setWatchSpeakerPlayability(output)
+        }
+
+        Log.d("Hardik", "Output: " + output.isPlayable)
+
+        _available.value = devices
+        _output.value = output
+    }
+
+    @RequiresApi(VERSION_CODES.VANILLA_ICE_CREAM)
```
wrkben, line 127: If `watchSpeakerSuitabilityChecker` is nullable this function could be without RequiresAPI and always called, and instead the return below could return true if `watchSpeakerSuitabilityChecker?.isWatchSpeakerSelected() == true`, and I would believe that'd do the same.

media/audio/src/main/java/com/google/android/horologist/audio/SystemAudioRepository.kt
```diff
@@ -95,14 +109,41 @@ public class SystemAudioRepository(
...
+        } else {
+            AudioOutput.WatchSpeaker(audioOutput.id, audioOutput.name, false)
+        }
     }
 
     override fun close() {
         mediaRouter.removeCallback(callback)
         _output.value = AudioOutput.None
         _available.value = listOf()
+        if (isOsVersionAtLeastV) {
```
wrkben, line 144: Same here, if it's nullable, just call `watchSpeakerSuitabilityChecker?.unregisterControllerCallback()`

media/audio/src/main/java/com/google/android/horologist/audio/SystemAudioRepository.kt
```diff
@@ -118,6 +159,48 @@ public class SystemAudioRepository(
...
         }
+
+        private val isOsVersionAtLeastV = android.os.Build.VERSION.SDK_INT >= VERSION_CODES.VANILLA_ICE_CREAM
+    }
+
+    @RequiresApi(VERSION_CODES.VANILLA_ICE_CREAM)
+    private inner class WatchSpeakerSuitabilityChecker {
+        private val executor = Executor { command -> command.run() }
+        private val mediaRouter = MediaRouter2.getInstance(application)
+        private var wasWatchSpeakerSelectedPreviously = isWatchSpeakerSelected()
```
wrkben, line 170: I'm not seeing this `var` changed anywhere else`. Is that intentional, and if so, could it be `val`? Or is this intended to be the previously selected state for the using app?

media/audio/src/main/java/com/google/android/horologist/audio/SystemAudioRepository.kt
hardikagr: We can not remove the `@RequiresAPI` as the APIs we are using for MediaRouter2 are introduced on Android V.

media/audio/src/main/java/com/google/android/horologist/audio/SystemAudioRepository.kt
```diff
@@ -119,6 +150,47 @@ public class SystemAudioRepository(
             )
         }
     }
+
+    @RequiresApi(VERSION_CODES.VANILLA_ICE_CREAM)
+    private inner class WatchSpeakerSuitabilityChecker {
+        private val executor = Executor { command -> command.run() }
```
wrkben, line 156: One last comment: From what I remember of the MediaRouter2 implementation I believe that in practice this means that this is called on the main thread. Although I wouldn't expect that aspect of the implementation to change I still think it's better to be explicit here, and use a main executor for these calls. The reason being that the implementation requires that the callback is executed on a single thread, since it's not thread safe.

This inner class does require API 35 already, so I think it should be safe to use `context. getMainExecutor()` in place of this.

hardikagr: Done. Used `context.MainExecutor`.

kul3r4 (approved):

PR #2386, opened 2024-09-13
Add result logic to ActivityResultImplementation for VoiceInput

### LINKED ISSUE #1958: Add a ActivityResultContract for VoiceInput
suggestion: could add this to a ActivityResultContract implementation that can be reused, as it looks like this would be a common pattern

_Originally posted by @luizgrp in https://github.com/google/horologist/pull/1955#discussion_r1450063131_

oas004:
#### WHAT
Add result logic to ActivityResultImplementation for SamplePromptScreen

#### WHY
resolves #1958 

#### HOW
Adding a an implementation of `ActivityResultContract` and extracting the logic into this.

#### Checklist :clipboard:
- [x] Run spotless check
- [ ] Run tests

### DIFF
--- a/ai/sample/wear-prompt-app/src/main/java/com/google/android/horologist/ai/sample/wear/prompt/prompt/VoiceContract.kt
+++ b/ai/sample/wear-prompt-app/src/main/java/com/google/android/horologist/ai/sample/wear/prompt/prompt/VoiceContract.kt
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2024 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.android.horologist.ai.sample.wear.prompt.prompt
+
+import android.content.Context
+import android.content.Intent
+import android.speech.RecognizerIntent
+import androidx.activity.result.contract.ActivityResultContract
+
+class VoiceContract : ActivityResultContract<Intent, VoiceContract.Result>() {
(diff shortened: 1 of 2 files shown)


ai/sample/wear-prompt-app/src/main/java/com/google/android/horologist/ai/sample/wear/prompt/prompt/VoiceContract.kt
```diff
@@ -0,0 +1,41 @@
...
+        val res = intent?.getStringArrayListExtra(RecognizerIntent.EXTRA_RESULTS)
+        val enteredPrompt = res?.get(0)
+        return if (!enteredPrompt.isNullOrBlank()) {
+            Result.EnteredPrompt(enteredPrompt)
+        } else {
+            Result.Empty
+        }
+    }
+
+    sealed class Result {
```
yschimke, line 37: The built in ones seem to use a nullable result value, which avoids needing to handle the different result cases.

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:activity/activity/src/main/java/androidx/activity/result/contract/ActivityResultContracts.kt

Is that useful here?

No strong view, I'm just trying to understand what people expect? Focus on the happy path?

luizgrp (approved):
LGTM - thanks!

I've got no strong opinion on the matter that Yuri raised, nice to discuss it nonetheless.

ai/sample/wear-prompt-app/src/main/java/com/google/android/horologist/ai/sample/wear/prompt/prompt/VoiceContract.kt
yschimke: @oas004 not blocking - just for discussion, I'm happy to land if you think it's right as is.

oas004: No strong view from my side as well. I was kinda following `GoogleSignInContract` (https://github.com/google/horologist/blob/625e18a0f5dc63dd1d70f8d257a959ec8389d688/auth/ui/src/main/java/com/google/android/horologist/auth/ui/googlesignin/signin/GoogleSignInScreen.kt#L165) .  If we change change to the nullable as you say, does it make sense to change this one as well then perhaps? :) Wdyt? :)

yschimke: That's a good signal. Let's leave as is

Issue #2281, opened 2024-06-27
android_wear_capabilities and resource shrinker


Tolriq:
See: https://issuetracker.google.com/issues/348688201

But to resume `tools:keep="@array/android_wear_capabilities" ` is not a valid way to ensure that the resource is kept when present inside the xml itself.
Docs should be amended with for example the solution from https://issuetracker.google.com/issues/348688201#comment4

yschimke:
Note to update  https://developer.android.com/training/wearables/data/messages and https://developer.android.com/training/wearables/apps/standalone-apps as well

oas004:
@yschimke Updating the https://developer.android.com/training/wearables/data/messages and https://developer.android.com/training/wearables/apps/standalone-apps is that internal or is that open-source as well? :)

yschimke:
Yep, I can make those changes.

Tolriq:
@oas004 @yschimke the PR is wrong.

The array list still needs to be in value, it's only the keep that needs to be extracted and put in the raw folder.

And BTW in the array file the "tools:keep="@array/android_wear_capabilities"" can be removed as it's the non working part it avoids mistakes.

yschimke:
@Tolriq Thanks, I'll follow up with a test on the release target.

oas004:
@yschimke What is missing on this issue now? :) Is It something I should do? @Tolriq Does it look correct after the latest PR with adjustments? :)

yschimke:
I just wanted to have a test that builds the APK and confirms the resources.  I'm happy to do it, but if you felt keen, go for it.

And maybe see if we can push the raw file up into library code and have apps built correctly if they just depend on horologist datalayer.

Nek-12:
The documentation PRs were not deployed to the main docs side and I just ran into this problem when using `tools:shrinkMode="strict"` because I followed the older guide. Can the documentation be updated please?

yschimke:
Yep, we should update 

https://developer.android.com/training/wearables/data/messages
https://developer.android.com/training/wearables/apps/standalone-apps

PR #2388, opened 2024-09-13
Move wear capabilities to values folder

oas004:
#### WHAT
Follow up on https://github.com/google/horologist/issues/2281#issuecomment-2349319397 and move wear file back to values folder

#### WHY
Due to https://github.com/google/horologist/issues/2281#issuecomment-2349319397

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


oas004:
@yschimke I didn't run a test on release target, but maybe we can do that if u want? :)

docs/datalayer-helpers-guide.md
```diff
@@ -28,7 +28,7 @@ phone.
 
 2. Add the capabilities
 
-   Add a `wear.xml` file in the `res/raw` folder with the following content. It needs to be
+   Add a `wear.xml` file in the `res/values` folder with the following content. It needs to be
```
yschimke, line 31: The docs don't show the two files

res/values - the string-array
res/raw - the tools:keep

Can you split in the docs also?

yschimke:
> @yschimke I didn't run a test on release target, but maybe we can do that if u want? :)

If you want to add that go for it. It might need a new target to be added to CI workflows.

docs/datalayer-helpers-guide.md
oas004: Yeah I'll try to split them up :) Good idea :)

oas004:
> I do wonder whether the keep file if put in a library would also work for other apps using the library?

Maybe it's worth testing?

docs/datalayer-helpers-guide.md
oas004: How about this? Now I also added the NB notice to make people see that it is a different folder :)

yschimke: I think the

```
`res/raw` folder due to https://issuetracker.google.com/issues/348688201
```

Needs to move down to the next paragraph.

oas004: Oh yes you are right! 🤦 😅  Thanks! Fixing

PR #2394, opened 2024-09-18
Cargo Cult JVM Settings

yschimke:
#### WHAT

Copy https://github.com/android/nowinandroid/pull/1492 and https://www.jasonpearson.dev/metaspace-in-jvm-builds/

#### WHY

Improve CI and local performance?

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -22,8 +22,8 @@ org.gradle.parallel=true
 # Declare we support AndroidX
 android.useAndroidX=true
 
-# Increase memory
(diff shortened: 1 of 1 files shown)


kaeawc:
Hey - so while these settings are better, I wanted to let you know I [found an issue with how Kotlin Daemon picks up and doesn't pick up JVM args](https://youtrack.jetbrains.com/issue/KT-71564/Kotlin-Daemon-JVM-args-behavior-does-not-match-documentation). I'm digging into it and will ping y'all when I have an update.

kaeawc:
@yschimke FYI I updated my recommendations. Let me know if you have any questions or further issues with metaspace.

Issue #1563, opened 2023-08-08
[Feature Suggestion] TextField Composable for Wear OS


uripre:
In my opinion, it's crazy that we have to implement text input with startActivityForResult/RemoteInputIntentHelper etc. I think there should be a TextField composable to offer this functionality out of the box.

Would love to hear thoughts.

yschimke:
Discussed here 

https://github.com/google/horologist/issues/1355

We could implement something, but basically combining a text input with the launch for result.

uripre:
Thanks for the reference, Yuri. I think we should. Even if the behavior of TextInputs on Wear OS is to launch a popup over the entire screen and then return to the app, I think this behavior should be seamlessly encapsulated in a TextInput component.

Fra078:
My suggestion is to implement a composable which use KeyboardInputController classes because RemoteInput is not always supported on non-wearOS wearables like some Huawei watches and Oppo. Now is difficult to publish apps stores different from Google play because of this problem

oas004:
Bringing the discussion from #2395 into here to discuss requirements for this component. As @luizgrp mentioned here https://github.com/google/horologist/pull/2395#discussion_r1767107055. "This intent is for gboard, which doesn't come pre-installed in every watch ", how would be the best way to handle this from a UX perspective? If we where to use the `KeyboardInputController` is that already handling that perhaps?

yschimke:
closing as stale, we haven't done this already, we probably won't.

PR #2395, opened 2024-09-18
Add API and launcher for NonEditableBasicTextField

oas004:
#### WHAT
Adding NonEditableBasicTextField composable to open the keyboard on click.

Currently a draft as I need to get it to work. Opening a draft PR here so I can get some feedback on the solution :) 

#### WHY
Ref. #1563 

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


composables/src/main/java/com/google/android/horologist/composables/NonEditableBasicTextField.kt
```diff
@@ -0,0 +1,115 @@
...
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.SolidColor
+import androidx.compose.ui.text.TextLayoutResult
+import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.input.TextFieldValue
+import androidx.compose.ui.text.input.VisualTransformation
+
+/**
+ * This is a wrapper for [BasicTextField] that when clicked launches the keyboard. It cannot be
```
oas004, line 40: Need to improve these, but maybe wait until API is settled

composables/src/main/java/com/google/android/horologist/composables/NonEditableBasicTextField.kt
```diff
@@ -0,0 +1,114 @@
...
+        decorationBox = decorationBox,
+    )
+}
+
+private class KeyBoardContract : ActivityResultContract<Intent, KeyBoardContract.Result>() {
+    override fun createIntent(context: Context, input: Intent): Intent = input
+
+    override fun parseResult(resultCode: Int, intent: Intent?): Result {
+        return if (resultCode == RESULT_OK) {
+            val resultText = intent?.extras?.getString("result_text")
```
oas004, line 99: This I need to fix

oas004: I think it works, it's just that it's overridden by the `BasicTextField` implementation. Maybe using that wasn't such a good idea 🤔

composables/src/main/java/com/google/android/horologist/composables/NonEditableBasicTextField.kt
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+    val keyboardIntent = Intent("com.google.android.wearable.action.LAUNCH_KEYBOARD")
+    BasicTextField(
+        modifier = modifier.clickable {
+            keyboardLauncher.launch(keyboardIntent)
+        },
+        value = value,
+        onValueChange = {},
+        enabled = false,
```
oas004, line 76: This made it work since its suppressing the `BasicTextField`, but it feels quite hacky, maybe not the solution we want to go for I guess 😅

composables/src/main/java/com/google/android/horologist/composables/NonEditableBasicTextField.kt
```diff
@@ -0,0 +1,112 @@
...
+        @Composable { innerTextField -> innerTextField() },
+) {
+    val keyboardLauncher = rememberLauncherForActivityResult(
+        KeyBoardContract(),
+    ) {
+        if (it is KeyBoardContract.Result.TextChanged) {
+            onTextChanged(it.text)
+        }
+    }
+    val keyboardIntent = Intent("com.google.android.wearable.action.LAUNCH_KEYBOARD")
```
luizgrp, line 69: This intent is for gboard, which doesn't come pre-installed in every watch (e.g. samsung watches). 
This component would need to check if the intent exists, if it doesn't, we would need to think on a good behavior, e.g. call RemoteInputIntentHelper.startActivityForResult or let the user of the component provide the fallback behavior?

composables/src/main/java/com/google/android/horologist/composables/NonEditableBasicTextField.kt
luizgrp: What was not working before disabling the component?

luizgrp (commented):
Thanks for working on this!

I will do another round of review tomorrow, just leaving these comments for now.

Could you please add a video showing the usage of the component to the PR description? A sequence of screenshots is also fine.

composables/src/main/java/com/google/android/horologist/composables/NonEditableBasicTextField.kt
oas004: Before doing this, I think the BasicTextField was interfering (just my assumption). The text was not able to update correctly

composables/src/main/java/com/google/android/horologist/composables/NonEditableBasicTextField.kt
oas004: Ah I didn't know, maybe providing some callback would be good? No strong opinions on how best to handle this from my side. Is this maybe a UX question perhaps? :)

oas004:
> Thanks for working on this!
> 
> I will do another round of review tomorrow, just leaving these comments for now.
> 
> Could you please add a video showing the usage of the component to the PR description? A sequence of screenshots is also fine.

Thank you for checking this out and giving valuable feedback 🙏  Currently very in draft stage. I can provide a video on how it's working now. From what I understood in the issue we want to have a wrapper on `BasicTextField` exposing mostly the same API, that when clicking it it launches this intent instead to go to another screen to do the input. Is that the correct assumption? :)

oas004:
I think maybe I'll close this for now until we get a bit more input from the UX. Not quite sure how this component should work to provide most value to the users

luizgrp:
Thanks for the patience, Odin. I will ping you here once I get more information.

PR #2402, opened 2024-09-20
Fix double time text

yschimke:
#### WHAT

Fix double time text, TimeText on ScreenScaffold should only be drawn by the AppScaffold.

#### WHY

Before 
image

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


garanj (approved):

kul3r4 (approved):
thanks for fixing! tried on the Compose starter sample and works well.

PR #2407, opened 2024-09-23
Update Uamp Player Screen bottom buttons layout

rajat4064g:
#### WHAT
Update Uamp Player Screen for two buttons and app icon layout at the bottom

PR #2409, opened 2024-09-25
Remove remnants of splash screen

yschimke:
#### WHAT

Remove remnants of splash screen

#### WHY

Causing robolectric failures

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
Landing to unblock https://github.com/google/horologist/pull/2407

PR #2412, opened 2024-09-28
Use metalava plugin from version catalog

oas004:
#### WHAT
Use metalava plugin from version catalog instead of hardcoding the string with the id.

#### HOW
Replace all usages with alias

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/ai/ui/build.gradle.kts
+++ b/ai/ui/build.gradle.kts
@@ -17,7 +17,7 @@
 plugins {
     id("com.android.library")
     id("org.jetbrains.dokka")
-    id("me.tylerbwong.gradle.metalava")
+    alias(libs.plugins.metalavaGradle)
     kotlin("android")
     alias(libs.plugins.roborazzi)
     alias(libs.plugins.compose.compiler)
--- a/annotations/build.gradle.kts
+++ b/annotations/build.gradle.kts
@@ -18,7 +18,7 @@ plugins {
     id("java-library")
     id("org.jetbrains.kotlin.jvm")
     id("org.jetbrains.dokka")
-    id("me.tylerbwong.gradle.metalava")
+    alias(libs.plugins.metalavaGradle)
 }
 
 java {
--- a/auth/composables/build.gradle.kts
+++ b/auth/composables/build.gradle.kts
@@ -17,7 +17,7 @@
 plugins {
     id("com.android.library")
     id("org.jetbrains.dokka")
-    id("me.tylerbwong.gradle.metalava")
+    alias(libs.plugins.metalavaGradle)
     alias(libs.plugins.dependencyAnalysis)
     kotlin("android")
     alias(libs.plugins.roborazzi)
--- a/auth/data-phone/build.gradle.kts
+++ b/auth/data-phone/build.gradle.kts
@@ -18,7 +18,7 @@ plugins {
     id("com.android.library")
     id("kotlin-android")
     id("org.jetbrains.dokka")
-    id("me.tylerbwong.gradle.metalava")
+    alias(libs.plugins.metalavaGradle)
     alias(libs.plugins.dependencyAnalysis)
 }
 
--- a/auth/data/build.gradle.kts
+++ b/auth/data/build.gradle.kts
@@ -18,7 +18,7 @@ plugins {
     id("com.android.library")
     id("org.jetbrains.dokka")
     id("com.google.devtools.ksp")
-    id("me.tylerbwong.gradle.metalava")
+    alias(libs.plugins.metalavaGradle)
     alias(libs.plugins.dependencyAnalysis)
     kotlin("android")
 }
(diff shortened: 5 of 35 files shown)


ai/ui/build.gradle.kts
```diff
@@ -16,8 +16,8 @@
 
 plugins {
     id("com.android.library")
-    id("org.jetbrains.dokka")
-    id("me.tylerbwong.gradle.metalava")
+    id(libs.plugins.dokka.get().pluginId)
```
yschimke, line 19: I wonder if this is an improvement. I think yes for consistency.

But these strings are fundamental. Maybe we should change all of these to simple string ids, and save version catalogs for the places we need a version in the top level build file?

But not blocking.

yschimke: Basically since there is no version, do we prefer

```
id("org.jetbrains.dokka")
```

to 

```    
id(libs.plugins.dokka.get().pluginId)
```

oas004: Ah yeah okay I think that makes sense :) perhaps I can revert that commit?

gradle/libs.versions.toml
```diff
@@ -221,6 +221,7 @@ wearcompose-tooling = { module = "androidx.wear.compose:compose-ui-tooling", ver
 [plugins]
 compose-compiler = { id = "org.jetbrains.kotlin.plugin.compose", version.ref = "kotlin" }
 dependencyAnalysis = { id = "com.autonomousapps.dependency-analysis", version.ref = "dependencyAnalysis" }
+dokka = { id = "org.jetbrains.dokka", version.ref = "dokka" }
```
yschimke, line 224: +1 

Should `build.gradle.kts` use this via plugins instead of classpath?

https://github.com/google/horologist/blob/main/build.gradle.kts#L38

oas004: I think that could be a good idea!

ai/ui/build.gradle.kts
yschimke: Also the metalava id is safe because we have this here https://github.com/google/horologist/blob/main/build.gradle.kts#L50

```
plugins {
  alias(libs.plugins.metalavaGradle) apply false
}
```

yschimke:
With the grpc changes, it was a lot more obvious to avoid updating versions in multiple places.  With this change, I think removing all the classpath entries in top level build.gradle.kts, and using plugins cleanly might be an improvement?

ai/ui/build.gradle.kts
oas004: So in that case it is also better to use string Ids for metalava? 🤔

yschimke: This is pure opinion, no hard and fast rule. But definitely easier to understand for someone new to the project

oas004:
> With the grpc changes, it was a lot more obvious to avoid updating versions in multiple places. With this change, I think removing all the classpath entries in top level build.gradle.kts, and using plugins cleanly might be an improvement?

I think that would be quite nice! :) With "using plugins cleanly" are you then referring to the string representation in each module? Or can that come from version catalog using `alias(libs.plugin.x)`? Wdyt? :)

ai/ui/build.gradle.kts
oas004: I kinda agree with the `id(libs.x.get.pluginId)` workaround from https://github.com/gradle/gradle/issues/17968#issuecomment-894742093 is not quite nice 😅 . However if I move them out from class path, IMO it would be nicer to refer to them from version catalog with the `alias` instead of the string representation, but it's not a strong opinion. My only concern with the string rep is that it feels like it is a bit more prone to errors since the IDE is not giving like an error message right away if you misstype something. It will just fail sync. I guess that is fine as well, I can revert that to use the string rep if you think that is cleaner :)

build.gradle.kts
```diff
@@ -35,23 +35,22 @@ buildscript {
...
     alias(libs.plugins.ksp) apply false
-    alias(libs.plugins.protobuf) apply false
-    alias(libs.plugins.gradleMavenPublishPlugin)
     alias(libs.plugins.metalavaGradle) apply false
-    alias(libs.plugins.dependencyAnalysis)
+    alias(libs.plugins.protobuf) apply false
     alias(libs.plugins.roborazzi) apply false
-    alias(libs.plugins.kotlinx.serialization) apply false
-    alias(libs.plugins.compose.compiler) apply false
+    alias(libs.plugins.spotless)
```
oas004, line 53: Should probably have split this up into one more commit to make it easy to see. It's just add dokka and re-arrange them in alphabetically order

yschimke (approved):

SongWoody:
"Updating the distributionUrl in root/gradle/wrapper/gradle-wrapper.properties to use gradle-8.8-bin.zip or a later version will resolve this issue."

oas004:
> "Updating the distributionUrl in root/gradle/wrapper/gradle-wrapper.properties to use gradle-8.8-bin.zip or a later version will resolve this issue."

What will it resolve? 🤔😅

PR #2114, opened 2024-03-13
Adds Tile prompt on phone side

kul3r4:
#### WHAT
Add tile prompt on the phone side, needs a second PR for the watch side

#### WHY
Allow prompting Tiles install for watches that support this functionality

#### HOW
Adds a prompt to deep link into Tiles editor

#### Checklist :clipboard:
- [ ] Add explicit visibility modifier and explicit return types for public declarations
- [X] Run spotless check
- [ ] Run tests
- [ ] Update metalava's signature text files

PR #2414, opened 2024-10-02
Improve Tracking tile installation

### LINKED ISSUE #2066: Improve tracking Tile installation
Since [getActiveTilesAsync](https://developer.android.com/reference/androidx/wear/tiles/TileService#getActiveTilesAsync(android.content.Context,java.util.concurrent.Executor)) became available, our guidance should be updated to use it, instead of using the information provided by [onTileAddEvent](https://developer.android.com/reference/androidx/wear/tiles/TileService#onTileAddEvent(androidx.wear.tiles.EventBuilders.TileAddEvent)) and [onTileRemoveEvent](https://developer.android.com/reference/androidx/wear/tiles/TileService#onTileRemoveEvent(androidx.wear.tiles.EventBuilders.TileRemoveEvent)).

---

Suggestion:

1) Remove `markTileAsInstalled` and `markTileAsRemoved` functions;
2) Add function to update the tracking of installed tiles:

```kotlin
    public suspend fun updateInstalledTiles() {
        val activeTiles = TileService.getActiveTilesAsync(context, executor).await()

        surfacesInfoDataStore.updateData { info ->
            info.copy {
                tiles.clear()
                for (activeTileIdentifier in activeTiles) {
                    tiles.add(tileInfo {
                        timestamp = System.currentTimeMillis().toProtoTimestamp()
                        name = activeTileIdentifier.componentName.className
                    })
                }
            }
        }
    }
```
3) Update guidance in https://google.github.io/horologist/datalayer-helpers-guide/ to call `updateInstalledTiles` in `onTileAddEvent` and `onTileRemoveEvent`.

oas004:
#### WHAT
Add updateInstalledTiles and deprecate markTileAsRemoved and markTileAsInstalled.

#### WHY
resolves #2066 

#### HOW
Add updateInstalledTiles and deprecate markTileAsRemoved and markTileAsInstalled.

Making this as a draft, as I have to update the documentation and test that it works correctly

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


datalayer/watch/api/current.api
```diff
@@ -13,7 +13,8 @@ package com.google.android.horologist.datalayer.watch {
     method public suspend Object? markSetupNoLongerComplete(kotlin.coroutines.Continuation<? super kotlin.Unit>);
     method public suspend Object? markTileAsInstalled(String tileName, kotlin.coroutines.Continuation<? super kotlin.Unit>);
     method public suspend Object? markTileAsRemoved(String tileName, kotlin.coroutines.Continuation<? super kotlin.Unit>);
-    method @CheckResult public suspend Object? startCompanion(String nodeId, kotlin.coroutines.Continuation<? super com.google.android.horologist.data.AppHelperResultCode>);
+    method @CheckResult public suspend Object? startCompanion(String nodeId, kotlin.coroutines.Continuation<? super error.NonExistentClass>);
```
oas004, line 16: I think this I have to revert

datalayer/watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
```diff
@@ -168,6 +171,36 @@ public class WearDataLayerAppHelper internal constructor(
...
+    /**
+     * Updates the installed tiles
+     *
+     * Gets all the installed tiles async and updates them.
+     *
+     * @param executor The executor on which methods should be invoked.
+     * To dispatch events through the main thread of your application, you can
+     * use Context. getMainExecutor().
+     */
+    public suspend fun updateInstalledTiles(executor: Executor) {
```
luizgrp, line 183: I'd we should use an executor from a Dispatcher injected to WearDataLayerAppHelper instead of passing the responsibility to the caller, given this is already a suspended function and will not block.

@yschimke I see [WearDataLayerRegistry](https://github.com/google/horologist/blob/2ec7825ff650d0db3f3fc1be3ed66e97cc8bfcc6/datalayer/core/src/main/java/com/google/android/horologist/data/WearDataLayerRegistry.kt#L56) has a CoroutineScope not in use and I remember you had plans for that. I'd suggest to change the constructor of this class (WearDataLayerAppHelper) and of WearDataLayerRegistry to accept a Dispatcher instead of CoroutineScope. Do you agree?

datalayer/watch/src/test/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelperTest.kt
```diff
@@ -99,13 +99,13 @@ class WearDataLayerAppHelperTest {
...
 
-        helper.markTileAsInstalled("my.SampleTileService")
+        helper.updateInstalledTiles(context.mainExecutor)
 
         val infoUpdated = testDataStore.data.first()
         assertThat(infoUpdated.tilesList).hasSize(1)
         assertThat(infoUpdated.tilesList.first().name).isEqualTo("my.SampleTileService")
 
-        helper.markTileAsRemoved("my.SampleTileService")
+        helper.updateInstalledTiles(context.mainExecutor)
```
luizgrp, line 108: I'd say we should duplicate this test and keep the unit test for the deprecated functions until they are removed - I know we are probably not changing them anymore but good to keep them being tested during library/API updates.

luizgrp (commented):
Thanks!!

Could you please add evidence that the sample apps are still working with the new solution? Either via screenshots or a video.

The sample apps are datalayer.sample.wear and datalayer.sample.phone. 
We should check: 
- The [tracking](https://github.com/google/horologist/pull/1883) of tiles ("Tracking" button on the wear app, "Nodes" button on the phone app).
- The [Install Tile prompt](https://github.com/google/horologist/pull/2114) ("Install Tile" button on the phone app). cc: @kul3r4.

Please let me know if you need more information on how to test them!

---

Failing in CI:
```
WearDataLayerAppHelperTest > testTiles FAILED
    java.lang.NoSuchMethodError at WearDataLayerAppHelperTest.kt:81
```

datalayer/watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
```diff
@@ -168,6 +171,36 @@ public class WearDataLayerAppHelper internal constructor(
         }
     }
 
+    /**
+     * Updates the installed tiles
```
luizgrp, line 175: We should probably add this disclaimer to this kdoc:

```
This function has some limitations on older SDK versions, please see the docs for [TileService#getActiveTilesAsync](https://developer.android.com/reference/androidx/wear/tiles/TileService#getActiveTilesAsync(android.content.Context,java.util.concurrent.Executor))
```

oas004: Aha! Good point! I'll add that!

datalayer/watch/src/test/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelperTest.kt
oas004: Good point! I'll keep the test as it was for the deprecated functions until they are gone :)

oas004:
> Thanks!!
> 
> Could you please add evidence that the sample apps are still working with the new solution? Either via screenshots or a video.
> 
> The sample apps are datalayer.sample.wear and datalayer.sample.phone. We should check:
> 
> * The [tracking](https://github.com/google/horologist/pull/1883) of tiles ("Tracking" button on the wear app, "Nodes" button on the phone app).
> * The [Install Tile prompt](https://github.com/google/horologist/pull/2114) ("Install Tile" button on the phone app). cc: @kul3r4.
> 
> Please let me know if you need more information on how to test them!
> 
> Failing in CI:
> 
> ```
> WearDataLayerAppHelperTest > testTiles FAILED
>     java.lang.NoSuchMethodError at WearDataLayerAppHelperTest.kt:81
> ```

Thank you for the help! 🙏 To setup the communication between the phone and wear  device, is that possible to do with the emulator? Or do you need a physical devices to do that? :)

luizgrp:
> To setup the communication between the phone and wear device, is that possible to do with the emulator?

Yes, it is! This [page](https://developer.android.com/training/wearables/get-started/connect-phone) should guide you, let me know if you have any trouble!

oas004:
I managed to setup the testing of this, but I see that I can't install tiles from the tracking part of the wear app. I can only install it from the phone. I will try to do some debugging of it today. Sorry about that @luizgrp

luizgrp:
No worries, take your time! I don't get what you mean that you can only install the "tiles" from the phone?

To clarify, this is what I'm expecting to see, when you flip these toggles for the tiles in the tracking screen of the watch, the information should get updated on the phone, on the nodes screen:

| Screenshot_20241004_090746 |  | 
|--------|--------|

datalayer/watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
yschimke: Sure.  I think you convinced me that WearDataLayerRegistry was not worth the bother. I had considered it a useful app singleton, but clients are cheap.

The only other purpose was registering Serializers, but there are other ways to achieve that.

oas004:
> No worries, take your time! I don't get what you mean that you can only install the "tiles" from the phone?
> 
> To clarify, this is what I'm expecting to see, when you flip these toggles for the tiles in the tracking screen of the watch, the information should get updated on the phone, on the nodes screen:
> 
> Screenshot_20241004_090746	

Yeah I should have been more clear. The case you posted, I'm not able to do with this change, so I have to figure out why. What I was able to do was to run the demo for install tiles, but I'm realising that might be a bit unrelated 😅

oas004:
I tired to debug this a little, and it seems like it is updating from the surfacesInfo. It's triggering the state to update, but when I debug it, there is some internal error showing in the debugger. Have you seen that before @luizgrp ? Maybe I'm doing something wrong when I'm adding the tile perhaps 🤔 
image

luizgrp:
>  when I debug it, there is some internal error showing in the debugger.

Just tried here with the latest stable Android Studio, it seems to be working when I debug that same line. I do see similar issue (not internal error, but ClassNotFoundException) when debugging inside WearDataLayerAppHelper.updateInstalledTiles when I check `activeTileIdentifier` with the mouse, but the debug panel inspects `activeTileIdentifier` no problem.

---

For that `tilesMap`, before the availability of this getActiveTilesAsync API, we had to control the existence of the tiles within the library, so it was possible to use "fake" tiles. Now that the library uses getActiveTilesAsync API, getActiveTilesAsync API won't return any "fake" tiles, only real ones, so the logic in TrackingScreenViewModel won't match any of the real tiles with the fake list. 

We would have to update the list to have the real tile (com.google.android.horologist.datalayer.sample.SampleTileService), and update the logic to only send "SampleTileService" to the UI model, so it can be seen on the screen. Also, the toggle should not trigger the removal/addition of the tile anymore, the onTileCheckedChanged callback can be removed.

See in the video how the screen gets updated automatically as it observes the changes in wearDataLayerAppHelper.surfacesInfo:

[Screen_recording_20241007_091101.webm](https://github.com/user-attachments/assets/9da5ad26-614f-4682-967a-5ae98e4d649a)

oas004:
Thanks for testing! I didn't quite get why I was getting that, but might be the AS version I'm on.

Aha! Thanks for the thorough explanation! That makes more sense :) Sorry for not checking this. 

I updated the PR with removing that callback and adding the real tiles in the list instead. Then the tracking screen shows like this now

image

datalayer/watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
oas004: So if I understand this correctly, it is better if I provide the executor as a constructor injection here? Are we providing the executor somewhere? :)

datalayer/watch/src/test/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelperTest.kt
```diff
@@ -113,6 +113,42 @@ class WearDataLayerAppHelperTest {
         coroutineContext.cancelChildren()
     }
 
+    @Test
+    fun testTilesWithUpdate() = runTest {
```
oas004, line 117: Can I still do this test with a unit test @luizgrp ? Maybe also the logic of this test is a bit strange now with the `updateInstalledTiles` 🤔 Trying to debug this was a bit hard. Locally I'm getting a different error than on CI. Locally I'm getting this for all the tests in this class

`java.lang.NoClassDefFoundError: org/conscrypt/OpenSSLProvider`

Maybe there is something I need to configure to run these tests?

luizgrp: You would probably need to explore the [androidx.wear.tiles:tiles-testing](https://developer.android.com/jetpack/androidx/releases/wear-tiles#declaring_dependencies) library to see if it helps you to test this. You could potentially use [sendOnTileAddedEvent](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/tiles/tiles-testing/src/main/java/androidx/wear/tiles/testing/TestTileClient.kt;l=131;drc=c93d3c05451c91bd2ea3e7371fc2e4da9de4e5f6) and sendOnTileRemovedEvent to emulate what we do in the test above with markTileAsInstalled/markTileAsRemoved.

The [test](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/tiles/tiles-testing/src/test/java/androidx/wear/tiles/testing/TestTileClientTest.kt) of that class could be used for inspiration on how to create a [fake TileService](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/tiles/tiles-testing/src/test/java/androidx/wear/tiles/testing/TestTileClientTest.kt;l=134;drc=c93d3c05451c91bd2ea3e7371fc2e4da9de4e5f6) and set up [RoboletricTestRunnner](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/tiles/tiles-testing/src/test/java/androidx/wear/tiles/testing/TestTileClientTest.kt;l=36;drc=c93d3c05451c91bd2ea3e7371fc2e4da9de4e5f6).

luizgrp:
> Then the tracking screen shows like this now

Somehow the image is not loading for me!

oas004:
> > Then the tracking screen shows like this now
> 
> Somehow the image is not loading for me!

hmm strange,  maybe its since I copied it in. Does it work now when I upload it?
Screenshot 2024-10-11 at 17 51 23

datalayer/watch/src/test/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelperTest.kt
oas004: Ooo thanks! I'll give it a try! 🙏

datalayer/watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
oas004: I can do that as an intermediate step, that makes sense :)

oas004: I'll require that to be non-null though, is that okay?

luizgrp: Sounds good!

datalayer/watch/src/test/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelperRoboTest.kt
```diff
@@ -0,0 +1,156 @@
...
+import org.junit.runner.RunWith
+import org.junit.runners.model.FrameworkMethod
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.android.util.concurrent.InlineExecutorService
+import org.robolectric.annotation.internal.DoNotInstrument
+import org.robolectric.internal.bytecode.InstrumentationConfiguration
+
+@RunWith(TilesTestingTestRunner::class)
+@DoNotInstrument
+class WearDataLayerAppHelperRoboTest {
```
oas004, line 50: @luizgrp Shouldn't this work? It's almost the same as the example you provided 🤔 I'm getting an initialisation error from the FakeTileService I think. Is there something I'm missing? 😅 

```
java.lang.NoClassDefFoundError: com/google/wear/services/tiles/TileInstance
	at java.base/java.lang.Class.getDeclaredMethods0(Native Method)
	at java.base/java.lang.Class.privateGetDeclaredMethods(Class.java:3402)
	at java.base/java.lang.Class.getDeclaredMethod(Class.java:2673)
	at org.robolectric.util.ReflectionHelpers.lambda$callInstanceMethod$6(ReflectionHelpers.java:339)
	at org.robolectric.util.ReflectionHelpers.traverseClassHierarchy(ReflectionHelpers.java:543)
	at org.robolectric.util.ReflectionHelpers.callInstanceMethod(ReflectionHelpers.java:335)
	at org.robolectric.android.controller.ComponentController.lambda$invokeWhilePaused$0(ComponentController.java:57)
	at org.robolectric.shadows.ShadowPausedLooper.runPaused(ShadowPausedLooper.java:243)
	at org.robolectric.android.controller.ComponentController.invokeWhilePaused(ComponentController.java:64)
	at org.robolectric.android.controller.ComponentController.invokeWhilePaused(ComponentController.java:56)
	at org.robolectric.android.controller.ServiceController.create(ServiceController.java:64)
	at androidx.wear.tiles.testing.TestTileClient.maybeBind(TestTileClient.kt:153)
	at androidx.wear.tiles.testing.TestTileClient.requestTile(TestTileClient.kt:108)
```

luizgrp: Raised: https://issuetracker.google.com/issues/374901735

oas004: Added this to the ignore for this test now :)

datalayer/watch/src/main/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelper.kt
```diff
@@ -168,6 +173,43 @@ public class WearDataLayerAppHelper internal constructor(
...
+     */
+    @OptIn(ExperimentalStdlibApi::class)
+    public suspend fun updateInstalledTiles() {
+        val executor = scope.coroutineContext[CoroutineDispatcher]?.asExecutor()
+
+        require(executor != null) {
+            "Executor is null, something is wrong during initalization of WearDataLayerAppHelper"
+        }
+
+
```
oas004, line 192: Remove this line perhaps

docs/datalayer-helpers-guide.md
```diff
@@ -201,13 +201,13 @@ To determine whether your Tile(s) are installed, add the following to your `Tile
 In `onTileAddEvent`:
 
 ```kotlin
-wearAppHelper.markTileAsInstalled("SummaryTile")
+wearAppHelper.updateInstalledTiles()
```
oas004, line 204: Is this okay on the docs @luizgrp ? Or should I maybe put them together as it's the same function? Wdyt?

luizgrp: Sounds good!

luizgrp (approved):
Thanks! Apologies for taking too long to get it merged.

datalayer/watch/src/test/java/com/google/android/horologist/datalayer/watch/WearDataLayerAppHelperRoboTest.kt
luizgrp: This seems to have been [fixed](https://android-review.git.corp.google.com/c/platform/frameworks/support/+/3341018) in [alpha-05](https://developer.android.com/jetpack/androidx/releases/wear-tiles#1.5.0-alpha05).

Raised https://github.com/google/horologist/issues/2487

PR #2420, opened 2024-10-05
Update CombinedInferenceServiceRegistry.kt

GeorgCantor:
#### WHAT
**Use flatMap for Combining Flows**: Instead of using combine and then flattening the result, we can use flatMapConcat, which is more efficient for this use case.

#### WHY
**Performance Improvement**: Using flatMapConcat is generally more efficient than combining and flattening lists afterward.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/ai/sample/core/src/main/java/com/google/android/horologist/ai/core/registry/CombinedInferenceServiceRegistry.kt
+++ b/ai/sample/core/src/main/java/com/google/android/horologist/ai/core/registry/CombinedInferenceServiceRegistry.kt
@@ -24,8 +24,9 @@ class CombinedInferenceServiceRegistry(
     val registries: List<InferenceServiceRegistry>,
 ) : InferenceServiceRegistry {
     override fun models(): Flow<List<InferenceServiceGrpcKt.InferenceServiceCoroutineImplBase>> {
-        return combine(registries.map { registry -> registry.models() }) {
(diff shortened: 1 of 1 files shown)


yschimke:
It's not compiling

```

> Task :ai:sample:core:compileDebugKotlin FAILED
e: file:///home/runner/work/horologist/horologist/ai/sample/core/src/main/java/com/google/android/horologist/ai/core/registry/CombinedInferenceServiceRegistry.kt:27:27 Unresolved reference 'asFlow'.
e: file:///home/runner/work/horologist/horologist/ai/sample/core/src/main/java/com/google/android/horologist/ai/core/registry/CombinedInferenceServiceRegistry.kt:28:30 Cannot infer type for this parameter. Please specify it explicitly.
e: file:///home/runner/work/horologist/horologist/ai/sample/core/src/main/java/com/google/android/horologist/ai/core/registry/CombinedInferenceServiceRegistry.kt:28:51 Unresolved reference 'models'.
e: file:///home/runner/work/horologist/horologist/ai/sample/core/src/main/java/com/google/android/horologist/ai/core/registry/CombinedInferenceServiceRegistry.kt:30:14 Unresolved reference 'asFlow'.
```

yschimke:
But generally the new logic seems different anyway.

PR #2422, opened 2024-10-07
Bug Fix :  Observed Talkback Doesn't Announces Increase/Decrease Volume as "Buttons"

2-coder:
#### WHAT

Added Semantic type to icon in VolumeScreen

#### WHY
Accessibility fix 

#### Checklist :clipboard:
- [x] Run spotless check
- [ ] Run tests


media/audio-ui/src/main/java/com/google/android/horologist/audio/ui/VolumeScreen.kt
```diff
@@ -231,7 +233,7 @@ public object VolumeScreenDefaults {
     @Composable
     public fun IncreaseIcon() {
         Icon(
-            modifier = Modifier.size(26.dp),
+            modifier = Modifier.size(26.dp).semantics { role = Role.Button },
```
yschimke, line 236: Icon itself isn't a button. Shouldn't Stepper (Wear Compose) handle this?

We wrap it slightly, but called here.

https://github.com/google/horologist/blob/0a7a0d4e0e546d12b48495d14d3b25b02f5e8e84/compose-material/src/main/java/com/google/android/horologist/compose/material/Stepper.kt#L39

2-coder: Stepper is not handing this.  Default implementation of Increase Icon and decrease Icon description doesn't fulfill our requirement.

yschimke: No objection to landing this, but we should probably track a bug in Wear Compose to fix this, and revert once it happens.

yschimke (approved):

Issue #2417, opened 2024-10-04
Exception android.net.ConnectivityManager$TooManyRequestsException


Tolriq:
I'm seeing some crash in prod with the following stacktrace

```
Exception android.net.ConnectivityManager$TooManyRequestsException:
  at android.net.ConnectivityManager.convertServiceException (ConnectivityManager.java:4236)
  at android.net.ConnectivityManager.sendRequestForNetwork (ConnectivityManager.java:4437)
  at android.net.ConnectivityManager.sendRequestForNetwork (ConnectivityManager.java:4444)
  at android.net.ConnectivityManager.requestNetwork (ConnectivityManager.java:4582)
  at android.net.ConnectivityManager.requestNetwork (ConnectivityManager.java:4560)
  at com.google.android.horologist.networks.highbandwidth.StandardHighBandwidthNetworkMediator$processRequest$1.invoke (StandardHighBandwidthNetworkMediator.java)
  at com.google.android.horologist.networks.highbandwidth.StandardHighBandwidthNetworkMediator$Requests.update (StandardHighBandwidthNetworkMediator.kt:101)
  at com.google.android.horologist.networks.okhttp.impl.HighBandwidthCall$enqueue$1.invokeSuspend (HighBandwidthCall.kt)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:29)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:81)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run (CoroutineScheduler.kt)
```

Please note that the line numbers are not always correct thanks to ... Play Store that remove the information in the stack traces. And to reduce battery usage I've decided to no embed Crashlytics in the Wear app.

While the issue is maybe not caused by Horologist (but my code when requesting Wifi to download media and having some issue that I'll investigate), I think that horologist should catch this error as it's an well known and expected possible issue with the `requestNetwork`  call.

Tolriq:
I'm also seeing 
```
Exception java.lang.IllegalStateException:
  at com.google.android.horologist.networks.highbandwidth.StandardHighBandwidthNetworkMediator$Requests.update (StandardHighBandwidthNetworkMediator.kt:101)
  at com.google.android.horologist.networks.highbandwidth.StandardHighBandwidthNetworkMediator.access$processCancel (StandardHighBandwidthNetworkMediator.kt)
[...]

PR #2424, opened 2024-10-08
Catch exceptions requesting network

yschimke:
#### WHAT

Catch exceptions requesting a high bandwidth network

#### WHY

fixes https://github.com/google/horologist/issues/2417

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/network-awareness/core/api/current.api
+++ b/network-awareness/core/api/current.api
@@ -360,9 +360,9 @@ package com.google.android.horologist.networks.request {
 
   public interface NetworkLease extends java.lang.AutoCloseable {
     method public java.time.Instant getAcquiredAt();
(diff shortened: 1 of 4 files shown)


yschimke:
cc @Tolriq

Tolriq:
Thanks.

Does it make sense to only catch that specific known error? Usually Google does not like large catch all.

yschimke:
@Tolriq I can't catch on the type. I can rethrow if the class name doesn't match.

Thoughts?

Tolriq:
It was more a conceptual best practices question, I have no issue on my side for a catch all.

Tolriq (approved):

yschimke:
@Tolriq Do you need this in 0.6.x or 0.7.x-alpha?

PR #2428, opened 2024-10-17
Refactoring UI models module

niting1112:
#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


media/audio-ui-model/build.gradle.kts
```diff
@@ -0,0 +1,136 @@
...
+        assets.srcDir("src/main/assets")
+    }
+
+    lint {
+        disable += listOf("MissingTranslation", "ExtraTranslation")
+        checkReleaseBuilds = false
+        textReport = true
+    }
+
+    namespace = "com.google.android.horologist.audio.ui"
```
yschimke, line 75: I think this should be unique.

media/audio-ui-model/build.gradle.kts
```diff
@@ -0,0 +1,136 @@
...
+    filename.set("api/current.api")
+    reportLintsAsErrors.set(true)
+}
+
+dependencies {
+    api(projects.media.audio)
+    api(libs.kotlin.stdlib)
+    api(projects.annotations)
+    implementation(projects.composeLayout)
+    implementation(projects.composeMaterial)
```
yschimke, line 98: we shuold strip this down, remove any classes with composables.

media/audio-ui-model/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
yschimke, line 1: Where did all these come from?

media/audio-ui-model/src/main/java/com/google/android/horologist/audio/ui/RotaryVolumeControls.kt
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2023 The Android Open Source Project
+ * Copyright 2024 The Android Open Source Project
```
yschimke, line 2: If it's also classes like this, perhaps media-ui-core is a better name than media-ui-model?

media/audio-ui/build.gradle.kts
```diff
@@ -97,6 +97,7 @@ dependencies {
     implementation(projects.composeLayout)
     implementation(projects.composeMaterial)
     implementation(project(":composables"))
+    implementation(project(":media:audio-ui-model"))
```
yschimke, line 100: can we use the typesafe API here?

```suggestion
    implementation(projects.media.audioUiModel)
```

yschimke: And elsewhere

media/sample/proguard-rules.pro
```diff
@@ -4,3 +4,5 @@
 -keep,allowobfuscation,allowshrinking interface retrofit2.Call
 -keep,allowobfuscation,allowshrinking class retrofit2.Response
 -keep,allowobfuscation,allowshrinking class kotlin.coroutines.Continuation
+# Keep all classes annotated with @Serializable
+-keep @kotlinx.serialization.Serializable class ** { *; }
```
yschimke, line 8: Is this a fix for navigation?

media/ui-model/build.gradle.kts
```diff
@@ -0,0 +1,125 @@
...
+    }
+    lint {
+        checkReleaseBuilds = false
+        textReport = true
+        disable += listOf("MissingTranslation", "ExtraTranslation")
+
+        // https://buganizer.corp.google.com/issues/328279054
+        disable.add("UnsafeOptInUsageError")
+    }
+    namespace = "com.google.android.horologist.media.ui"
```
yschimke, line 87: Should be unique

media/ui/src/test/java/com/google/android/horologist/media/ui/screens/player/PlayerScreenTest.kt
```diff
@@ -71,7 +71,8 @@ class PlayerScreenTest {
         // given
         val playerRepository =
             FakePlayerRepository()
-        val playerViewModel = PlayerViewModel(playerRepository)
+        val playerViewModel =
```
yschimke, line 74: deliberate reformat?

media/ui-model/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
```
yschimke, line 1: Are these used?

media/audio-ui-model/build.gradle.kts
niting1112: Actually I kept it same deliberately to ensure the clients don't have to change the import statements. For e.g. if it is unique then every file (e.g. https://github.com/google/horologist/blob/main/media/ui/src/main/java/com/google/android/horologist/media/ui/screens/player/PlayerScreen.kt#L41)  which references these symbols would need to get updated .

media/audio-ui-model/proguard-rules.pro
niting1112: I think there were created automatically when I created the new module in Android Studio. I will delete these.

media/sample/proguard-rules.pro
niting1112: Yes. When I tried to run the release version of the APK, it was throwing a runtime exception due to missing serializer here https://github.com/google/horologist/blob/main/media/sample/src/main/java/com/google/android/horologist/mediasample/ui/navigation/UampNavigationScreen.kt#L38

This issue did not happen in the debug build though.

media/ui/src/test/java/com/google/android/horologist/media/ui/screens/player/PlayerScreenTest.kt
niting1112: I guess this also was done by android studio. I will remove it.

media/ui-model/proguard-rules.pro
niting1112: No. I will remove them. Actually they got created automatically by Android Studio.

media/audio-ui-model/build.gradle.kts
yschimke: This doesn't affect that. This only affects the manifest package

```
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:tools="http://schemas.android.com/tools"
    package="com.google.android.horologist.audio.ui">
```

source code packages are separate

media/audio-ui-model/src/main/java/com/google/android/horologist/audio/ui/RotaryVolumeControls.kt
yschimke: I think it's reusable, so maybe we do want to share this. Your call.

media/audio-ui-model/build.gradle.kts
niting1112: Oh..yes sure..it got added because of RotaryVolumeControls. But I have reverted it to original module.

media/audio-ui-model/build.gradle.kts
niting1112: Sure..Thanks for the info! Updated!

media/audio-ui-model/src/main/java/com/google/android/horologist/audio/ui/RotaryVolumeControls.kt
niting1112: Sure, for now I think we can keep the non-composable part as this is the only file.

yschimke (approved):

media/sample/proguard-rules.pro
yschimke: It makes sense it's release only.

But we shouldn't need to do anything https://github.com/Kotlin/kotlinx.serialization/blob/master/rules/common.pro

But thanks for fixing.

PR #2431, opened 2024-10-21
Demonstrate an alternative architectural approach

JolandaVerhoef:
Approach is hard-linked with the Media3 `Player` interface.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
This looks promising.

If you go to https://github.com/JolandaVerhoef/horologist/actions/workflows/fixup.yml, and run against your branch, it should fix API, screenshots and some lint.

yschimke:
@JolandaVerhoef nice, this is looking promising.

media/sample/src/main/java/com/google/android/horologist/mediasample/ui/experimentaluimode/ExperimentalUiModePlayerScreen.kt
```diff
@@ -14,26 +14,21 @@
...
 
 @Composable
-fun NewHotnessPlayerScreen(
-    newHotnessPlayerScreenViewModel: NewHotnessPlayerScreenViewModel = hiltViewModel(),
+fun ExperimentalUiModePlayerScreen(
+    modifier: Modifier = Modifier,
+    experimentalUiModePlayerScreenViewModel: ExperimentalUiModePlayerScreenViewModel = hiltViewModel(),
 ) {
-    val player: Player? by newHotnessPlayerScreenViewModel.player.collectAsStateWithLifecycle()
+    val player by experimentalUiModePlayerScreenViewModel.player.collectAsStateWithLifecycle()
```
yschimke, line 31: I feel like we should use the null pattern* to handle this outside the screen. And rely on the ViewModel to cache the current player on resuming the player activity.

So it's never null, you can always show something, but no media will be shown, no actions enabled, and the screen can check and use that as a loading state if it wants.

* https://en.wikipedia.org/wiki/Null_object_pattern

PR #2432, opened 2024-10-23
Update Player screen margins

martenxi:
Update Player Screen margins to match new UX specs.

This is primarily to allow the bottom buttons of the Player Screen to not sit on the border of the tap target.


yschimke (approved):

media/ui/src/test/screenshots/MediaPlayerFontScaleTest_small_round_large.png
yschimke, line 1: This looks clipped.

yschimke: FYI @kul3r4 it sounds like some following fixes with correct this, but needed for an import.

PR #2435, opened 2024-10-24
Update player screen margins

martenxi:


yschimke:
Dupe of https://github.com/google/horologist/pull/2432?

PR #2436, opened 2024-10-24
Update pager spec

yschimke:
#### WHAT

Avoid jank on fling.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


amankhoza (approved):
lgtm

PR #2439, opened 2024-10-28
Use wear specific pager to remove edgeSwipeToDismiss

kul3r4:
#### WHAT
use the Wear foundation pager

#### WHY
Simplify the code: it includes by default the swipe to dismiss

#### HOW
use the wear component instead of the phone component

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke (approved):
Good idea.

PR #2463, opened 2024-11-21
Use Roborazzi A11y checks

yschimke:
#### WHAT

This change updates the handling of accessibility checks for off-screen elements in the UI tests. It suppresses accessibility errors for elements which are off-screen but are intended to be scrolled on-screen.

This improves the accuracy and efficiency of the accessibility tests by ignoring false-positives related to these elements.

Additionally, it removes unnecessary custom shadow classes that were previously used for accessibility testing. These shadows caused conflicts and were replaced by existing Roborazzi features. This simplifies the testing setup and relies on external libraries for better maintainability.

Finally, it updates the version of the Roborazzi library to bring in the new accessibility checking capabilities. This allows the tests to use newer, more robust and accurate accessibility checking features.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
@garanj landing as test only, will follow up on review comments.

PR #2471, opened 2024-12-03
Add provision for custom package name while launching output switcher

anubhavkakkar:
#### WHAT
Adds the provision to provide package name of app corresponding to which output switcher UI should be displayed

#### WHY
Starting with API35, the list of output devices displayed in the output switcher's UI may vary with the package name. Therefore, the audio repository APIs should have the provision to configure the package name.

#### HOW
The output switcher launcher API will now take-in the calling package name as the parameter that will be used to launch the output switcher.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media/core/src/main/java/com/google/android/horologist/media/model/Media.kt
+++ b/media/core/src/main/java/com/google/android/horologist/media/model/Media.kt
@@ -27,6 +27,7 @@ public data class Media(
     val uri: String,
     val title: String,
     val artist: String,
+    val clientPackageName: String? = null,
     val artworkUri: String? = null,
     val artworkColor: Int? = null,
     val extras: Map<String, Any> = emptyMap(),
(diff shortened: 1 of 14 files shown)


media/core/src/main/java/com/google/android/horologist/media/model/Media.kt
```diff
@@ -27,6 +27,7 @@ public data class Media(
     val uri: String,
     val title: String,
     val artist: String,
+    val clientPackageName: String? = null,
```
yschimke, line 30: Is client the accurate label? If my app (UI) exposes other media service (playback). Is this the UI app or the playback app?

yschimke (approved):

media/core/src/main/java/com/google/android/horologist/media/model/Media.kt
anubhavkakkar: This data class is used by the apps which own their media session themselves as well as by those that display the UI for the media session owned by other apps. 

The name signifies that it should be the name of the package that owns the media session.

yschimke: That's not the client then is it?

PR #2472, opened 2024-12-04
Refactor: AmbientAware API change

yschimke:
#### WHAT

This change refactors the `AmbientAware` API to make it more intuitive and easier to use.

The following changes were made:

- The `AmbientAware` composable moves to the screen level
- The `AmbientStateUpdate` class was merged into `AmbientState`.
- The `AmbientState` sealed interface now has a data object `Interactive` and a data class `Ambient` to store AmbientDetails, as well as Inactive.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
@@ -20,8 +20,9 @@ import android.app.Activity
 import android.content.Context
 import android.content.ContextWrapper
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.DisposableEffect
-import androidx.compose.runtime.State
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.SideEffect
+import androidx.compose.runtime.compositionLocalOf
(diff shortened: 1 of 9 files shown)


garanj (approved):

kul3r4 (approved):

yschimke:
@garanj @kul3r4 Can you take a look again, it's much simpler. If we ignore some of the weirdnesses with two things on screen, I think we can just use the current nav lifecycle to control when it's enabled.

Won't work for a pager, but maybe that's ok?

compose-layout/src/main/java/com/google/android/horologist/compose/ambient/AmbientAware.kt
```diff
@@ -20,87 +20,94 @@ import android.app.Activity
...
-                    AmbientState.Ambient(null)
-                } else {
-                    AmbientState.Interactive
-                }
-                ambientUpdate = AmbientStateUpdate(initialAmbientState)
-            }
-        }
+    val observer = remember {
+        if (activity != null) {
+            println("Creating observer")
```
yschimke, line 61: TODO remove

PR #2481, opened 2024-12-09
 Add lifecycle parameter to AmbientAware

yschimke:
#### WHAT

This change adds an optional `lifecycle` parameter to the `AmbientAware` composable function, allowing developers to specify a custom lifecycle instead of relying on the `LocalLifecycleOwner`.

This provides more flexibility when using `AmbientAware` in different contexts, such as within individual screens in navigation routes. If not specified, it defaults to the current `LocalLifecycleOwner`'s lifecycle.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/api/current.api
+++ b/compose-layout/api/current.api
@@ -2,7 +2,7 @@
 package com.google.android.horologist.compose.ambient {
 
   public final class AmbientAwareKt {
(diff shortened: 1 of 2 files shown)


kul3r4 (approved):
Can we backport those to latest stable version of Horologist?

PR #2498, opened 2025-01-15
Modify MediaUiModel and AudioOutput to support remote media output an…

anubhavkakkar:
Modify MediaUiModel and AudioOutput to support remote media output and artwork seed color.

#### WHAT
The `MediaUiModel` data class now includes properties for the currently selected media output and artwork color seed. The `AudioOutput` interface now uses the Android Media Router type for the represented media route. An AudioOutput implementation for remote media output has also been added.

#### WHY
The `MediaUiModel`'s added properties will help build the related UI. The provision for `MediaRoute2Info.getType` in the AudioOutput will aid in recognizing the different types of media output supported by the Android Media Router.

#### HOW
The modifications are directly made in the MediaUiModel and AudioOutput classes.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media/ui-model/src/main/java/com/google/android/horologist/media/ui/state/model/MediaUiModel.kt
+++ b/media/ui-model/src/main/java/com/google/android/horologist/media/ui/state/model/MediaUiModel.kt
@@ -18,10 +18,26 @@ package com.google.android.horologist.media.ui.state.model
 
 import androidx.compose.ui.graphics.Color
 import com.google.android.horologist.annotations.ExperimentalHorologistApi
+import com.google.android.horologist.audio.AudioOutput
 import com.google.android.horologist.images.base.paintable.Paintable
 
 @ExperimentalHorologistApi
+/** A sealed class representing the UI state of a media item. */
 public sealed class MediaUiModel {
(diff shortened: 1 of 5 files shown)


media/ui-model/src/main/java/com/google/android/horologist/media/ui/state/model/MediaUiModel.kt
```diff
@@ -18,18 +18,36 @@ package com.google.android.horologist.media.ui.state.model
...
+
+    /**
+     * Represents the UI state when media data is available.
+     *
+     * @param id The unique identifier of the media item.
+     * @param title The title of the media item.
+     * @param subtitle The subtitle of the media item (optional).
+     * @param artwork The artwork to display for the media item (optional).
+     * @param artworkColor The primary color to use for the artwork background (optional).
+     * @param artworkColorSeed The seed color to use for generating the artwork color pallet
```
yschimke, line 36:
```suggestion
     * @param artworkColorSeed The seed color to use for generating the artwork color palette
```

PR #2520, opened 2025-02-15
Add coroutines executeAsync extension to OkHttp in sample

yschimke:
#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


network-awareness/okhttp/src/test/java/com/google/android/horologist/networks/rules/NetworkSelectingCallFactoryTest.kt
```diff
@@ -160,32 +157,14 @@ class NetworkSelectingCallFactoryTest {
...
-        val thrown = assertThrows(IOException::class.java) {
-            callFactory.newCall(request).executeAsync()
+        val responseAsync = async { callFactory.newCall(request).executeAsync() }
+        val result = runCatching {
+            responseAsync.await()
         }
 
-        assertThat(thrown).hasMessageThat().isEqualTo("Unable to use BT for media-download")
+        assertThat(result.isFailure).isTrue()
+        assertThat(result.exceptionOrNull()).hasMessageThat().isEqualTo("Unable to use BT for media-download")
```
luizgrp, line 166: nit: should it also assert the exception instance type of was it intentionally to relax it and only check the message

luizgrp (approved):

network-awareness/okhttp/src/test/java/com/google/android/horologist/networks/rules/NetworkSelectingCallFactoryTest.kt
yschimke: good catch

PR #2531, opened 2025-03-01
Use combinedClickable as basis of UnboundedRippleButton

kevinGC:
#### WHAT
Allow users of user of UnboundedRippleButton to use features like long press.

#### WHY
Long press is useful for UI interactions, e.g. [for a button that skips forward 30 seconds upon clicking, and skips to the end when held](https://github.com/Automattic/pocket-casts-android/issues/3634).

#### HOW
Use `combinedClickable` instead of `clickable`.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/composables/api/current.api
+++ b/composables/api/current.api
@@ -161,7 +161,7 @@ package com.google.android.horologist.composables {
   }
 
   public final class UnboundedRippleButtonKt {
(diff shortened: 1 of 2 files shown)


yschimke:
This isn't passing in any additional lambda, so not sure it's letting anyone pass this in.

kevinGC:
Friendly ping, please let me know if there's anything still needing to be addressed.

yschimke (approved):

yschimke:
Nice, thanks for the PR. I'll try to get a release done by Monday or so.

PR #2581, opened 2025-04-24
Move string resources of media ui to common model module

rajat4064g:
#### WHAT
Move string resources of media ui to common model module

#### WHY
Common string resource files will be used for material and material3 versions of media module

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


media/ui-material3/build.gradle.kts
```diff
@@ -67,6 +67,7 @@ android {
 
     sourceSets.getByName("main") {
         assets.srcDir("src/main/assets")
+        res.srcDirs("src/main/res", "../ui-model/src/main/res")
```
yschimke, line 70: Shouldn't you get them as an R file from the other module?

yschimke (approved):

PR #2611, opened 2025-05-28
Fix recompositions issue in VolumeScreen.

Kpeved:
Changes in volumeUiState were triggering recompositions which resulted in missing animation transition of LevelIndicator.

#### WHAT
StepperLevelIndicator was not properly used, which was triggering extra recompositions . That resulted in transition animation not properly played.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/media/audio-ui-material3/src/main/java/com/google/android/horologist/audio/ui/material3/VolumeLevelIndicator.kt
+++ b/media/audio-ui-material3/src/main/java/com/google/android/horologist/audio/ui/material3/VolumeLevelIndicator.kt
@@ -61,7 +61,6 @@ public fun VolumeLevelIndicator(
                 value = false
             }
         }
-    val uiState = volumeUiState()
 
     AnimatedVisibility(
         visible = visible,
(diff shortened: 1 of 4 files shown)


media/audio-ui-material3/src/main/java/com/google/android/horologist/audio/ui/material3/VolumeLevelIndicator.kt
```diff
@@ -70,8 +69,8 @@ public fun VolumeLevelIndicator(
     ) {
         StepperLevelIndicator(
             modifier = modifier.fillMaxSize(),
-            value = { uiState.current.toFloat() },
```
yschimke, line 73: Why is this not using it properly?

This looks like working around recomposition bugs in StepperLevelIndicator.

Kpeved: Because calling  `val uiState = volumeUiState()` triggers recomposition, as the state every time is different. ( it's immutable)

PR #2617, opened 2025-06-09
Reduce the opacity to 0.12f from the existing 0.16f for player settin…

anubhavkakkar:
…gs button

#### WHAT
Reduce the opacity to 0.12f from the existing 0.16f

#### WHY
Background of disabled buttons is currently not much different from active buttons

#### HOW
Changed the background opacity

#### Checklist :clipboard:
- [x ] Add explicit visibility modifier and explicit return types for public declarations
- [ x] Run spotless check
- [ x] Run tests
- [x ] Update metalava's signature text files


yschimke (changes requested):
Add a screenshot test including these buttons.

anubhavkakkar:
A new PR will be raised along with other screenshot changes.

PR #2625, opened 2025-06-17
Fix position and appearances of title and subtitle in ambient mode.

anubhavkakkar:
#### WHAT
Fix position and appearances of title and subtitle in ambient mode.

#### WHY
Position of title & subtitle currently does not match with interactive mode. Also, the ellipse towards end of the title & subtitle is not expected.

#### HOW
Modified the Text property and added spacer towards beginning of the title.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media/ui-material3/src/main/java/com/google/android/horologist/media/ui/material3/components/display/TextMediaDisplay.kt
+++ b/media/ui-material3/src/main/java/com/google/android/horologist/media/ui/material3/components/display/TextMediaDisplay.kt
@@ -26,10 +26,20 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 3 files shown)


media/ui-material3/src/main/java/com/google/android/horologist/media/ui/material3/components/display/TextMediaDisplay.kt
```diff
@@ -48,45 +58,63 @@ public fun TextMediaDisplay(
     modifier: Modifier = Modifier,
     titleIcon: Paintable? = null,
     colorScheme: ColorScheme = MaterialTheme.colorScheme,
+    titleOverflow: TextOverflow = TextOverflow.Ellipsis,
+    subtitleOverflow: TextOverflow = TextOverflow.Ellipsis,
+    titleSoftWrap: Boolean = true,
+    subtitleSoftWrap: Boolean = true,
 ) {
+    val textMeasurer = rememberTextMeasurer()
+    var measuredTextWidth by remember { mutableStateOf(0.dp) }
```
yschimke, line 67: I'm not sure you need to remember `textMeasurer.measure` as it's cached internally. Every Text calls this inline anyway.

If you definitely do, I think cleaner to remember keyed directly on the changeable fields, and avoid 0 as default.

yschimke (approved):
After reviewing remember comment.

anubhavkakkar:
Oh, I missed this comment.
Let me address that in a follow-up CL where I intend to add gradient.

PR #2626, opened 2025-06-18
Add gradient for title text in ambient mode.

anubhavkakkar:
#### WHAT
Add gradient for title text in ambient mode.

#### WHY
The interactive mode has gradient for the title. The text in ambient should be same.

#### HOW
Added the gradient at the end

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/media/ui-material3/src/main/java/com/google/android/horologist/media/ui/material3/components/display/TextMediaDisplay.kt
+++ b/media/ui-material3/src/main/java/com/google/android/horologist/media/ui/material3/components/display/TextMediaDisplay.kt
@@ -26,19 +26,21 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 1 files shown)


media/ui-material3/src/main/java/com/google/android/horologist/media/ui/material3/components/display/TextMediaDisplay.kt
```diff
@@ -64,18 +65,16 @@ public fun TextMediaDisplay(
     subtitleSoftWrap: Boolean = true,
 ) {
     val textMeasurer = rememberTextMeasurer()
```
anubhavkakkar, line 67: Should we replace `rememberTextMeasurer` with ` TextMeasurer(LocalFontFamilyResolver.current, LocalDensity.current, LocalLayoutDirection.current)`?

I see standard usage of `rememberTextMeasurer` in MarqueeText.

yschimke: I think you need `rememberTextMeasurer` so it caches results in each measure call.  So no.

yschimke:
@anubhavkakkar submitted with build failing, fix here https://github.com/google/horologist/pull/2630

PR #2643, opened 2025-07-04
Migrate main screens of media sample to Material3 except settings

kul3r4:
#### WHAT
Migrate most of the screens to Material 3

#### WHY
To update the sample to the latest recommended UI

#### HOW
use the material3-ui packages + use material 3 navigation

There are some issues with the material3 navigation Horologist package, such as not being Serializable, which prevents code such as `private val route = savedStateHandle.toRoute<NavigationScreen.Collection>()` to work properly, hence this PR needs more work.
Moreover, screens from Settings haven't been migrated and the Theme has not been updated.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


yschimke:
would need to be rebased, closing for now

PR #2667, opened 2025-07-28
Fixups of screenshot tests & adds some bug fixes and speed ups to the FastScrollingTransformingLazyColumn

brysonPaul:
#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


brysonPaul (commented):
Confused why this test is failing now... Seems to be gradle-related though as opposed to this change, and the other WFs seem to working better.

Will do some more debugging of this tomorrow,  but have you seen this problem before Yuri?

yschimke:
I'd expect this branch to be a minimal set of commits. You might want to move this to another branch. Then reset your main branch to match google.

yschimke:
The test failure seems a transient network error. So maybe resolved with a re-run.

But I don't see that sort of thing often, so not sure why?

brysonPaul:
Going to close this PR and make another with the new changes from the new branch.

PR #2668, opened 2025-07-29
using RotaryScrollableBehavior to perform RSB actions + speedups to the FastScrollingTLC

brysonPaul:
#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/layout/m3/FastScrollingTransformingLazyColumn.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/layout/m3/FastScrollingTransformingLazyColumn.kt
@@ -50,8 +50,6 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
-import androidx.compose.ui.focus.FocusRequester
-import androidx.compose.ui.focus.focusRequester
 import androidx.compose.ui.graphics.graphicsLayer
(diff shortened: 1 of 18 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/layout/m3/FastScrollingTransformingLazyColumn.kt
```diff
@@ -306,48 +303,54 @@ public fun FastScrollingTransformingLazyColumn(
...
-                            rsbScrollCount += scrollCount
-                            if (rsbScrollCount > 5) {
-                                isFirstFastScroll = true
-                                pixelsScrolledBy = 0f
-                                isSkimming = true
-                                rsbScrollCount = 0
-                            }
+            rotaryScrollableBehavior = object : RotaryScrollableBehavior {
+                override suspend fun CoroutineScope.performScroll(
+                    timestampMillis: Long,
```
yschimke, line 308: For the sake of testing should you use this instead of currentTimeMillis?

yschimke (approved):

compose-layout/src/main/java/com/google/android/horologist/compose/layout/m3/FastScrollingTransformingLazyColumn.kt
brysonPaul: Just tried this, it seems like using timeStampMillis is more stale causing the scroll to be unresposive when trying to scroll to the next header. Will do some more investigation on why as this would fit great.

PR #2674, opened 2025-08-14
Add API for capturing composable in Tiles

yschimke:
#### WHAT

API for generating a Tile resource from a Composable.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


castedmo (approved):
This is great. Thanks!

PR #2676, opened 2025-08-15
Cleanup the coroutines code in CaptureComposable.kt

yschimke:
#### WHAT

Split out functions to separate layers of abstraction (Compose, Coroutines, Normal).

#### WHY

Got some warnings about mixing suspend and non suspend methods. False alarm due to nesting, but removing that confusion seems worthwhile. 

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/tiles/src/main/java/com/google/android/horologist/tiles/composable/CaptureComposable.kt
+++ b/tiles/src/main/java/com/google/android/horologist/tiles/composable/CaptureComposable.kt
@@ -17,30 +17,26 @@
 package com.google.android.horologist.tiles.composable
 
 import android.app.Application
+import android.app.Dialog
 import android.app.Presentation
 import android.content.Context
-import android.content.Context.DISPLAY_SERVICE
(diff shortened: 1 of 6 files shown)


luizgrp (approved):

tiles/src/main/java/com/google/android/horologist/tiles/composable/CaptureComposable.kt
```diff
@@ -78,18 +78,18 @@ public interface ComposableBitmapRenderer {
...
                     "virtualDisplay",
                     1,
                     1,
                     72,
                     surface,
                     DisplayManager.VIRTUAL_DISPLAY_FLAG_OWN_CONTENT_ONLY,
-                )
+                ) ?: return null
 
             val result = callback(virtualDisplay!!.display)
```
castedmo, line 94: `!!` is now redundant here I guess.

PR #2683, opened 2025-08-20
Support configurable encoding

yschimke:
#### WHAT

Support configurable encoding for tile images

#### WHY

Avoid double encoding

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


castedmo (approved):
Thanks!

PR #2696, opened 2025-09-16
Adds yields to animations to ensure subcompositions during composition n errors do not occur, and lowers timeout to leave skimming mode.

brysonPaul:
…
#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


compose-layout/src/main/java/com/google/android/horologist/compose/layout/m3/FastScrollingTransformingLazyColumn.kt
```diff
@@ -176,6 +177,7 @@ public fun FastScrollingTransformingLazyColumn(
             // We run animateScrollBy with a movement of 0 just to remove the timeText from the screen and
             // show the position indicators, as animateScrollToItem will fling from each section.
             state.animateScrollBy(0f)
+            yield()
```
yschimke, line 180: is there existing precedent for this in AndroidX compose? 

No issue landing, if it fixes a real issue, but maybe there is some other pattern we should be using?

yschimke (approved):
Approving with any yield changes for follow up

PR #2697, opened 2025-09-22
Add buttons for SignInPrompt screen

jason-sigma:
#### WHAT
Add buttons for SignInPrompt screen

#### Checklist :clipboard:
- [X] Run spotless check
- [x] Run tests


luizgrp (approved):
Thanks!

PR #2698, opened 2025-10-01
Adds velocity-based checks to go into skimming mode instead of basing it off of the distance scrolled in a given scroll event.

brysonPaul:
#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/compose-layout/src/main/java/com/google/android/horologist/compose/layout/m3/FastScrollingTransformingLazyColumn.kt
+++ b/compose-layout/src/main/java/com/google/android/horologist/compose/layout/m3/FastScrollingTransformingLazyColumn.kt
@@ -57,7 +57,6 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.graphicsLayer
 import androidx.compose.ui.hapticfeedback.HapticFeedbackType
-import androidx.compose.ui.input.pointer.util.VelocityTracker1D
 import androidx.compose.ui.platform.LocalContext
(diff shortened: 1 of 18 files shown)


compose-layout/src/main/java/com/google/android/horologist/compose/layout/m3/FastScrollingTransformingLazyColumn.kt
```diff
@@ -264,34 +273,39 @@ public fun FastScrollingTransformingLazyColumn(
                 ) {
                     // Track current time to ensure that we should ingest the rotary scroll event.
                     val currentTime = System.currentTimeMillis()
+                    if (currentTime - lastRotaryScrollTime > 750L) {
```
yschimke, line 276: Add a constant?

compose-layout/src/main/java/com/google/android/horologist/compose/layout/m3/FastScrollingTransformingLazyColumn.kt
```diff
@@ -264,34 +273,39 @@ public fun FastScrollingTransformingLazyColumn(
...
+                        // We reset tracking if there was too much time between rotary scroll events
+                        velocityTracker.resetTracking()
+                        velocityTracker.addDataPoint(currentTime, delta)
+                        lastRotaryScrollTime = currentTime
```
yschimke, line 280: Why isn't this also updated in the other case?

yschimke (approved):

PR #2711, opened 2025-12-09
Update scroll offset to what it was initially

brysonPaul:
###WHAT

this was changed by someone else in g3, causing the header to be a bit more unpredictable on scrolls.


yschimke:
Will land and fix on main

PR #2745, opened 2026-07-21
Add Lottie tests

MarkYav:
Add Lottie deserialization and animation interpolation unit tests to :remotecompose:lottie.

This will ensure that the library displays the Lottie animation the right way.

Details:
- Add ParsingTest.kt to verify Moshi JSON parsing of Lottie structures.
- Add AnimationTest.kt to test vector keyframe animation evaluation.
- Add heartbeat.json test asset in src/test/resources.
- Configure KSP Moshi code generator and unit test dependencies in build.gradle.kts.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


remotecompose/lottie/src/test/resources/heartbeat.json
yschimke, line 1: Can we start with the gear icon?

https://skottie.skia.org/?h=128&w=128

PR #2771, opened 2026-08-10
Upstream media UI and audio UI improvements 

yschimke:
#### WHAT

- VolumeLevelIndicator: Reset initial state in produceState on key change
- RotatingWavyProgressIndicator: Cache Path, MutableCubic, PathMeasure, and Stroke to reduce allocations
- AnimatedMediaButton, AnimatedSeekToNextButton, AnimatedSeekToPreviousButton: Allow custom content slot and expose content composables
- CustomActionMediaButton: Support custom content slot for configurable gestures
- RotaryVolumeControls: Expose performHapticFeedback as internal

#### WHY

Keep source code in sync.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


media/ui-material3/src/main/java/com/google/android/horologist/media/ui/material3/components/animated/RotatingWavyProgressIndicator.kt
```diff
@@ -74,8 +75,8 @@ internal fun RotatingWavyProgressIndicator(
   shapeMorphProgress: State<Float>,
   rotationProgress: State<Float>,
   strokeWidth: Dp,
-  modifier: Modifier = Modifier,
   colorScheme: ColorScheme = MaterialTheme.colorScheme,
+  modifier: Modifier = Modifier,
```
luizgrp, line 79: nit: it should be the first optional

media/ui-material3/src/main/java/com/google/android/horologist/media/ui/material3/components/CustomActionMediaButton.kt
```diff
@@ -53,6 +54,7 @@ import com.google.android.horologist.media.ui.material3.composables.UnboundedRip
  *   emitting [Interaction]s for this button (Optional).
  * @param iconSize The size of the icon. Defaults to [IconButtonDefaults.SmallIconSize] (Optional).
  * @param border [BorderStroke] to be applied to the button. If null, no border is drawn (Optional).
+ * @param content A composable content to be used instead of the default one (Optional).
```
luizgrp, line 57: nit: if content is passed, icon won't be used, but it is still a required param. It would be better to have two functions.

luizgrp (approved):

media/ui-material3/src/main/java/com/google/android/horologist/media/ui/material3/components/animated/RotatingWavyProgressIndicator.kt
yschimke: had auto submit on, but I will follow up with fixes.

PR #2774, opened 2026-08-10
Remove ambient compose library

yschimke:
#### WHAT

Remove ambient compose

#### WHY

Now provbided by Wear Compose

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


garanj (approved):
Thanks Yuri!

PR #2781, opened 2026-08-11
Support Bezier curve

MarkYav:
#### WHAT
Adds full Remote Compose Bezier spline animation support by replacing static path translations with dynamic expression tree `RemoteFloat` evaluations. Furthermore, completely restructures the underlying Keyframe/Bezier rendering classes into single-responsibility, SLAP-enforced modular files.

#### WHY
Previously, bezier path lookups used static cubic easing arrays parsed from Lottie keyframes. To correctly enable state-driven multi-node spline paths (i.e. morphing Lottie vector geometries dynamically through variables on wear targets), Remote Compose paths required fully-typed `RemoteFloat` chained logic translating easing tangents seamlessly. The structural refactor addresses code density rules, moving monolithic handlers into dry helpers (less than 25 lines/method) which keeps the interpolation math legible and sustainable long-term.

#### HOW
- **Interpolation Engine:** Migrates `RemoteLottiePath` generation strictly onto `RemoteFloat` evaluations relying heavily on native `cubicEasing` operations instead of simple interpolated float arrays.
- **Node Chaining:** Built `chainBezierAnimation`, enabling deep array segments to execute Remote Compose property conditional cascades cleanly over timeline boundaries.
- **Topology Resiliency:** Closed the rendering loops gracefully inside `gatherShapes` parsing vectors rather than singular objects, stopping runtime open path geometry boundary failures (`IndexOutOfBoundsException`) inherently.
- **Decoupled Architecture:** Replaced the monolithic `Animation.kt` core with 4 explicitly-scoped responsibilities (`AnimationUtils.kt`, `PositionAnimation.kt`, `VectorAnimation.kt`, `BezierAnimation.kt`) all abiding strictly by SLAP paradigms (< 25-line ceilings by isolating low-level initialization routines in cleanly named builder methods).
- Encapsulated inner iterative draw segments into an extension block via `RemotePathScope.drawSplinePaths`.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/RemoteShape.kt
```diff
@@ -29,10 +29,32 @@ internal interface RemoteShape {
...
 
 @SuppressLint("RestrictedApi")
-internal class RemoteLottiePath(val path: RemotePath) : RemoteShape {
+internal class RemoteLottiePath(val bezierValue: RemoteBezierValue) : RemoteShape {
   override fun draw(drawScope: RemoteDrawScope, canvas: RemoteCanvas) {
+    val path = drawScope.remotePath { drawSplinePaths(bezierValue) }
     canvas.drawPath(path)
   }
+
+  private fun androidx.compose.remote.creation.compose.vector.RemotePathScope.drawSplinePaths(
```
yschimke: fix imports.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/Shape.kt
```diff
@@ -111,39 +110,20 @@ private fun group(group: Group, animationSettings: LottieSettings): RemoteGroup?
...
-    val p0 = vertices[i]
-    val lastIndex = if (i == vertices.size - 1 && path.closed) 0 else i + 1
-    val p4 = vertices[lastIndex]
-    val inTangent = inTangents[lastIndex]
-    val outTangent = outTangents[i]
-    val p1 = listOf(p0[0] + outTangent[0], p0[1] + outTangent[1])
-    val p2 = listOf(p4[0] + inTangent[0], p4[1] + inTangent[1])
-
-    rcPath.cubicTo(p1[0], p1[1], p2[0], p2[1], p4[0], p4[1])
+private fun path(lottiePath: Path, animationSettings: LottieSettings): List<RemoteLottiePath> {
```
yschimke, line 113: If we are splitting to a List of paths, what happens with any effects like caps on the start or the ends? do we treat these differently after this?

yschimke (commented):
I'm unclear if it's passing tests because it produces identical output, or because we don't have good existing coverage of these tests.

Can you add a screenshot test covering this.

Also, are there any other unit tests we should consider for various cases.

PR #2800, opened 2026-08-17
Lottie grandchild layer

### LINKED ISSUE #2795: Grandparent layer transformation is lost
Passed parameters:
- **layer**: the current Lottie layer to display.
- **parentTransforms**: a map (containing all layers with non-NULL id and transformation) from layer ID to its own single Transform, applied to this exact layer.
- **transform**: the transformation for the current layer.

Note: a transformation could be rotation of a layer, a shift by a coordinate, or scaling. Or a combination of these.

https://github.com/google/horologist/blob/64b50708af4da9f710aa9fceee91bb63fbe9c020/remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/Layer.kt#L26-L48

If we have Layer C (child) -> parented to Layer B (parent) -> parented to Layer A (grandparent): when Layer C is drawn, it will fetch Layer B's transform. Then, it will attach its own transform. The list sent to the canvas will be [LayerC, LayerB]. Layer A's transform is completely lost.

### Option 1. `while` loop to the root
Instead of just checking `parentTransforms[layer.parent]`, we need a while loop that climbs the ancestry tree until it hits the root:
```kotlin
    val transformStack = mutableListOf<Transform>()                                          
    var currentParentId = layer.parent                                                       
                                                                                             
    // Keep climbing the family tree and collecting transforms!                              
    while (currentParentId != null) {                                                        
        val parentLayer = animation.layers.find { it.index == currentParentId }              
        if (parentLayer?.transform != null) {                                                
            transformStack.add(parentLayer.transform)                                        
        }                                                                                    
        currentParentId = parentLayer?.parent // move up to the grandparent                  
    }  
```
Pros: simple implementation, minimal changes.
Cons: we may pass through the same layers multiple times.

### Option 2. Pre-computing the parent layer transformation
Use an algorithm, similar to a topological sort, to include the parent transformation into the child transformation while constructing the `parentTransforms`.

Pros: pass through the layers only once, transformation will be pre-computed --> no need for a mutable list --> simpler `internal fun Layer(...)` method.
Cons: more complex implementation (topological sort).

## Implementing TODO: using `Modifier` instead of  `Transform`
Considering the TODO regarding the `graphicsLayer`, it is connected to this refactoring. The idea is similar to the Option 2: we have to go through layers as a tree, not flat-loop. But instead of joining transformations (multiplying matrices) we treat each layer as a `Box(...)` composable function and apply a modifier to it.

Current implementation in `LottieAnimation.kt`:
https://github.com/google/horologist/blob/286324b6ea7c3e9f1d10a4c397d8f850e529797a/remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/LottieAnimation.kt#L136-L152

Suggested change:
```kotlin
    val childrenMap = animation.layers.groupBy { it.parent } // Maps ParentID -> List<Child Layers>                                                                                    
                                                                                             
    RemoteBox(...) {                                                                         
        // Only kick off the rendering for Root layers (parent == null)                      
        val rootLayers = childrenMap[null] ?: emptyList()                                    
        for (layer in rootLayers) {                                                          
            LayerNode(layer, childrenMap)                                                    
        }                                                                                    
    } 
```

In the updated `Layer(...)` function, we create a composable that applies its own graphicsLayer, renders itself, and then recursively places its children inside its own Box:
```kotlin
    @Composable                                                                              
    @RemoteComposable                                                                        
    internal fun LayerNode( // suggestion: rename Layer to LayerNode                                                                 
        layer: Layer,                                                                        
        childrenMap: Map<Int?, List<Layer>>                                                  
    ) {                                                                                      
        // 1. Convert exactly ONE Lottie Transform to a Compose Modifier                     
        val modifier = if (layer.transform != null) {                                        
            RemoteModifier.graphicsLayer(                                                    
                scaleX = layer.transform.scaleX,                                             
                scaleY = layer.transform.scaleY,                                             
                translationX = layer.transform.translationX,                                 
                translationY = layer.transform.translationY,                                 
                rotationZ = layer.transform.rotation                                         
            )                                                                                
        } else RemoteModifier                                                                
                                                                                             
        // 2. Wrap this specific layer AND all of its children in this modifier!             
        RemoteBox(modifier = modifier) {                                                     
                                                                                             
            // Draw the local shape (it no longer needs ANY transformStack logic)            
            if (layer.type == LayerType.Shape) {                                             
                ShapeLayer(layer as Layer.ShapeLayer) // No stack passed!                    
            }                                                                                
                                                                                             
            // 3. Recursively draw all children INSIDE this RemoteBox.                       
            // Compose's graphicsLayer will now natively affect them all!                    
            val myChildren = childrenMap[layer.index] ?: emptyList()                         
            for (child in myChildren) {                                                      
                 LayerNode(child, childrenMap)                                               
            }                                                                                
        }                                                                                    
    }   
```

MarkYav:
Fixes https://github.com/google/horologist/issues/2795

#### WHAT

Fix missing nested layer transforms (grandparent transformation) in RemoteCompose Lottie Player by migrating to hierarchical `RemoteModifier.graphicsLayer` nodes.

#### WHY

Previously, `LottieAnimation` processed layers sequentially in a flat list and only injected the immediate parent's transform, discarding any transformations applied further up the tree (e.g., grandparents). This caused multi-level hierarchical transforms to be completely lost. Furthermore, the underlying Remote Compose player now natively supports `graphicsLayer` (resolving b/408913726). Therefore, the temporary workaround of passing a list of `transformStack` matrices deep into the raw `RemoteCanvas` drawing step is obsolete and can be cleanly offloaded to the Compose engine.

#### HOW

- `LottieAnimation.kt`: Replaced the flat layer loop by grouping layers into a `childrenMap` (by parent). The `RemoteBox` now only invokes the root layers (where `parent == null`).
- `Layer.kt`: Evolved Layer into a recursive `Composable` component `LayerNode`. Each node now recursively launches its own children from the `childrenMap`.
- `Layer.kt`: Eliminated the legacy `transformStack` list. `LayerNode` directly applies its local transform to a `RemoteBox(modifier = ...)`, allowing standard Compose bounds to flawlessly compound nested matrices down the UI tree.
- `Transform.kt`: Added a `Transform.toModifier()` extension mapping Lottie `AnimatableProperty` values (scale, `positionTranslation`, `opacity`, `anchorPoint`, `rotation`) into native `graphicsLayer` properties (`scaleX`, `translationX`, `alpha`, `transformOriginX`, `rotationZ`).
- Added a grandparent regression screenshot test inside `LottieBasicScreenshotTest.kt` featuring a 3-deep layer hierarchy to verify the correct propagation of matrix scales.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


yschimke:
It seems to be a regression.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/Transform.kt
```diff
@@ -47,3 +51,29 @@ internal fun transform(
...
+  val translation = animatePosition(this.positionTranslation, animationSettings)
+  val opacity = animateScalar(this.opacity, animationSettings)
+  val anchorPoint = animatePosition(this.anchorPoint, animationSettings)
+  val scale = animateVector(this.scale, animationSettings)
+
+  val scaleX = scale[0] / 100f
+  val scaleY = scale[1] / 100f
+  val alpha = opacity / 100f
+
+  return RemoteModifier.graphicsLayer(
```
yschimke: Are these values all normalised the same way? Can you document the range for lottie.

With compose some of these are 0..1

```
    /** The horizontal center of the transform, as a fraction of the layer width. */
    public var transformOriginX: RemoteFloat
```

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/Transform.kt
```diff
@@ -47,3 +51,29 @@ internal fun transform(
...
+    scaleX = scaleX,
+    scaleY = scaleY,
+    translationX = translation.x,
```
yschimke: These are pixels, but do they need to be scaled to the current display size?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/Layer.kt
```diff
@@ -16,48 +16,46 @@
...
-      mutableListOf()
+internal fun LayerNode(layer: Layer, childrenMap: Map<Int?, List<Layer>>) {
+  if (layer.hidden == true) {
+    return
+  }
+
+  val transform = layer.transform
+  val modifier =
+    if (transform != null) {
+      transform.toModifier(LocalAnimationSettings.current)
```
yschimke, line 39: I wonder whetehr I put the scaling in the wrong place, in

https://github.com/google/horologist/blob/f86078ac069a5f4ae8d9c3826b518a572d66b4f0/remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/Shape.kt#L63

And whether we should fix the ordering of how we apply these.

cc @AshleyIngram

yschimke: I suspect we should try to pull my scaling up such that everything in the lottie document works in lottie coordinates, and not the document/compose.

But you might want to size your test/previews specifically to the lottie document size to take scaling out of the equation.

PR #2802, opened 2026-08-17
Lottie grandchild fix v2

MarkYav:
Fix for https://github.com/google/horologist/issues/2795

#### WHAT

- Replaced the simple parentTransforms mapping with a topological traversal algorithm (buildAncestorTransforms) to capture the entire hierarchy of ancestral transforms for each layer.
- Refactored Layer node parameters to expect and resolve the full topologically sorted ancestorStack: List<Transform>.
- Updated ShapeLayer structure so it correctly merges transforms, strictly preserving the top-down cascading render sequence (e.g., [Grandparent, Parent, Self]).
- Wrapped the heavy hierarchical allocation in a remember(animation.layers) block in LottieAnimation to prevent severe GC churn and UI jank during recompositions.

#### WHY

- Previously, Lottie layers completely discarded transforms generated past their immediate parent, causing visual breaks for deeply nested Lottie files. Providing the entire ancestral lineage as a sequential List<Transform> allows the RemoteCanvas affine matrix to apply each layer's intrinsic rotation, translation, and custom anchors gracefully and truthfully to the Lottie spec. 

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests


remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/Layer.kt
```diff
@@ -25,25 +25,27 @@ import com.google.android.horologist.remotecompose.lottie.format.LayerType
...
+  layer: Layer,
+  parentTransforms: Map<Int, List<Transform>>,
+  transform: Transform?,
+) {
+  val ancestorStack = parentTransforms[layer.index] ?: emptyList()
 
   // TODO: Replace passing a transform param in and applying it to the transform stack with
   // graphicsLayer transforms in the calling composable, once the ANDROID_NATIVE player supports
   // graphicsLayer
   // (b/408913726)
```
yschimke: Can we remove this comment, it's not true any more.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/LottieAnimation.kt
```diff
@@ -147,8 +148,59 @@ internal fun LottieAnimation(
...
+    populateAncestorTransforms(layer, emptyList(), childrenMap, map)
+  }
+
+  return map
+}
+
+private fun populateAncestorTransforms(
+  layer: com.google.android.horologist.remotecompose.lottie.format.Layer,
+  currentStack:
+    List<com.google.android.horologist.remotecompose.lottie.format.GraphicElement.Transform>,
```
yschimke: imports

remotecompose/lottie/src/test/java/com/google/android/horologist/remotecompose/lottie/LottieBasicScreenshotTest.kt
```diff
@@ -36,6 +36,17 @@ class LottieBasicScreenshotTest : WearScreenshotTest() {
...
+
+  @Test
+  fun grandparent() {
+    composeRule.setContent {
+      Box(modifier = Modifier.background(Color.White).testTag("Box")) {
+        LottiePreview(R.raw.grandparent, clock = SettableRemoteClock())
+      }
+    }
+
+    composeRule.onNodeWithTag("Box").captureRoboImage(testName("_grandparent"))
```
yschimke, line 50: It's hard from the test to reason about whether it's right or wrong.  Can we either describe how to tell if it's working.

or alternatively, is there a way to make the image tell a story. Such as each layer being a some item like a circle with a number, and the transforms make each layer move out of the way, but without the grandparent, you won't see all 3?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/LottieAnimation.kt
```diff
@@ -168,8 +171,45 @@ internal fun LottieAnimation(
...
     ) {
       for (layer in animation.layers) {
-        Layer(layer, parentTransforms, null)
+        Layer(layer, ancestorTransforms, null)
       }
     }
   }
 }
+
+private fun buildAncestorTransforms(layers: List<Layer>): Map<Int, List<Transform>> {
```
yschimke, line 180: After we land this, I might see if we can implement this as nested transforms, which is closer to the Compose way.

But not blocking for now.

yschimke (approved):
After fixing the screenshot

PR #2814, opened 2026-09-02
Lottie migrate values and primitives

MarkYav:
Support flexible Bézier curve closed flags and multi-stop gradient color values

#### Impact Summary
Enables resilient deserialization of Lottie Bézier paths with integer-encoded closed flags (`c: 0` / `c: 1`) and empty point arrays, and introduces the `GradientValue` primitive model with full color and opacity stop resolution. This allows Lottie animations containing gradients and integer closed flags to parse accurately without deserialization failures.

#### Motivation
In the Lottie JSON specification and various authoring tools (such as Bodymovin and After Effects), the Bézier path `c` (closed) property is often exported as an integer (`0` or `1`) rather than a strict boolean, and partial or empty point arrays can be emitted for degenerated paths. Furthermore, gradient fills and strokes rely on interleaved color and opacity stops in `GradientValue`, requiring stop decomposition and offset interpolation (`resolveStops()`). Adding these value primitives establishes foundational parser support for upcoming gradient properties and advanced styling.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests


remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Bezier.kt
```diff
@@ -16,14 +16,101 @@
...
+import kotlinx.serialization.json.jsonObject
+import kotlinx.serialization.json.jsonPrimitive
+import kotlinx.serialization.json.put
 
 /** A path defined by a set of bezier curves. */
-@Serializable
+@Serializable(with = BezierValueSerializer::class)
 internal data class BezierValue(
   @SerialName("c") val closed: Boolean = false,
   @SerialName("i") val inTangents: List<List<Float>> = emptyList(),
```
yschimke: is emptyList() really the default? or should these be required params?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Bezier.kt
```diff
@@ -16,14 +16,101 @@
...
   @SerialName("o") val outTangents: List<List<Float>> = emptyList(),
```
yschimke: as with other changes, we should move to Remote types in follow up.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Bezier.kt
```diff
@@ -16,14 +16,101 @@
...
+    }
+
+  override fun deserialize(decoder: Decoder): BezierValue {
+    val jsonDecoder = decoder as JsonDecoder
+    val element = jsonDecoder.decodeJsonElement()
+    val obj = element.jsonObject
+
+    val closed =
+      when (val cElem = obj["c"]) {
+        is JsonPrimitive -> cElem.booleanOrNull ?: ((cElem.intOrNull ?: 0) != 0)
```
yschimke: is it actually the case this can come as `true` or (0|1)

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Bezier.kt
```diff
@@ -16,14 +16,101 @@
...
+        put("i", buildPointArray(value.inTangents))
+        put("o", buildPointArray(value.outTangents))
+        put("v", buildPointArray(value.vertices))
+      }
+    )
+  }
+
+  private fun parsePointList(element: JsonElement?): List<List<Float>> {
+    if (element !is JsonArray) return emptyList()
+    return element.mapNotNull { item ->
```
yschimke: should we be silently accepting other values with mapNotNull? IT seems like we would use the values in the wrong place.

Does the spec say anything?

yschimke: same for it.jsonPrimitive.floatOrNull

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.buildJsonArray
+import kotlinx.serialization.json.buildJsonObject
+import kotlinx.serialization.json.floatOrNull
+import kotlinx.serialization.json.intOrNull
+import kotlinx.serialization.json.jsonPrimitive
+import kotlinx.serialization.json.put
+
+@Serializable
+internal data class ColorStop(val offset: Float, val red: Float, val green: Float, val blue: Float)
```
yschimke: use Remote types

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+
+@Serializable internal data class OpacityStop(val offset: Float, val alpha: Float)
+
+internal data class ResolvedColorStop(val offset: Float, val color: Color)
+
+@Serializable(with = GradientValueSerializer::class)
+internal data class GradientValue(
+  @SerialName("p") val numberOfColors: Int = 0,
+  @SerialName("k") val values: List<Float> = emptyList(),
```
yschimke: do we need these defaults?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+@Serializable
+internal data class ColorStop(val offset: Float, val red: Float, val green: Float, val blue: Float)
+
+@Serializable internal data class OpacityStop(val offset: Float, val alpha: Float)
+
+internal data class ResolvedColorStop(val offset: Float, val color: Color)
+
+@Serializable(with = GradientValueSerializer::class)
+internal data class GradientValue(
+  @SerialName("p") val numberOfColors: Int = 0,
```
yschimke: Are we missing some fields?

Image

yschimke: Ahhh, I think this is the earlier one.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+  @SerialName("k") val values: List<Float> = emptyList(),
+) {
+  val hasTransparency: Boolean
```
yschimke: Can we add links to the spec and maybe some comments

> All color and opacity stops are stored sequentially by ascending offsets in a flattened float array (color stops followed by opacity stops), with 4 floats per color stop and 2 floats per opacity stops. Thus, given color stops and opacity stops, the expected size for the gradient data array is 4 * Nc + 2 * No.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+    get() =
+      if (numberOfColors > 0) {
+        values.size > numberOfColors * 4
+      } else {
+        values.size > 0 && values.size % 4 != 0
```
yschimke: Why is this not just values.size > 0? or why do we need this else?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+          )
+        }
+      }
+      return stops
+    }
+
+  val opacityStops: List<OpacityStop>
+    get() {
+      val stops = mutableListOf<OpacityStop>()
+      val colorCount = if (numberOfColors > 0) numberOfColors else values.size / 4
```
yschimke: is this part of the spec?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+      val (r, g, b) = interpolateRgbAt(cStops, offset)
+      val alpha = interpolateAlphaAt(oStops, offset)
+      ResolvedColorStop(offset = offset, color = Color(red = r, green = g, blue = b, alpha = alpha))
+    }
+  }
+
+  private fun normalizeColorComponent(v: Float): Float =
+    if (v > 1f) (v / 255f).coerceIn(0f, 1f) else v.coerceIn(0f, 1f)
+
+  private fun interpolateRgbAt(stops: List<ColorStop>, offset: Float): Triple<Float, Float, Float> {
```
yschimke: are these color utils common in other places?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+      }
+      return stops
+    }
+
+  val opacityStops: List<OpacityStop>
+    get() {
+      val stops = mutableListOf<OpacityStop>()
+      val colorCount = if (numberOfColors > 0) numberOfColors else values.size / 4
+      val opacityOffset = colorCount * 4
+      var i = opacityOffset
```
yschimke: what about using kotlin collection utils.

values.drop(4 * numberOfColors).chunked(2).map { chunk -> ... }

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+
+@Serializable
+internal data class ColorStop(val offset: Float, val red: Float, val green: Float, val blue: Float)
+
+@Serializable internal data class OpacityStop(val offset: Float, val alpha: Float)
+
+internal data class ResolvedColorStop(val offset: Float, val color: Color)
+
+@Serializable(with = GradientValueSerializer::class)
+internal data class GradientValue(
```
yschimke, line 67: I'm now unclear if this is

2.5 Gradient

or

3.2.6 Gradient

https://lottie.github.io/lottie-spec/latest/single-page/#specs-shapes

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+      }
+    }
+    return stops.last().alpha
+  }
+}
+
+internal fun parseGradientValueElement(element: JsonElement?): GradientValue {
+  return when (element) {
+    null -> GradientValue()
+    is JsonArray -> {
```
yschimke: I suspect this when means we are combining Gradient property and Graphic Element

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+      val floatList = element.mapNotNull { it.jsonPrimitive.floatOrNull }
+      val count = if (floatList.size >= 4 && floatList.size % 4 == 0) floatList.size / 4 else 0
```
yschimke: is there a spec or can you comment about why this expression for count?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+    if (oStops.isEmpty()) {
+      return cStops.map {
+        ResolvedColorStop(
+          offset = it.offset,
+          color = Color(red = it.red, green = it.green, blue = it.blue, alpha = 1f),
+        )
+      }
+    }
+
+    val allOffsets = (cStops.map { it.offset } + oStops.map { it.offset }).distinct().sorted()
```
yschimke: distinct worries me

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
```diff
@@ -0,0 +1,213 @@
...
+        }
+      val count =
+        if (p > 0) p else if (values.size >= 4 && values.size % 4 == 0) values.size / 4 else 0
+      GradientValue(numberOfColors = count, values = values)
+    }
+    else -> GradientValue()
+  }
+}
+
+internal object GradientValueSerializer : KSerializer<GradientValue> {
```
yschimke: Does this actually Round Trip cleanly? It feels like it is handling two types, with a special value numberOfColors, discriminating.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Bezier.kt
MarkYav: Here I am not sure:
On one hand, the specs say "Represents boolean values as an integer. `0` is false, `1` is true."
On the other hand, in the JSON schema it is defined as
```json
"bezier": {
                "type": "object",
                "title": "Bezier",
                "description": "Cubic polybezier",
                "properties": {
                    "c": {
                        "title": "Closed",
                        "type": "boolean",
                        "default": false
                    },
                    ...
```
And according to the JSON spec, "The boolean type matches only two special values: true and false. Note that values that evaluate to true or false, such as 1 and 0, are not accepted by the schema."
Link: https://json-schema.org/understanding-json-schema/reference/boolean

So it is safer to support both of these options.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Bezier.kt
MarkYav: Removed the default values for required parameters

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Bezier.kt
MarkYav: `Bezier` is now completely migrated using Remote types. In `Gradient` there are timestamps (progress) that are not migrated yet, because they are used as indices.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Bezier.kt
MarkYav: Outdated after refactoring

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
MarkYav: Using remote types for colors

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
MarkYav: removed defaults for mandatory values

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
MarkYav: Added doc comments to the classes

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
MarkYav: It was a check that the colors are not corrupted, as we should have a progress (int) + three integers for RGB. But not I moved the verification to one place to avoid smoothing this logic over the code.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
MarkYav: Sometimes, if the gradient does not have the opacity values, the "p" (number of color stops) can be committed. This logic handles that. But after refactoring this codeblock is outdated.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
MarkYav: Search tells me we use it only in the `Gradient`.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
MarkYav: 2.5 Gradient
(it is a value, not a property)

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
MarkYav: There are two potential ways how the `GradientValue` can be encoded in json. Added the doc comment in code to explain this.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
MarkYav: Added a link to the specs in the doc comments

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
MarkYav: outdated after the refactoring

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
MarkYav: The thing is that the info in specs and JSON validation schema is different... So we support both formats.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
yschimke: I think it's still combining value and property.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
yschimke: let's reconsider.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/values/Gradient.kt
yschimke: discussed elsewhere, I thikn it's just a float array.

yschimke (approved):

MarkYav:
I was thinking what is the right way to parse the "gradient" value from json. The problem is the information we need is contained in different places:
- the parameter "p" (number of color stops in the gradient) is in the very root.
- the actual values for the colors are stored under "s" and encoded -- we need "p" to extract the colors and opacity.

See below:
```json
{
    *"p": 2,**
    "k": {
        "a": 1,
        "k": [
            {
                "t": 0,
                **"s": [0, 1, 0, 0, 1, 0, 0, 1],**
                "o": {"x": [0.333], "y": [0]},
                "i": {"x": [0.667], "y": [1]}
            },
            {
                "t": 60,
                **"s": [0, 0, 1, 0, 1, 1, 1, 0]**
            }
        ]
    },
    "ty": "g"
}
```

So, I decided to do it this way:
- have a `RawGradientValue` that is used for (de)serialization.
- have a `fun RawGradientValue.toGradient(colorStopCount: Int): GradientValue` to convert `RawGradientValue` to `GradientValue` we can work with.

Later, in the Gradient **Property**, we can call this conversion. What do you think?
```kotlin
@Serializable
data class GradientProperty(
    val p: Int,
    val k: AnimatableProperty<RawGradientValue>,
    ...
) {
    val resolvedGradient: AnimatableProperty<GradientValue>
        get() = k.mapValues { rawGradient -> rawGradient.toGradient(colorStopCount = p) }
}
```

Here is the link to the commit: https://github.com/google/horologist/pull/2814/changes/140b6c862c578e29cc5b5cf1dbfd706329cba866

yschimke:
> I was thinking what is the right way to parse the "gradient" value from json. The problem is the information we need is contained in different places:
> ...
> 
> Here is the link to the commit: [140b6c8](https://github.com/google/horologist/commit/140b6c862c578e29cc5b5cf1dbfd706329cba866)

I wonder if the typical way is with a custom root Serializer working on the JsonElement.  It could create a non-object Serializer configured with p?

Does GradientValue show up in other places, other than Gradient Property?

MarkYav:
> Does GradientValue show up in other places, other than Gradient Property?

Only in the Gradient Property and during the rendering of the Gradient in `renderer/.../Gradient.kt`.

> I wonder if the typical way is with a custom root Serializer working on the JsonElement. It could create a non-object Serializer configured with p?

This was also considered: then we can't use the `@Serializable` annotation on the `GradientValue` class. But we can still manually call the deserializer to parse the JSON and get the object.

Here is a demo commit: https://github.com/MarkYav/horologist/commit/1c3c7f2722cf2d1271ef202e0aa0609b72077b21

yschimke:
```
> Task :remotecompose:lottie:testDebugUnitTest

LottieDecoderResilienceTest > gradientProperty_scaled255Integers_samplesNormalizedColor FAILED
    com.google.common.truth.AssertionErrorWithFacts at LottieDecoderResilienceTest.kt:261

ParsingTest > gradientValue_stopDecomposition_normalizes255ScaledValues FAILED
    com.google.common.truth.AssertionErrorWithFacts at ParsingTest.kt:447

ParsingTest > gradientValue_getColorForPosition_normalizes255ScaledValues FAILED
    com.google.common.truth.AssertionErrorWithFacts at ParsingTest.kt:609
```

MarkYav:
> ```
> > Task :remotecompose:lottie:testDebugUnitTest
> 
> LottieDecoderResilienceTest > gradientProperty_scaled255Integers_samplesNormalizedColor FAILED
>     com.google.common.truth.AssertionErrorWithFacts at LottieDecoderResilienceTest.kt:261
> ...
> ```

Yes, I thought to remove the extra verification for 0..225 values (the default one is 0..1), but apparently some older versions of the Lottie may have values in this range. So returning this back.

yschimke (approved):

PR #2823, opened 2026-09-07
Align Lottie AST properties with Lottie specifications

MarkYav:
#### WHAT
Refactors and standardizes all foundational Lottie AST property models in `:remotecompose:lottie` to use native `RemoteCompose` types and enforcing Lottie 1.0.1 schema compliance.

#### WHY
1. **RemoteCompose state alignment**: Following the migration of primitives and values in PR #2814 (`Point`, `BezierValue`, `GradientValue`). Embedding `RemoteFloat`, `RemoteBoolean`, and `RemoteColor` directly into AST properties avoids runtime conversion adapters, prevents repeated per-frame allocations during timeline evaluation, and unifies the data representation between the format AST and the renderer.
2. **Lottie 1.0.1 Specification alignment**: Adding parsing support for all the properties encoded in the official Lottie JSON schema.

#### HOW
- Created custom serializers:
  - `RemoteFloatSerializer` and `IntBooleanRemoteSerializer` for streaming token decoding.
  - `RemoteColorSerializer` decoding 3- or 4-element arrays into `RemoteColor`.
  - `ScalarKeyframeValueSerializer` for single-element float arrays ([100.0]) on scalar keyframe "s".
  - `ScalarKeyframeEasingSerializer` supporting both numeric primitives and single-element arrays.
- Updated property renderers (`Scalar.kt`:63, `Position.kt`:41, `Vector.kt`:38) to consume AST Remote types directly.
- Added RGBA interpolation with cubic Bézier easing to `Color.kt`:48.

#### Checklist :clipboard:
- [x] Run spotless check
- [x] Run tests

### DIFF
--- a/remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/properties/Bezier.kt
+++ b/remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/properties/Bezier.kt
@@ -16,66 +16,148 @@
 
 package com.google.android.horologist.remotecompose.lottie.format.properties
 
+import androidx.compose.remote.creation.compose.state.rb
 import com.google.android.horologist.remotecompose.lottie.format.values.BezierValue
+import com.google.android.horologist.remotecompose.lottie.format.values.SerializableRemoteBoolean
+import com.google.android.horologist.remotecompose.lottie.format.values.SerializableRemoteFloat
 import kotlinx.serialization.DeserializationStrategy
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.SerializationException
 import kotlinx.serialization.json.JsonContentPolymorphicSerializer
 import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.intOrNull
-import kotlinx.serialization.json.jsonObject
 import kotlinx.serialization.json.jsonPrimitive
 
-/** A base class for bezier properties. */
+/**
+ * Base class for all Lottie animatable Bézier properties conforming to
+ * [Bézier Property](https://lottie.github.io/lottie-spec/dev/specs/properties/#bezier-property).
+ *
+ * Bézier properties define cubic polybezier shape paths (used in shapes such as vector paths and
+ * masks).
+ *
+ * Essential Invariants:
+ * - The property is partitioned into two mutually exclusive branches identified by the
+ *   integer-boolean discriminator [animated]:
+ *     - `0` (`false.rb`): [StaticBezierProperty], holding a constant [BezierValue].
+ *     - `1` (`true.rb`): [AnimatedBezierProperty], holding a sequence of keyframes over time.
+ * - [slotId]: Optional slot identifier (`sid`) enabling runtime value replacement via Lottie slots.
+ */
 @Serializable(with = BaseBezierPropertySerializer::class)
 internal sealed class BaseBezierProperty {
-  abstract val animated: Boolean
+  abstract val animated: SerializableRemoteBoolean
+  abstract val slotId: String?
 }
 
 /**
- * A static bezier. The value is an array of floats with 4 values, describing the 2 control points
- * of the curve.
+ * Conforms to
+ * [Bézier Property](https://lottie.github.io/lottie-spec/dev/specs/properties/#bezier-property)
+ * (Not animated branch):
+ * - Required Fields: `"a"` (const 0), `"k"` (Bézier shape value).
+ * - Optional Fields: `"sid"` (slot identifier, default null).
+ *
+ * Invariants:
+ * - [animated] is guaranteed to represent integer `0` (`false.rb`).
+ * - [value] contains the constant [BezierValue] geometry.
  */
 @Serializable
 internal data class StaticBezierProperty(
-  @SerialName("a") val animatedInt: Int = 0,
+  @SerialName("sid") override val slotId: String? = null,
+  @SerialName("a") override val animated: SerializableRemoteBoolean,
   @SerialName("k") val value: BezierValue,
-) : BaseBezierProperty() {
-  override val animated: Boolean
-    get() = animatedInt == 1
-}
+) : BaseBezierProperty()
 
-/** An animated bezier. */
+/**
+ * Conforms to
+ * [Bézier Property](https://lottie.github.io/lottie-spec/dev/specs/properties/#bezier-property)
+ * (Animated branch):
+ * - Required Fields: `"a"` (const 1), `"k"` (array of Bézier keyframes).
+ * - Optional Fields: `"sid"` (slot identifier, default null).
+ *
+ * Invariants:
+ * - [animated] is guaranteed to represent integer `1` (`true.rb`).
+ * - [keyframes] defines the temporal evolution of the Bézier shape across animation frames.
+ */
 @Serializable
 internal data class AnimatedBezierProperty(
-  @SerialName("a") val animatedInt: Int = 1,
+  @SerialName("sid") override val slotId: String? = null,
+  @SerialName("a") override val animated: SerializableRemoteBoolean,
   @SerialName("k") val keyframes: List<BezierKeyframe>,
-) : BaseBezierProperty() {
-  override val animated: Boolean
-    get() = animatedInt == 1
-}
+) : BaseBezierProperty()
 
-/** A single keyframe for an animated bezier property. */
+/**
+ * A single Bézier keyframe conforming to
+ * [Bézier Keyframe](https://lottie.github.io/lottie-spec/dev/specs/properties/#bezier-keyframe).
+ *
+ * Defines the Bézier shape value and optional easing interpolation parameters at a specific
+ * timeline frame.
+ *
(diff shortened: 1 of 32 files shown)


remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/properties/Color.kt
```diff
@@ -17,20 +17,189 @@
...
+            val r = lerpFloat(startR, endR, easedT)
+            val g = lerpFloat(startG, endG, easedT)
+            val b = lerpFloat(startB, endB, easedT)
+            val a = lerpFloat(startA, endA, easedT)
+            return Color(
+                r,
+                g,
+                b,
+                a,
+                androidx.compose.ui.graphics.colorspace.ColorSpaces.ExtendedSrgb,
```
yschimke, line 135: Worth a comment? And does it work? I'm pretty sure RemoteColor needs this to be eventually ARGB, does it get converted on the fly?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/properties/Vector.kt
```diff
@@ -41,23 +40,23 @@ internal fun animateVector(
...
+    is StaticVectorProperty -> vector.value
     is AnimatedVectorProperty -> {
       if (vector.keyframes.size == 1) {
-        return vector.keyframes[0].value.map { it.rf }
+        return vector.keyframes[0].value
       }
 
       val animationSegments = mutableListOf<List<AnimationSegment>>()
 
       val firstKeyframe = vector.keyframes[0]
```
yschimke, line 51: can it be empty at this point?

remotecompose/lottie/src/test/java/com/google/android/horologist/remotecompose/lottie/LottieDecoderResilienceTest.kt
```diff
@@ -78,7 +79,7 @@ class LottieDecoderResilienceTest {
               {
                 "ty": "fl",
                 "nm": "RedFill",
-                "c": { "k": [1.0, 0.0, 0.0, 1.0] }
+                "c": { "a": 0, "k": [1.0, 0.0, 0.0, 1.0] }
```
yschimke, line 82: How do I see that "a" is required?

remotecompose/lottie/src/test/java/com/google/android/horologist/remotecompose/lottie/BezierPropertyTest.kt
```diff
@@ -0,0 +1,1158 @@
...
+    when (value) {
+      null -> false
+      is RemoteBoolean -> value.constantValue
+      is Boolean -> value
+      else -> error("Unexpected boolean value type: ${value::class}")
+    }
+
+  private fun extractSlotId(prop: Any): String? =
+    try {
+      val getter = prop::class.java.methods.firstOrNull { it.name == "getSlotId" }
```
yschimke, line 57: can we avoid reflection?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/properties/Color.kt
```diff
@@ -17,20 +17,189 @@
...
+    is AnimatedColorProperty -> {
+      val keyframes = property.keyframes
+      if (keyframes.isEmpty()) {
+        return Color.Transparent.rc
+      }
+      if (keyframes.size == 1) {
+        return keyframes[0].value
+      }
+
+      val constantFrame = animationSettings.currentFrame.constantValueOrNull
```
yschimke, line 79: is currentFrame actually going to be constant? is it worth optimising for?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/properties/Bezier.kt
```diff
@@ -16,66 +16,149 @@
...
-    return if (animated) {
-      AnimatedBezierProperty.serializer()
-    } else {
-      StaticBezierProperty.serializer()
+    val obj = element as? JsonObject ?: throw SerializationException("Expected JSON object")
+    val animated = obj["a"]?.jsonPrimitive?.intOrNull
+    return when (animated) {
+      1 -> AnimatedBezierProperty.serializer()
+      0 -> StaticBezierProperty.serializer()
+      null ->
```
yschimke, line 159: can a be null if there is a slotId?

https://github.com/lottie/lottie-spec/blob/4b55957472f8718e34e9c1298f0e8f021b6c597f/schema/helpers/slottable-property.json#L5

yschimke (approved):
If it's at least better than the current logic, then some of these can be TODOs

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/properties/Bezier.kt
```diff
@@ -43,24 +44,30 @@ internal fun animateBezier(
...
+      val endKeyFrame = path.keyframes[1]
 
-      if (startKeyFrame.frame != 0f) {
-        return p.keyframes[0].value[0]
+      if (startKeyFrame.frame.constantValue != 0f) {
+        return path.keyframes[0].value[0]
       }
 
-      val duration = endKeyFrame.frame - startKeyFrame.frame
+      val duration = endKeyFrame.frame.constantValue - startKeyFrame.frame.constantValue
```
yschimke, line 70: Why do you need constantValue here?

If they are constants then you would get a RemoteFloat with a constant value anyway.  and animatePoints could take RemoteFloat.

It's possible we are doing Remote Types too early, but given it's use is TODO, it's hard to be sure.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/properties/Bezier.kt
MarkYav: Interesting point, as it affects Bezier property indirectly, but rather scalar, color, vector, and position: https://lottie.github.io/lottie-spec/1.0.1/single-page/#specs-helpers-slottable-property

"Slots are a way to define a property value once and use the value in multiple properties." -- it is a nice feature, but should go after the main ones in our priority list.

Added as a TODO.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/properties/Color.kt
MarkYav: This is definitely on the roadmap, but I tried to restrict the scope of this commit to changes in the `format/` folder. The changes in all other places are minimal and are the reflection of the changes in `format/`.

remotecompose/lottie/src/test/java/com/google/android/horologist/remotecompose/lottie/BezierPropertyTest.kt
MarkYav: Added as TODO, tests fill be fixed in one of the upcoming PRs.

remotecompose/lottie/src/test/java/com/google/android/horologist/remotecompose/lottie/LottieDecoderResilienceTest.kt
MarkYav: It is a line of JSON that defines a static color property (red color with 100% opacity). A [comment](https://github.com/google/horologist/pull/2823/changes/BASE..94a5916f93ea80bd8c8c2355558559a547f62db7#diff-abc2b487cbdec98d0812c7f9c16fc31675cbae80f51b404193cef440f905a68cR58) was added to the StaticColorProperty class description that the value "a" is required with the reference to the Lottie specs.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/properties/Vector.kt
MarkYav: I think yes. I did not implement a custom serializer to check if the list of keyframes is empty or not.

The specs say "When animated, k will be an array of keyframes."; and from a purely schema-validation perspective, the array of keyframes can be empty.

So it is a valid case, but it is not clear how to react in this case. Added as TODO, and will inspect the `android-lottie` implementation.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/renderer/properties/Color.kt
MarkYav: Added as TODO to check, as this class will be heavily refactored.

remotecompose/lottie/src/test/java/com/google/android/horologist/remotecompose/lottie/LottieDecoderResilienceTest.kt
yschimke: For all entries in this file?

PR #2831, opened 2026-09-08
Align Lottie AST layers with Lottie specifications.

MarkYav:
#### WHAT
Refactors the Lottie layer AST (`Layer`, `NullLayer`, `ShapeLayer`, `SolidColorLayer`) to strictly adhere to the Lottie 1.0.1 specification (Visual Layer Properties, Visual Layer, Shape Layer, Solid Layer, and Null Layer).

Layer lifecycle boundaries (`ip` / `op`) now store frame indices as `RemoteFloat`, visibility (`hd`) uses standard JSON booleans via `BooleanRemoteSerializer`, and integer dimensions in `SolidColorLayer` (`sw` / `sh`) use `RemoteInt`. `SolidColorLayer.solidColor` (`sc`) is unified to `RemoteColor`, supported by extending `RemoteColorSerializer` to parse both canonical float arrays and hex color strings.

Introduces missing Lottie visual layer attributes: Track Matte modes (`tt` / `MatteMode`), Matte Parent index (`tp`), Auto Orient (`ao` / `SerializableRemoteBoolean`), and layer clipping masks (`masksProperties` / `Mask`). Removes obsolete non-spec Bodymovin attributes (`st`, `sr`, `ddd`, and `td`).

#### WHY
1. **Lottie 1.0.1 Specification Conformance:**
   - Base `Layer` now models all common visual layer attributes conforming to Section 3 of the Lottie specification.
   - Replaced legacy Bodymovin track matte flag `"td"` (matte target) with the official `"tp"` (matte parent) and `"tt"` (`MatteMode`) per Lottie specification.
   - Removed unsupported 3D and legacy time-stretching fields (`"ddd"`, `"st"`, `"sr"`).
   - Introduced `Mask` and `MaskMode` to support layer clipping paths and visibility operations.
2. **RemoteCompose State Type Alignment:**
   - Frame boundaries (`startFrame`, `endFrame`) are stored directly as `RemoteFloat`, eliminating `.rf` conversions during timeline evaluation.
   - `hidden` is represented as `RemoteBoolean` (`false.rb`), enabling consistent boolean state handling.
   - `solidWidth` and `solidHeight` are stored as `RemoteInt` (`.ri`), honoring the integer specification while integrating with RemoteCompose state.
   - `solidColor` adopts `RemoteColor`, decoupling downstream rendering from raw string parsing.
3. **Spec-Driven Serializer Separation & Resilience:**
   - Dedicated `BooleanRemoteSerializer` handles JSON boolean primitives (`true`/`false`), preserving strict integer validation on `IntBooleanRemoteSerializer` (`0`/`1`).
   - Dedicated `RemoteIntSerializer` provides streaming zero-allocation decoding of 32-bit integers into `RemoteInt`.
   - Polymorphic `LayerSerializer` defaults unsupported layer types (`Precomposition`, `Image`, `Text`, `Audio`) to `NullLayer`, ensuring parent transform hierarchies remain intact without crashing composition loading.

#### Checklist :clipboard:
- [ ] Run spotless check
- [ ] Run tests

### DIFF
--- a/remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
+++ b/remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
@@ -38,14 +38,13 @@ import kotlinx.serialization.json.jsonPrimitive
 
 /**
  * Base class for all Lottie animation layers conforming to
- * [Lottie Layers](https://lottie.github.io/lottie-spec/1.0.1/specs/layers/#common-properties) and
  * [Visual Layer](https://lottie.github.io/lottie-spec/1.0.1/specs/layers/#visual-layer).
  *
  * Layers are independent visual, temporal, and spatial nodes arranged in a compositing tree.
  *
  * Essential Invariants:
  * - Discriminator: Partitioned by integer [type] ([Layer
- *   Type](https://lottie.github.io/lottie-spec/1.0.1/specs/constants/#layer-type)).
+ *   Type](https://lottie.github.io/lottie-spec/1.0.1/specs/layers/#layer-types)).
  * - Parenting Hierarchy: Child layer transforms are concatenated with their parent's current
  *   transformation matrix: CTM(child) = CTM(parent) * Transform(child).
  * - Timeline Visibility Window: A layer is active on frame t when ip <= t < op.
@@ -70,7 +69,7 @@ internal sealed class Layer {
 
 /**
  * Canonical layer types defined in the
- * [Lottie Specification](https://lottie.github.io/lottie-spec/1.0.1/specs/constants/#layer-type).
+ * [Lottie Specification](https://lottie.github.io/lottie-spec/1.0.1/specs/layers/#layer-types).
  */
 @Serializable(with = LayerTypeSerializer::class)
 internal enum class LayerType(val value: Int) {
@@ -78,9 +77,7 @@ internal enum class LayerType(val value: Int) {
   Solid(1),
   Image(2),
   Null(3),
-  Shape(4),
-  Text(5),
-  Audio(6);
+  Shape(4);
 
   companion object {
     fun fromValueOrNull(value: Int): LayerType? {
@@ -91,7 +88,7 @@ internal enum class LayerType(val value: Int) {
 
 /**
  * Polymorphic serializer for [Layer] discriminating on the integer "ty" field per
- * [Layer Type](https://lottie.github.io/lottie-spec/1.0.1/specs/constants/#layer-type).
+ * [Layer Type](https://lottie.github.io/lottie-spec/1.0.1/specs/layers/#layer-types).
  *
  * Contract:
  * - Deserialization Preconditions: [element] must be a [JsonObject].
--- a/remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
+++ b/remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
@@ -17,6 +17,10 @@
 package com.google.android.horologist.remotecompose.lottie.format.layer
 
 import com.google.android.horologist.remotecompose.lottie.format.graphicelement.grouping.Transform
+import com.google.android.horologist.remotecompose.lottie.format.mask.Mask
+import com.google.android.horologist.remotecompose.lottie.format.values.SerializableBoolean
+import com.google.android.horologist.remotecompose.lottie.format.values.SerializableRemoteBoolean
+import com.google.android.horologist.remotecompose.lottie.format.values.SerializableRemoteFloat
 import kotlinx.serialization.DeserializationStrategy
 import kotlinx.serialization.KSerializer
 import kotlinx.serialization.Serializable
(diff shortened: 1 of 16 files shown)


MarkYav:
I would encourage here a discussion if we should merge the serialization logic into the same serializer or not.

For example, a boolean value in JSON can be encoded as true/false or as 1/0. And specs clearly indicate what approach should be taken.

A similar situation is with color: it can be stored as RGB(A) or in a hex form. And the specs say that that a hex format is expected in the Solid Color Layer: https://lottie.github.io/lottie-spec/latest/specs/layers/#solid-layer:~:text=sc,Solid%20fill%20color

On one hand, we can have the deserialization logic in on serializer class. It make less cognitive load to think what serializer class we should use. On the other hand, the serializetion would not work correctly for some cases (or we have to store additional info about the original data format). But we don't use the serializetion.

yschimke:
I think you need to split them. You identified why yourself. Particularly if we think you should be able to roundtrip.

HexColorRemoteSerializer and RemoteColorSerializer ?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
yschimke, line 109: Null layer has fields, so can it fail if they don't match?

Should we have some UnknownLayer.serializer() instead? or we know it's safe?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/mask/Mask.kt
```diff
@@ -0,0 +1,102 @@
...
+  @SerialName("x") val expand: BaseScalarProperty? = null,
+)
+
+/**
+ * Mask mode indicating how the mask path combines with other masks conforming to
+ * [Mask Mode](https://lottie.github.io/lottie-spec/1.0.1/specs/constants/#mask-mode).
+ */
+@Serializable(with = MaskModeSerializer::class)
+internal enum class MaskMode(val value: String) {
+  Add("a"),
```
yschimke, line 62: Where did we get the extgra values from?

https://lottie.github.io/lottie-spec/latest/specs/constants/#mask-mode

remotecompose/lottie/src/test/java/com/google/android/horologist/remotecompose/lottie/LottieScalingDiffScreenshotTest.kt
```diff
@@ -246,8 +246,8 @@ class LottieScalingDiffScreenshotTest(
             ShapeLayer(
```
yschimke, line 246: Ideally we would have unit tests covering more of this,but can be a TODO for follow up.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
```diff
@@ -27,42 +31,77 @@ import kotlinx.serialization.encoding.Decoder
...
+ */
 @Serializable(with = LayerTypeSerializer::class)
 internal enum class LayerType(val value: Int) {
+  Precomposition(0),
   Solid(1),
+  Image(2),
   Null(3),
-  Shape(4);
+  Shape(4),
+  Text(5),
```
yschimke: Where did Text and Audio come from?

https://lottie.github.io/lottie-spec/latest/specs/layers/

yschimke (changes requested):

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
```diff
@@ -27,42 +31,77 @@ import kotlinx.serialization.encoding.Decoder
...
   abstract val transform: Transform?
+  abstract val autoOrient: SerializableRemoteBoolean
+  abstract val matteMode: MatteMode
+  abstract val matteParent: Int?
+  abstract val masks: List<Mask>?
 }
 
+/**
+ * Canonical layer types defined in the
+ * [Lottie Specification](https://lottie.github.io/lottie-spec/1.0.1/specs/constants/#layer-type).
```
yschimke: wrong link?

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
```diff
@@ -27,42 +31,77 @@ import kotlinx.serialization.encoding.Decoder
...
-      return values().firstOrNull { it.value == value }
+      return entries.firstOrNull { it.value == value }
     }
   }
 }
 
-/** Polymorphic serializer for [Layer] based on integer "ty" field. */
+/**
+ * Polymorphic serializer for [Layer] discriminating on the integer "ty" field per
+ * [Layer Type](https://lottie.github.io/lottie-spec/1.0.1/specs/constants/#layer-type).
```
yschimke: wrong link, and others.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
MarkYav: The Null layer is not for the case if something fails, but rather to aggregate sub-elements (like a folder). It is a good question what to do if the deserialization fails. I see three options:
1. Don't render the layer and all the nested elements
2. Don't render the whole animation
3. Through a parsing exception

I lean towards option three, as working animation can be misleading and hide the error.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
MarkYav: Fixed the link.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
MarkYav: Fixed the link.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
MarkYav: Removed the layers that are outside of Lottie spec 1.0.1.

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/mask/Mask.kt
MarkYav: Removed the values that are outside of Lottie spec.

remotecompose/lottie/src/test/java/com/google/android/horologist/remotecompose/lottie/LottieScalingDiffScreenshotTest.kt
MarkYav: Added as TODO.

MarkYav:
> I think you need to split them. You identified why yourself. Particularly if we think you should be able to roundtrip.
> 
> HexColorRemoteSerializer and RemoteColorSerializer ?

Refactored to have two separate serializers 👍

remotecompose/lottie/src/main/java/com/google/android/horologist/remotecompose/lottie/format/layer/Layer.kt
yschimke: I think the spec argues for unknown? so 1?

yschimke (approved):

